PR #3, opened 2021-11-29
User model

scottPlusPlus:
Adding a rough POC for a UserStore, currently only works in-memory. 

Also needed to bang on the configs a bit to pull in the shared Data-Model folder to the typescript build process


backend/src/auth/test/TestMockUserStore.ts
```diff
@@ -0,0 +1,53 @@
+import { UserModel } from "../data_model/UserModel";
```
ArendPeter, line 1: Thanks for adding tests!

backend/src/auth/data_model/UserModel.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Uid } from "../../../../domain_model/Uid"
+
+
+export class UserModel {
+    id: Uid;
+    email: Email;
+    hashedPassword: string;
+    salt: string;
+
+    constructor(data: Pick<UserModel, "id" | "email" | "hashedPassword" | "salt">) {
```
ArendPeter, line 11: I haven't used Pick before, I'm assuming that his requires that id, email, hashedPassword, and salt are given as separate parameters and then it's combined into the data parameter?

backend/src/auth/data_model/UserModel.ts
```diff
@@ -0,0 +1,31 @@
...
+        this.id = data.id;
+        this.email = data.email;
+        this.hashedPassword = data.hashedPassword;
```
ArendPeter, line 14: At what point will we generate the hashedPassword from the raw password? Could we take the raw password as a parameter and do the hashing here?

ArendPeter:
I think the c52dcf92e8e64063350e700aa9ec5ad8b692f845 commit made it into both reviews

ArendPeter (approved):
I mostly had some questions for my understanding, but overall it looks good to me, nice work!

backend/src/auth/test/TestMockUserStore.ts
scottPlusPlus: One thing we'll need to do is find a good testing framework

backend/src/auth/data_model/UserModel.ts
scottPlusPlus: My understanding is its a subset of fields from the initial type (in this case UserModel).  This was effectively a shortcut from defining an interface with the same fields as the UserModel, and the constructor will take anything (including an anonymous object) that has all the fields in the Pick

backend/src/auth/data_model/UserModel.ts
scottPlusPlus: We'd do that when a user first registers, or changes their password.  I didn't stick that in the constructor as I imagine those being explicit methods.

PR #6, opened 2021-12-17
Add placeholder login page

ArendPeter:
Here I've added a basic login page 

image

This also gives an idea for how we'd manage compatibility with multiple identity providers, we can have a dict with all the details that the identity provider needs, and someone could fork the repo and update the fields to match their identity provider

Here's what it looks like if I update the fields to cognito

image

We're probably going with KeyCloak, but I wanted to include the cognito example anyway so that we can verify our system can work with multiple providers


mikefranze:
Looks good, I'll need to study up on the auth config stuff.

ArendPeter:
> Looks good, I'll need to study up on the auth config stuff.

Thanks!

Here's my favorite overview of oAuth

https://www.youtube.com/watch?v=t18YB3xDfXI

PR #15, opened 2021-12-24
Converting main react app to typescript

mikefranze:


mikefranze:
Ran into some errors converting some of the components, so just updated main app for now.

ArendPeter:
I hadn't heard of .tsx files, I always called them .ts

https://stackoverflow.com/questions/56871384/what-is-the-difference-between-ts-and-tsx-extensions-both-are-used-as-extensi#:~:text=tsx%20extension%20is%20used%20when,not%20support%20adding%20JSX%20Elements.&text=when%20you%20use%20.,in%20that%20particular%20file%20while%20.

Sounds like .tsx is correct since you're using JSX. I learned something new

PR #19, opened 2021-12-28
Fix login invalid login path when deployed to heroku

### LINKED ISSUE #7: Login page doesn't work on heroku deployment
Goes to localhost:3000

ArendPeter:
The functionality works the same on my local version except I'm not referencing localhost anymore

fixes #7


mikefranze (approved):
After some googling, some people claim that window.location.origin might be an issue when using IE, but we can test that out and create a new issue if that's the case.

PR #23, opened 2022-01-10
Updates to front and back end to use common domain models and some minor UI updates.

mikefranze:
Some front end components are still a bit messy with how the data is structured. Mostly restructuring it to fit domain models prior to sending to backend. All these updates also assume one poll per election, will need to restructure when we want to add more polls.


backend/src/index.ts
```diff
@@ -50,12 +53,14 @@ app.get('/API/Elections', (req, res) => {
...
-        Candidates: Candidates,
-    }
+    // const Candidates = req.body.Election.Candidates.map((CandidateName:any,ind:any) => ({id: ind,CandidateName: CandidateName}))
+    // const newElection = {
+    //     id: Elections.length,
+    //     ElectionName: req.body.Election.ElectionName,
+    //     Candidates: Candidates,
+    // }
+    const newElection = req.body.Election as Election;
+    newElection.electionId = String(Elections.length);
```
ArendPeter, line 63: Incremental ids is fine for now, but I wonder if we should consider random ids in the future?

Incremental shows people how many elections are on the site, and it makes it easy to iterate over all elections if they wanted to (which could be a privacy issue). I'm not sure if these issues matter to us (since we'll require auth to actually vote), but it could be worth thinking about more

Tom Scott goes into more details on the pros and cons at the end of this video: https://www.youtube.com/watch?v=gocwRvLhDf8

frontend/src/components/AddElection.tsx
```diff
@@ -1,17 +1,22 @@
 import { useState } from "react"
 import React from 'react'
 import { useNavigate } from "react-router"
+import {Election} from './../../../domain_model/Election'
```
ArendPeter, line 4: nit

``./../../../domain_model/Election`` is equivalent to ``../../../domain_model/Election`` right?

frontend/src/components/AddElection.tsx
```diff
@@ -34,8 +40,30 @@ const AddElection = ( {  }) => {
...
+                    } as Candidate;
+            })
+        }
+
+        const NewElection : Election = {
+            electionId:    '0', // identifier assigned by the system
+            frontendUrl:   '', // base URL for the frontend
+            title:         electionName, // one-line election title
+            description:  '', // mark-up text describing the election
+            startUtc:     new Date(),   // when the election starts
```
ArendPeter, line 60: I added an issue to support these fields in the UI, I think that would be a good first issue for a newcomer :)

https://github.com/Equal-Vote/star-server/issues/24

frontend/src/components/VotePage.tsx
```diff
@@ -5,37 +5,46 @@ import useFetch from "../useFetch";
...
-            )
-          }
-          console.log(message)
+        // const message = {
+        //     ElectionID: id,
+        //     candidateScores: election.polls[0].candidates.map((candidate,i) => 
+        //       ({'id': election.polls[0].candidates[i].candidateId, 'score':rankings[i]})
+        //     )
+        //   }
+        const ballot = {
```
ArendPeter, line 30: nit

This might be another case where ``const ballot: Ballot = {`` would be preferred over ``as Ballot``

ArendPeter (approved):
I left a few comments, but none of them are major

Approved! Nice work

ArendPeter:
> Some front end components are still a bit messy with how the data is structured. Mostly restructuring it to fit domain models prior to sending to backend. All these updates also assume one poll per election, will need to restructure when we want to add more polls.

Makes sense for now! I added a issue for us to address this in the future https://github.com/Equal-Vote/star-server/issues/25

backend/src/index.ts
```diff
@@ -1,14 +1,16 @@
...
-
+import { Ballot } from '../../domain_model/Ballot';
+import { Score } from '../../domain_model/Score';
 const app = express();
 
 //Example data structures that would be sent in API responses
-const Elections = require('./Elections')
-const Election = require('./Election')
-const Ballots = new Array();
+const Elections = [] as Election[];//require('./Elections')
```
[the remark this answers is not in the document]
mikefranze: Some of the typescript syntax is confusing to me, I thought I had tried something like that but I'll give it a shot in the future.

backend/src/index.ts
mikefranze: Agreed, I think postgres would handle updating the ids? This is just a placeholder to store the data in memory until we get databases working.

ArendPeter (approved):

Issue #24, opened 2022-01-10
Include more optional fields in the CreateElection page


ArendPeter:
Currently our CreateElection page only allows for the title, and the candidate names, but our backend also supports the following

 - [x] #28
 - [x] #27
 - [x] #29

Let's update the CreateElection page to include these as well :)

PR #26, opened 2022-01-15
Adding a start/stop times, voting method, and number of winners

mikefranze:
Start and stop times aren't used by back end yet, but we can set them.


ArendPeter (approved):
Looks good to me! It mostly addresses the issue here https://github.com/Equal-Vote/star-server/issues/24

I also checked the deployment, and that looks good

PR #37, opened 2022-01-22
Adding more structure to back end

mikefranze:
Moved routes and other functions into their own folders and files.

### DIFF
--- a/backend/src/Controllers/elections.controllers.ts
+++ b/backend/src/Controllers/elections.controllers.ts
@@ -1,0 +1,63 @@
+import { Election } from '../../../domain_model/Election';
+import { Ballot } from '../../../domain_model/Ballot';
+import { Score } from '../../../domain_model/Score';
+const ElectionsModel = require('../Models/Elections')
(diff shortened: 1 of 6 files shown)


ArendPeter:
I'm not familiar with the ``.controllers`` / ``.routes`` file name pattern. It seems redundant since they match the folder name. Is there another benefit for them?

ArendPeter (approved):
Looks good! It's a lot cleaner and organized now :D

mikefranze:
> I'm not familiar with the `.controllers` / `.routes` file name pattern. It seems redundant since they match the folder name. Is there another benefit for them?

I'm not sure either, I've just run into 2 different example projects that used that convention so I used it too.

ArendPeter:
> > I'm not familiar with the `.controllers` / `.routes` file name pattern. It seems redundant since they match the folder name. Is there another benefit for them?
> 
> I'm not sure either, I've just run into 2 different example projects that used that convention so I used it too.

Cool, that's good enough for me :)

PR #55, opened 2022-02-20
Updates to results page and parser outputs

mikefranze:


backend/src/StarResults.cjs
```diff
@@ -242,7 +242,7 @@ function splitWinners(finalists, candidates, matrix) {
     return { winners: [], losers: [] };
   }
   if (candidateOrder.length === 1) {
-    return { winners: candidateOrder, losers: [] };
+    return { winners: candidateOrder, losers: [],votes};
```
ArendPeter, line 245: Should votes get it's own key in the dictionary?

```
return { winners: candidateOrder, losers: [], votes: votes};
```

backend/src/StarResults.cjs
```diff
@@ -503,8 +509,19 @@ export function splitPR(candidates, scores, nWinners) {
...
+  }
+  weightedSumsByRound.map((weightedSums,i) => {
+    weightedSums.map((weightedSum,j) => {
+      let candidate = candidatesByRound[i][j]
+      weightedSumsData[candidate.index][i] = weightedSum*5
+    })
+  })
   losers = candidates;
-  return { winners, losers, others, ties, debuginfo };
+  return { winners, losers, others, ties, debuginfo,weightedSumsData };
```
ArendPeter, line 524: I'm not familiar with this syntax, the braces look like a dictionary, but it's not using the key:value format. I tried looking it up, but I probably wasn't searching for the right thing

backend/src/StarResults.cjs
```diff
@@ -648,15 +665,18 @@ function flatten(cvr, sections) {
...
       title: data.winners.length > 1 ? "Winner (TIE)" : "Winner",
-      candidates: data.winners
+      candidates: data.winners,
+      votes: data.winnersVotes 
     },
     {
       title: data.losers.length > 1 ? "Runner-Up (TIE)" : "Runner Up",
-      candidates: data.losers
+      candidates: data.losers,
+      votes: data.losersVotes
```
ArendPeter, line 679: I'm a fan of using trailing commas. If ``candidates: data.losers`` already had a comma, then this would only be a single line change, likewise if we give ``votes: data.losersVotes`` a comma then this will make the code diff cleaner the next time we update this

That said it's totally personal preference, I'm fine either way as long as we're consistent in whichever one we choose

ArendPeter (changes requested):
I left a comment about whether votes should have it's own key, that's the only one I wanted to verify before approval

Aside from that, looks good!

backend/src/StarResults.cjs
mikefranze: I'm not certain but I think without the key it just uses the variable name as the key. There are other instances later in this function where we return doesn't have any keys. This did point out that I wasn't using consistent names in the return objects in the edge cases though which might cause an issue.

PR #58, opened 2022-03-02
Move keycloak secret to an environment variable

ArendPeter:
I updated keycloak to use the KEYCLOAK_SECRET environment variable

I also installed dotenv to make it easier to setup local environment variables

This will require changing the environment variable on the Heroku side, so I won't merge until that's completed


mikefranze:
Looks good

mikefranze (approved):

PR #60, opened 2022-03-04
Adding voter roll database table

mikefranze:
Initial structure for voter rolls and some debug methods

In the market for better names for voter rolls and entries.

### DIFF
--- a/backend/src/Models/VoterRolls.ts
+++ b/backend/src/Models/VoterRolls.ts
@@ -1,0 +1,125 @@
+import { VoterRoll } from '../../../domain_model/VoterRoll';
+var format = require('pg-format');
+class VoterRollDB {
+
+    _postgresClient;
+    _tableName: string;
+
+    constructor(client: any, tableName: string) {
+        this._postgresClient = client;
+        this._tableName = tableName;
+    }
+
+    init(): Promise<VoterRollDB> {
+        console.log("VoterRollDB.init");
+        var query = `
+        CREATE TABLE IF NOT EXISTS ${this._tableName} (
+            voter_roll_id  SERIAL PRIMARY KEY,
+            election_id     VARCHAR,
+            voter_id        VARCHAR,
+            ballot_id       INTEGER,
+            submitted       BOOLEAN
+          );
+        `;
+        console.log(query);
+        var p = this._postgresClient.query(query);
+        return p.then((_: any) => {
+            return this;
+        });
+    }
+
+    submitVoterRoll(election_id: string, voter_ids: string[],submitted:Boolean): Promise<boolean> {
+        console.log(`VoterRollDB.submit`);
+        var values = voter_ids.map((voter_id) => ([election_id,
+            voter_id,
+            submitted]))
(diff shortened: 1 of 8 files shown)


backend/src/Models/VoterRolls.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    constructor(client: any, tableName: string) {
+        this._postgresClient = client;
+        this._tableName = tableName;
+    }
+
+    init(): Promise<VoterRollDB> {
+        console.log("VoterRollDB.init");
+        var query = `
+        CREATE TABLE IF NOT EXISTS ${this._tableName} (
```
ArendPeter, line 16: I want to make sure I'm understanding the usage correctly, is this example correct?

A user wants to vote, so we query the voter roll db

```
SELECT * FROM table WHERE voter_id=voter
```

and then we allow the user to vote if there's a matching entry, and submitted is false?

backend/src/Models/VoterRolls.ts
```diff
@@ -0,0 +1,125 @@
...
+            voter_roll_id  SERIAL PRIMARY KEY,
+            election_id     VARCHAR,
+            voter_id        VARCHAR,
+            ballot_id       INTEGER,
```
ArendPeter, line 20: Could we remove submitted, and instead let ballot_id=null signify a non-submitted ballot? Not sure if that's a good idea or not, mostly curious if that's an option

backend/src/Models/VoterRolls.ts
mikefranze: Depends on the election settings. In a closed election correct. In an open election we'd still allow them to vote if there isn't a matching entry and then create one when the ballot is submitted.

backend/src/Models/VoterRolls.ts
mikefranze: The only reason I'd say no is that I'm not sure if ballot_id should be here in the first place. As we discussed in the last meeting we may or may not want to be able to retrieve ballots for a specified user. I decided to add it for now.

ArendPeter: Cool, that makes sense

On a separate note, election_id, and voter_id kind of feel like keys in this table. There can be multiple entries per election or per user, but each election_id - voter_id pair should only appear once

Maybe we could use [composite keys ](https://www.simplilearn.com/tutorials/sql-tutorial/composite-key-in-sql#:~:text=A%20composite%20key%20in%20SQL%20can%20be%20defined%20as%20a,can%20uniquely%20identify%20any%20record.) here? I haven't used them before, but it could be a good fit, and it would allow us to remove voter_roll_id

mikefranze: I didn't know about this, I was going to add code to check if the combination already exists but this is much simpler thanks! I looked up a postgres example and added it.

ArendPeter: Awesome, I'm glad it worked :D

ArendPeter (approved):
Looks good!

PR #61, opened 2022-03-05
Add ownership support

ArendPeter:
This PR does the following

1. Requires login for election creation
2. Passes actual user id when creating elections
3. My Elections now filters to elections that you created

Also I added a logout button


backend/src/Models/Elections.ts
```diff
@@ -64,13 +64,24 @@ class ElectionsDB {
         });
     }
 
-    getElections(): Promise<Election[] | null> {
+    getElections(filter: string): Promise<Election[] | null> {
+        // When I filter in trello it adds "filter=member:arendpetercastelein,overdue:true" to the URL, I'm following the same pattern here
```
mikefranze, line 68: Interesting way to do this, should allow us to easily add more filtering to the front end later.

frontend/src/components/AddElection.tsx
```diff
@@ -82,7 +82,7 @@ const AddElection = () => {
             description: description, // mark-up text describing the election
             start_time: new Date(startDate),   // when the election starts 
             end_time: new Date(stopDate),   // when the election ends
-            owner_id: '0',
+            owner_id: authSession.getIdField('sub'),
```
mikefranze, line 85: We'll want to verify that this matches the user id in the backend

mikefranze (approved):
Nice work

frontend/src/components/AddElection.tsx
ArendPeter: Yep, it matches up

image

PR #73, opened 2022-04-09
Add Edit election feature

### LINKED ISSUE #67: Update jest to be a dev dependency
jest is only used during development, so we can move it from a primary dependency to a dev depencency

### LINKED ISSUE #69: Add Edit Election support

ArendPeter:
This adds an edit button which allows you to edit elections, fixes #69 

# Included in this PR
## Backend
 * Cleaned up our logging statements
 * fixes #67 
 * Added editElection backend API
 * Add tests for the backend API
 
## Frontend
 * Add edit page and edit button
 * Create ElectionForm to contain shared code between EditElection and AddElection
 * Refactor election form to use fewer state variables and to be more consistent with the backend (the [commit diff](https://github.com/Equal-Vote/star-server/pull/73/commits/019589298caf36f31d2ff2e7e21a974c5637e904) is a little easier to follow for this change)
 * Add styling logic so edited fields are in bold (but only for textfields)
 * Tweaked the "title not set" error message to update the box instead of using an alert
 
## Docs
 * Update docs to include commands for running the frontend/server
 
# Issues in this PR

I didn't have time to address all of these, so I'll make issues and address them later

 * The voter roll isn't editable yet, this may also require some discussion for the best way to do this
 * I locked the react-router version to 6.0.0, because 6.3.0 was causing crashes (pretty sure it's unrelated to my changes though). We should upgrade that dependency later
 * When selecting the day, it converts the selected time to UTC and can sometimes end up with the wrong day. So if you select a date in the evening on April 8th and you're on PST, then it's already April 9th UTC so it'll select the wrong day
 * Edited textfields become bolded, if we like that feature then I can add a task for applying it to the selection and date boxes as well. If we don't then I'll add a task to remove the bolding


backend/src/test/app.test.ts
```diff
@@ -16,201 +16,308 @@ afterEach(() => {
     jest.clearAllMocks();
 });
 
-describe("Post Election", () => {
+// describe("Post Election", () => {
```
mikefranze, line 19: Why are all the other test cases commented out?

backend/src/test/app.test.ts
```diff
@@ -16,201 +16,308 @@ afterEach(() => {
...
+// 
+//     })
+// })
+
+
+
+describe("Edit Election", () => {
+    // TODO: I'm making a lot of calls to add elections within the edit election tests
+    //       Ideally I'd like to setup the mocks without relying on the other apis, but I couldn't figure out a way to do that
+    const setupInitialElection = async () => {
```
mikefranze, line 80: You should be able to accomplish this with a "before each" set up function https://jestjs.io/docs/setup-teardown

backend/src/test/app.test.ts
```diff
@@ -16,201 +16,308 @@ afterEach(() => {
...
+//         })
+// 
+//     })
+// })
+
+
+
+describe("Edit Election", () => {
+    // TODO: I'm making a lot of calls to add elections within the edit election tests
+    //       Ideally I'd like to setup the mocks without relying on the other apis, but I couldn't figure out a way to do that
```
mikefranze, line 79: I couldn't figure out a good way around that either, but any set up API that we need should be well tested earlier in the test suit

mikefranze (changes requested):
Awesome work! Just need some of the older test cases restored

backend/src/test/app.test.ts
ArendPeter: Thanks! I'll look into this next time

backend/src/test/app.test.ts
ArendPeter: ah mb, I was trying to test them individually forgot to uncomment. I've since learned better ways to run specific tests

When I uncommented the tests, some of them were failing, but I don't think they're related to changes in this PR. More details here https://github.com/Equal-Vote/star-server/issues/74

mikefranze (approved):

Issue #66, opened 2022-03-22
Rename voter_roll to election_roll


ArendPeter:
This helps clarify that the roll is related to the election, not an individual voter

This will require renaming databases, so it's a non trivial task

Related https://github.com/Equal-Vote/star-server/pull/64#discussion_r831690475

PR #79, opened 2022-04-17
Changing voter roll to election roll

mikefranze:
#66


frontend/src/components/VotePage.tsx
```diff
@@ -64,6 +74,7 @@ const VotePage = ({ }) => {
             readonly={false}
             onSubmitBallot={submit}
           />}
+        {<div> {submitError} </div>}
```
ArendPeter, line 77: Nice addition!

ArendPeter (approved):
Thanks for the fixes! The rename is very helpful

PR #95, opened 2022-04-24
Adding election state transitions and checks

mikefranze:
Checks to see if election state needs to update when retrieved.
Only allows edits while in draft state
Only allows ballots to be submitted in open state
Adds finalize button on voter home page (Doesn't check if owner of election though)

Updated tests to use id of correct test election, not a hard coded id. Should make it easier to add more tests.

### DIFF
--- a/backend/src/Controllers/elections.controllers.ts
+++ b/backend/src/Controllers/elections.controllers.ts
@@ -9,14 +9,44 @@ var ElectionsModel = new ElectionsDB();
 const getElectionByID = async (req: any, res: any, next: any) => {
     console.log(`-> elections.getElectionByID ${req.params.id}`)
     try {
-        const election = await ElectionsModel.getElectionByID(parseInt(req.params.id))
+        var election = await ElectionsModel.getElectionByID(parseInt(req.params.id))
         console.log(`get election ${req.params.id}`)
-        if (!election){
+        if (!election) {
             console.log('Error')
             return res.status('400').json({
                 error: "Election not found"
             })
         }
+        // Update Election State
+        const currentTime = new Date();
+        if (election.state === 'draft') {
+
+        }
+        if (election.state === 'finalized') {
+            if (election.start_time) {
+                console.log(`-> Election Transitioning To Open From ${election.state}`)
+                const startTime = new Date(election.start_time);
+                if (currentTime.getTime() > startTime.getTime()) {
+                    election.state = 'open';
+                    election = await ElectionsModel.updateElection(election)
+                }
+            } else {
+                console.log(`-> Election Transitioning To Open From ${election.state}`)
+                election.state = 'open';
+                election = await ElectionsModel.updateElection(election)
+            }
+        }
+        if (election.state === 'open') {
+            if (election.end_time) {
+                const endTime = new Date(election.end_time);
+                if (currentTime.getTime() > endTime.getTime()) {
+                    console.log('-> Election Transitioning To Closed')
+                    election.state = 'closed';
+                    election = await ElectionsModel.updateElection(election)
+                }
+            }
+        }
+
         req.election = election
         return next()
     } catch (err) {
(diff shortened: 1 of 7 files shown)


backend/src/Controllers/elections.controllers.ts
```diff
@@ -9,14 +9,44 @@ var ElectionsModel = new ElectionsDB();
...
+        const currentTime = new Date();
+        if (election.state === 'draft') {
+
+        }
+        if (election.state === 'finalized') {
+            if (election.start_time) {
+                console.log(`-> Election Transitioning To Open From ${election.state}`)
+                const startTime = new Date(election.start_time);
+                if (currentTime.getTime() > startTime.getTime()) {
+                    election.state = 'open';
```
ArendPeter, line 30: Looks like we're using a lazy initiazation approach to not update the election state until someone retrieves it? I like that approach for a lot of scenarios, but we may hit some issues with this particular case

1. The getElectionByID api name implies that it's only doing reads, but it performs writes as an unintended side effect
2. In frequent calls to getElectionByID would leave our database in an incorrect state (showing the finalized state when the election should really be in the open state), and I could see this leading to other bugs

Here are some options for fixing it, they're kind of tricky so this is definitely a task for later
 * Setup a process to update the database regardless of the customer traffic. Maybe a cron job? I'm not sure how this would work on heroku
 * Refactor our states to avoid time sensitive states. Here ``finalized``, ``open``, and ``closed`` would all be stored as ``finalized`` in the database, and we'd determine pending/open/closed on the fly based on the time

backend/src/Controllers/elections.controllers.ts
```diff
@@ -9,14 +9,44 @@ var ElectionsModel = new ElectionsDB();
...
+                election = await ElectionsModel.updateElection(election)
+            }
+        }
+        if (election.state === 'open') {
+            if (election.end_time) {
+                const endTime = new Date(election.end_time);
+                if (currentTime.getTime() > endTime.getTime()) {
+                    console.log('-> Election Transitioning To Closed')
+                    election.state = 'closed';
+                    election = await ElectionsModel.updateElection(election)
```
ArendPeter, line 45: nit: In theory ``updateElection()`` could get called twice if the first vote gets cast after the end period. It could be refactored as follows

```
if(election.state === 'finalized' or election.state === 'open' or election.state === 'closed'){
    if(election.end_time and currentTime.getTime() > endTime.getTime()){
       // become closed 
    }elif(election.start_time and currentTime.getTime() < startTime.getTime()){
      // become finalized
    }else{
      // become open
    }
}
```

backend/src/test/app.test.ts
```diff
@@ -191,11 +193,12 @@ describe("Email Roll", () => {
...
         // console.log(response.body)
         expect(response.statusCode).toBe(200)
-        // id = response.body.election.election_id
+        const responseObject = JSON.parse(response.body)
+        ID = responseObject.election.election_id
     })
     test("Get voter auth, is authorized and hasn't voted", async () => {
         const response = await request(app)
-            .post(`/API/Election/7/ballot`)
+            .post(`/API/Election/${ID}/ballot`)
```
ArendPeter, line 201: If I'm understanding this correctly, it looks like ID could be either 0 if "Create election, responds 200" wasn't previously called, or > 0 if "Create election, responds 200" was previously called. Could this create issues? 

Maybe "Create election, responds 200" should be part of beforeAll?

ArendPeter (approved):
Left some comments that might need to be addressed later, but they're not blocking

Thanks for making the update, nice work!

backend/src/Controllers/elections.controllers.ts
mikefranze: Yea I've been trying to figure out the best way to handle this for a while. I decided to just do the lazy check here and open an issue later for handling it better. There's a library node-cron that can handle timed events but I wasn't sure the best way to configure it in an express app. I was thinking every 5 minutes it could run a query for all finalized or open elections, update the state if needed and send any required emails. That could also be a good time to update election results rather than calculating them every time someone goes to the results page.

backend/src/Controllers/elections.controllers.ts
mikefranze: I think this would result in updating the election every time even if the state doesn't change. I added a check to only run the update once.

ArendPeter: ah good point

Thanks for adding the updates!

PR #98, opened 2022-05-14
Email interface and extra tabulators

mikefranze:
Adding basic sendgrid interface to allow us to send email invitations. 
Adding extra tabulators to experiment with. There's a lot of reused code from star tabulator, will probably try to refactor tabulators in the future.


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
ArendPeter, line 1: Shall we add .vscode/* to .gitignore?

ArendPeter (commented):

PR #101, opened 2022-05-30
Add feature to duplicate elections

### LINKED ISSUE #78: Add "Copy/Duplicate Election" feature.
With the new "Edit Election" feature added we should be able easily add a duplicate election feature. The flow would be

- User clicks duplicate election button, does not need to be original owner.
- Backend creates copies old election, changes name to "Copy of [Old Election Name]" changes owner ID to user's ID
- Redirects user to edit election page

ArendPeter:
fixes #78 

This allows you to duplicate existing elections


frontend/src/components/DuplicateElection.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const { id } = useParams();
+    let { data, setData, isPending, error } = useFetchSetData(`/API/Election/${id}`,{
+        method: 'GET',
+        headers: {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json',
+        },
+    })
+
+    // TODO: This approach for adding copyOf feels messy, let me know if there's a better way
```
mikefranze, line 18: I don't think we need to make changes to useFetch for this. We can instead create a new state dataCopy and use the useEffect hook to change that state when the data changes from the useFetch

``useEffect( () => { 
if(data && data.election ){
const newData = ...data;
newData.election.title = `Copy of ${newData .election.title}`
 setDataCopy(newData)
}
}
}, data)``

So when data changes it will trigger the rest of the code just once.

mikefranze (changes requested):

frontend/src/components/DuplicateElection.tsx
ArendPeter: Thanks, I'll give that a try!

ArendPeter: Sorry for the delay, I'm working on updating the approach, and here's what I've got so far

```
    const [data, setData] = useState(null);

    let { prevData, isPending, error } = useFetch(`/API/Election/${id}`,{
        method: 'GET',
        headers: {
            'Accept': 'application/json',
            'Content-Type': 'application/json',
        },
    })

    useEffect(
        () => {
            console.log("effect triggered");
            console.log(prevData);
            if(prevData && prevData.election){
                setData({...prevData, title: `Copy of ${prevData.election.title}`});
            } 
        }, 
        [prevData]
    )
    ```
    
    When I look at the console I see "effect triggered", and "undefined", any idea what I'm missing?

mikefranze: If you console.log(prevData) before the useEffect, do you see it switch from undefined?

mikefranze: I tried on the elections home page and saw that the effect was triggering on first render, then again when the data changed. Are you sure the fetch is working?

image

ArendPeter: Thanks! Looks like fetch was working, but useEffect wasn't triggering because prevData wasn't being updated

It seems to be breaking because I renamed the variable outputted from useFetch to prevData? So ``let { prevData, isPending, error } = useFetch`` breaks, but ``let { data, isPending, error } = useFetch`` works

I'll do some more testing, but if I need to use the name "data" then that works

mikefranze: Ah, that makes sense. Its trying to destructure the output of the fetch and prevData isn't a defined output. Try
`let { data: prevData, isPending, error } = useFetch`
to rename while destructuring.

ArendPeter: Awesome, that fixed it

mikefranze: Do we still need the changes to useFetch?

ArendPeter: mb, fixed!

mikefranze (approved):

PR #102, opened 2022-05-31
Updating results page

mikefranze:
Added round by round data to tabulator output
Updated results page to show table of scores, round runoff results with highlights
Added toggle to view preference matrix


ArendPeter (approved):
Looks good!

PR #104, opened 2022-06-18
Feat/logger

scottPlusPlus:
1. Adds a "contextId" to the Express Request object (as random uid).  Ideally this will be appended to all logs from that request.  We can return this with any error messages if we then need to debug the request-flow.
2. Adds an initial Logger as a wrapper around console.log

If this looks good I'll replace all the console.logs to use the logger, and work on replacing the implementation with Winston and/or DataDog or somesuch


ArendPeter (approved):
Looks good to me!

Although I'm curious why you opted for this approach rather than using a preexisting logging library?

mikefranze:
> Looks good to me!
> 
> Although I'm curious why you opted for this approach rather than using a preexisting logging library?

It looks like with this approach we can easily add in the preexisting logging library by just editing one Logger.ts file.

scottPlusPlus:
> > Although I'm curious why you opted for this approach rather than using a preexisting logging library?
> 
> It looks like with this approach we can easily add in the preexisting logging library by just editing one Logger.ts file.

Yes exactly, just abstracting so we can swap out implementation if needed.

PR #107, opened 2022-06-20
Feat/logger

scottPlusPlus:
1.  Adding ServiceLocator for making Postgres fetching more DRY.
2.  Moving a laughably small amount of console.logs to new logger.


mikefranze:
Looks good, thanks for cleaning that up

Issue #63, opened 2022-03-07
Use encryption methods from STAR-Vote (Secure, Transparent, Auditable, Reliable)


ArendPeter:
We've had several discussion on how to setup our databases in a way that maintains user privacy, and I thought looking into STAR-Vote (Secure, Transparent, Auditable, Reliable) could be relevant as a case study

# What is STAR-Vote (Secure, Transparent, Auditable, Reliable)?

The STAR-Vote project (same name, completely different organization), focuses on how to manage ballots instead of the voting method
https://www.youtube.com/watch?v=Cl8q_356zZs (description starts at 34:23)

Their method has come up a few times in Slack conversations, and it could be a good opportunity to collaborate and take advantage of the naming overlap
https://starvoting.slack.com/archives/C9U6425CM/p1635474100035800
https://starvoting.slack.com/archives/C01EBAT283H/p1616783581154400?thread_ts=1616684618.142400&cid=C01EBAT283H

Here's a case study of how their STAR-Vote system works, and which pieces can apply to our online system

# How does it work?

Keep

![image](https://user-images.githubusercontent.com/9289903/156949309-582508ed-9cc7-404b-815a-f64e310059b3.png)

TLDR: The vote ends up getting stored as a physical ballot, and an encrypted version on the voting terminals. These versions can be compared later for auditing

## Registration

The voter goes to the registration system, here their voting status is validated, and their cross off the online database

This is the ONLY ONLINE step, the rest of the steps are done on an offline network. The user gets a receipt (with a precinct id and a bar code) to be used in the following steps which validates that they were registered. The receipt only encodes the precinct, no other voter specific information is included

> **Voting Id**: At this point they also checked the voter's id, but this is specific to the travis county election rules rather than the STAR system it's self

> **Why precinct id?**: Travis County allows voters to choose any voting station rather than being limited to their precinct, so this information is important for the later steps

## Auth

At the controller machine, the voter delivers their receipt and a volunteer scans it, gives them a one-time use printed auth code

## Voting Terminal

There the voter will input the code, and begin voting. Once they've finished a encrypted version of the vote is stored on the machine, and a printed copy is made for the voter

## Ballot box

The voter submits the printed ballot to the ballot box, the vote is not officially counted until it's been stored in the ballot box

The ballot box scans the bar code on the ballot, and it can communicate with the voting terminals. If a duplicate ballot is given the ballot box will reject it

## Ballot Receipt

[...]

PR #108, opened 2022-06-20
Feat/ballot history

scottPlusPlus:
Adds "history" to the Ballot object.
Changes the ballot date to be server-authoritative
Updates vote request to return the ballot with its id

Tested locally:

firefox_bWOE13E1vr
firefox_JoMMiIJXJl

NOTE:  I will suggest changing the ballot ID to be random, so voters don't know which order they voted in.  That may require a DB migration, so we may want to collect all the migrations we want to do and then organize that.

### DIFF
--- a/backend/src/Models/Ballots.ts
+++ b/backend/src/Models/Ballots.ts
@@ -1,4 +1,5 @@
 import { Ballot } from '../../../domain_model/Ballot';
+import Logger from '../Services/Logging/Logger';
 const { Pool } = require('pg');
 
 class BallotsDB {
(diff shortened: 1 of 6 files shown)


backend/src/Models/Ballots.ts
```diff
@@ -34,20 +34,27 @@ class BallotsDB {
...
             ip_address      VARCHAR, 
-            votes           json NOT NULL
+            votes           json NOT NULL,
+            history         json
           );
         `;
         console.log(query);
         var p = this._postgresClient.query(query);
         return p.then((_: any) => {
+            return 'foo';
```
ArendPeter, line 44: Looks like this was some debug code that was left in?

ArendPeter (approved):

ArendPeter:
Random ids sound good for the ballots

Regarding exposing the order, this could come up again if we add hash chaining down the road (discussed a bit here https://github.com/Equal-Vote/star-server/issues/63 ). Maybe enabling the hash chain feature would also require that the election admin enables other features to anonymize the votes?

mikefranze:
I wouldn't worry about DB migrations, we don't have any important data yet so we can reset the database often.

backend/src/Models/Ballots.ts
scottPlusPlus: Lol whoops, will fix.  Great catch!

PR #109, opened 2022-06-29
Feat/logger

scottPlusPlus:
Did a pass for all the log statements on Elections Controller.  (will go through each file in separate diffs).
Cleaned up the logger a bit so multiple logs from a request now look like :

REQUEST: POST /API/Sandbox @ 2022-06-29T04:43:24.686Z ip:127.0.0.1  ctx:66195d4e
. . Elections.Controllers.getSandboxResults  ctx:66195d4e
. . Candidate names:  ["A","B","C","D","E"]  ctx:66195d4e

(added indentation for readability)
added responseErr util function for DRY

### DIFF
--- a/backend/src/Controllers/elections.controllers.ts
+++ b/backend/src/Controllers/elections.controllers.ts
@@ -2,21 +2,23 @@ import { Election } from '../../../domain_model/Election';
 import { Ballot } from '../../../domain_model/Ballot';
 import { Score } from '../../../domain_model/Score';
 import ServiceLocator from '../ServiceLocator';
+import Logger from '../Services/Logging/Logger';
+import { responseErr } from '../Util';
 const ElectionsDB = require('../Models/Elections')
 const StarResults = require('../Tabulators/StarResults.js');
 
 var ElectionsModel = new ElectionsDB(ServiceLocator.postgres());
+var className="Elections.Controllers";
 
 const getElectionByID = async (req: any, res: any, next: any) => {
-    console.log(`-> elections.getElectionByID ${req.params.id}`)
+    Logger.info(req, `${className}.getElectionByID ${req.params.id}`);
     try {
         var election = await ElectionsModel.getElectionByID(parseInt(req.params.id))
-        console.log(`get election ${req.params.id}`)
+        Logger.debug(req, `get election ${req.params.id}`);
+        var failMsg = "Election not found";
         if (!election) {
-            console.log('Error')
-            return res.status('400').json({
-                error: "Election not found"
-            })
+            Logger.info(req, `${failMsg} electionId=${req.params.id}}`);
+            return responseErr(res, req, 400, failMsg);
         }
         // Update Election State
         const currentTime = new Date();
@@ -25,15 +27,17 @@ const getElectionByID = async (req: any, res: any, next: any) => {
 
         }
         if (election.state === 'finalized') {
+            var openElection = false;
             if (election.start_time) {
                 const startTime = new Date(election.start_time);
                 if (currentTime.getTime() > startTime.getTime()) {
-                    console.log(`-> Election Transitioning To Open From ${election.state}`)
-                    stateChange = true;
-                    election.state = 'open';
+                    openElection = true;
                 }
             } else {
-                console.log(`-> Election Transitioning To Open From ${election.state}`)
+                openElection = true;
+            }
+            if (openElection){
+                Logger.info(req, `Election Transitioning to Open From ${election.state} (start time = ${election.start_time})`);
                 stateChange = true;
                 election.state = 'open';
             }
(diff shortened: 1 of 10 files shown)


backend/src/Controllers/elections.controllers.ts
```diff
@@ -2,21 +2,23 @@ import { Election } from '../../../domain_model/Election';
 import { Ballot } from '../../../domain_model/Ballot';
 import { Score } from '../../../domain_model/Score';
 import ServiceLocator from '../ServiceLocator';
+import Logger from '../Services/Logging/Logger';
+import { responseErr } from '../Util';
 const ElectionsDB = require('../Models/Elections')
 const StarResults = require('../Tabulators/StarResults.js');
 
 var ElectionsModel = new ElectionsDB(ServiceLocator.postgres());
+var className="Elections.Controllers";
```
ArendPeter, line 11: Instead of a className variable, could we retrieve the filename dynamically and then bake that logic into our logging library?

https://stackoverflow.com/questions/14201475/node-js-getting-current-filename

backend/src/Controllers/elections.controllers.ts
```diff
@@ -89,128 +93,120 @@ const getElectionResults = async (req: any, res: any, next: any) => {
...
-            return res.status('400').json({
-                error: "Elections not found"
-            })
-        res.json(Elections)
-    } catch (err) {
-        return res.status('400').json({
-            error: "Could not retrieve elections"
-        })
+        if (!Elections){
+            Logger.error(req, failMsg);
```
ArendPeter, line 121: I think the more specific Election DNE message could be helpful to the user

backend/src/Controllers/elections.controllers.ts
```diff
@@ -89,128 +93,120 @@ const getElectionResults = async (req: any, res: any, next: any) => {
...
+            return responseErr(res, req, 400, failMsg);
+        }
+        res.json(Elections);
+    } catch (err:any) {
+        Logger.error(req, failMsg+". "+err.message);
+        return responseErr(res, req, 400, failMsg);
```
ArendPeter, line 127: Maybe we can make this a 500 error? I think it only happens if something goes wrong on our end

backend/src/Services/Logging/LoggerImpl.ts
```diff
@@ -8,33 +8,42 @@ export class LoggerImpl {
...
     }
 
     info(context?:ILoggingContext,  message?: any, ...optionalParams: any[]):void{
-        this.log(context, "info", message, ...optionalParams);
+        this.log(context, "", message, ...optionalParams);
     }
 
     warn(context?:ILoggingContext,  message?: any, ...optionalParams: any[]):void{
-        this.log(context, "WARN", message, ...optionalParams);
+        this.log(context, "WARN ", message, ...optionalParams);
```
ArendPeter, line 19: nitpick: why the added space?

backend/src/Services/Logging/LoggerImpl.ts
```diff
@@ -8,33 +8,42 @@ export class LoggerImpl {
...
     log(context?:ILoggingContext, levelStr?:string,  message?: any, ...optionalParams: any[]):void { 
-        var msg = new Date(Date.now()).toISOString();
+        var msg = "";
+        var lvlStr = "";
+        var ctxStr = "";
+        var prefix = "";
         if (context!= null){
-            if (context.contextId == null || context.contextId == undefined){
-                context.contextId = randomUUID();
+            ctxStr = "  ctx:" + context.contextId;
```
ArendPeter, line 33: Good idea adding the context id at the end! Looks like it will be useful for filtering

package.json
```diff
@@ -18,5 +18,8 @@
   "bugs": {
     "url": "https://github.com/Equal-Vote/star-server/issues"
   },
-  "homepage": "https://github.com/Equal-Vote/star-server#readme"
+  "homepage": "https://github.com/Equal-Vote/star-server#readme",
+  "dependencies": {
+    "crypto": "^1.0.1"
```
ArendPeter, line 23: Is this being used?

ArendPeter (approved):

package.json
scottPlusPlus: it's being used in IRequest.ts
not sure how it was working before without the import?  (could have been a transient dependency anyway??)

backend/src/Services/Logging/LoggerImpl.ts
scottPlusPlus: makes everything line up in monospace fonts :)

backend/src/Controllers/elections.controllers.ts
scottPlusPlus: AFAIK, using __filename in the logging library would just return the LoggerImpl for everything, which is not helpful.

I've done what you're suggesting in other languages, so I assume it can be done.  A first glance at TS made it seem complicated so we can add that as a "nice to have"?

mikefranze:
Looks good

PR #111, opened 2022-07-01
Feat/logger 2

scottPlusPlus:
This should be merged AFTER feat/logger.

Wanted to separate this out onto separate diff for easier digestion...


mikefranze (approved):
Looks good, thanks for going through all these

PR #112, opened 2022-07-03
Admin, election roll, frontend structure updates

mikefranze:
Lots of little changes in this. Will probably have conflicts with Scott's PRs but can resolve that later. Just wanted to create this PR before I start adding even more to it.

-Adds credentialer role which has limited permissions to approve and flag election rolls.
-Adds history and state to election role
-Adds logic for how election role state can change. This is probably not the cleanest implementation so I'm open to other suggestions
-Moved routes of /Election/:id/* from main app to new election component. The new component does a single fetch for election data and flows it into subpages.
-Moved authSession code from main app to its own file
-Updated useFetch to output an async function that can be called to make the request, rather than call it on its own. The reason for that is so we can better control when it is called and can pass the function into subcomponents to be called there. For example, the request to fetch election data can be passed into the VoterAuth component so it can refetch the election data after voter id is entered. The function also returns true if the response is ok, false if not, for cases where we want to check before automatically proceeding with something like navigating to another page.
-Adds /Election/:id/Admin page that shows tables of election rolls and histories. Not fully fleshed out yet.
-Adds placeholder for sidebar menu.

### DIFF
--- a/backend/src/Controllers/voterRolls.controller.ts
+++ b/backend/src/Controllers/voterRolls.controller.ts
@@ -3,6 +3,7 @@ import ServiceLocator from "../ServiceLocator";
 const ElectionRollDB = require('../Models/ElectionRolls')
 const EmailService = require('../Services/EmailService')
 var ElectionRollModel = new ElectionRollDB(ServiceLocator.postgres());
+const { permissions, hasPermission } = require('../auth/permissions')
 
 const getRollsByElectionID = async (req: any, res: any, next: any) => {
     console.log(`-> electionRolls.getRollsByElectionID ${req.election.election_id}`)
@@ -34,7 +35,13 @@ const addElectionRoll = async (req: any, res: any, next: any) => {
 
     try {
         // console.log(req)
-        const NewElectionRoll = await ElectionRollModel.submitElectionRoll(req.election.election_id, req.body.VoterIDList, false)
+        
+        const history = [{
+            action_type: 'added',
+            actor: req.user.email,
+            timestamp: Date.now(),
+        }]
+        const NewElectionRoll = await ElectionRollModel.submitElectionRoll(req.election.election_id, req.body.VoterIDList, false, 'added', history)
         if (!NewElectionRoll)
             return res.status('400').json({
                 error: "Voter Roll not found"
@@ -53,6 +60,14 @@ const editElectionRoll = async (req: any, res: any, next: any) => {
     console.log(`-> electionRolls.editElectionRoll ${req.election.election_id}`)
 
     try {
+        if (req.body.electionRollEntry.history == null){
+            req.body.electionRollEntry.history = [];
+        }
+        req.body.electionRollEntry.history.push([{
+            action_type: 'edited',
+            actor: req.user.email,
+            timestamp: Date.now(),
+        }])
         const electionRollEntry = await ElectionRollModel.update(req.body.electionRollEntry)
         if (!electionRollEntry)
             return res.status('400').json({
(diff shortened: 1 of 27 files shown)


domain_model/ElectionRoll.ts
```diff
@@ -6,4 +6,14 @@ export interface ElectionRoll {
     voter_id: Uid; //ID of voter who cast ballot
     ballot_id?:  Uid; //ID of ballot, unsure if this is needed
     submitted: boolean; //has ballot been submitted
-}
```
scottPlusPlus, line 9: Could we add a comment to ElectionRoll just to clarify what it's for?

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -69,13 +84,78 @@ const editElectionRoll = async (req: any, res: any, next: any) => {
     }
 }
 
+const changeElectionRollState = async (req: any, res: any, next: any) => {
+    console.log(`-> electionRolls.changeElectionRollState ${req.election.election_id}`)
```
scottPlusPlus, line 88: can we start using Logger.info from now on?

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -69,13 +84,78 @@ const editElectionRoll = async (req: any, res: any, next: any) => {
...
+
```
scottPlusPlus, line 89: I'd suggest a public method / API for each separate flow, based on the action being taken:
Approve Voter, Flag Voter, Reject Voter, etc,  to increase code maintainability and audit-clarity.

These can rely on a private method for updating the roll-state in the DB

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -160,7 +240,12 @@ const getVoterAuth = async (req: any, res: any, next: any) => {
         req.authorized_voter = true;
         if (req.electionRollEntry.length == 0) {
             //Adds voter to roll if they aren't currently
-            const NewElectionRoll = await ElectionRollModel.submitElectionRoll(req.election.election_id, [req.voter_id], false)
+            const history = [{
+                action_type: 'added',
+                actor: req.user?.email || req.voter_id,
```
scottPlusPlus, line 245: I'd suggest relying on ids rather than email.  Theoretically we'd give users ability to change their emails, which would then break our data.

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -69,13 +84,78 @@ const editElectionRoll = async (req: any, res: any, next: any) => {
...
+        const updatedEntry = await ElectionRollModel.update(req.electionRollEntry)
+        if (!updatedEntry)
+            return res.status('400').json({
+                error: "Voter Roll not found"
+            })
+        req.electionRollEntry = updatedEntry
+        res.status('200').json()
+    } catch (err) {
+        console.log(err)
+        return res.status('400').json({
```
scottPlusPlus, line 145: general rule:  if it's in a catch, it's probably 500 :P

domain_model/ElectionRoll.ts
```diff
@@ -6,4 +6,14 @@ export interface ElectionRoll {
...
\ No newline at end of file
+    state: string;
```
scottPlusPlus, line 9: Can we go ahead and define the State as an enum?

backend/src/Controllers/voterRolls.controller.ts
mikefranze: That's a good point, the only issue is we'd want to display the user's email so we'd need some way to look that up.

domain_model/ElectionRoll.ts
mikefranze: Added in latest commit

backend/src/Controllers/voterRolls.controller.ts
mikefranze: Added separate api for each flow, but kept just a single middleware for now.

backend/src/auth/permissions.ts
```diff
@@ -18,4 +21,12 @@ const permissions = {
   canEditElectionState:     [roles.system_admin, roles.owner],
   canViewPreliminaryResults:[roles.system_admin, roles.owner, roles.admin, roles.auditor],
 }
-module.exports = permissions
\ No newline at end of file
+
+const hasPermission = (roles:string[],permission:string[]) => {
```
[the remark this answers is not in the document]
mikefranze: I should define the interface for this so it's clear. This only checks one permission, each permission is an array of roles with that permission. I suppose you could concatenate multiple permissions together and check them all.

backend/src/Controllers/voterRolls.controller.ts
mikefranze: Merged and added Logger

mikefranze:
Anything else on this or should I merge it?

scottPlusPlus (approved):
LGTM

PR #115, opened 2022-07-09
Feat/logger state pass

scottPlusPlus:
(based on top feat/logger-3)

Adds logging for all state changes.  Adds a new argument to all methods that mutate state, requiring a "reason".  With this in New Relic we can query for "STATE:" and see history of all state changes with some explanation.

Some seemed a little ambiguous ("User adding Election Roll") which suggests maybe we need to add more explicitness to some methods.


ArendPeter (approved):
LGTM!

mikefranze:
looks good, just need to clean up conflicts.

mikefranze (changes requested):
Needs conflicts resolved.

PR #117, opened 2022-07-13
Feat/testing

scottPlusPlus:
Doing a pass on testing.

This pass separates out some of the tests to their own files, and gets most tests passing.

Changes made NOT in the test:
1. addElectionRoll which was previously returning a doubly-serialized result.
2. editElection was taking the ID from the req.body.election, and ignoring the URL param.  Added check for that.
3. Made editElection return the updated election.
4. Mock ElectionRoll had a bug in update method

All tests now passing EXCEPT testing invalid data returning improper response.  Will work on creating / adding some validators for input, but not in this diff.


mikefranze:
Thanks for updating tests, separating into different files was a good call. Just need to fix conflict

PR #118, opened 2022-07-15
Feat/type validators

scottPlusPlus:
land on top of feat/testing

All tests passing like a BOSS.  Validators are half-assed, but not critical atm and make for an awesome "first issue" :p


mikefranze:
Nice! Should note the validators will probably only work in the backend. The frontend compiler complains if it imports anything outside /src that isn't a interface or type.

mikefranze:
Looks good

PR #121, opened 2022-07-17
Feat/test utils

scottPlusPlus:
Adding TestHelper to (hopefully) make testing more convenient.

So far only applied to createElection and editElection tests.

land atop feat/mock-db-serialization


backend/src/Services/Logging/TestLoggerImpl.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export class TestLoggerImpl {
+
+    logs:string[] = [];
+
+    constructor() {
+    }
+
+    debug(context?:ILoggingContext,  message?: any, ...optionalParams: any[]):void{
+        this.log(context, "", message, ...optionalParams);
```
ArendPeter, line 14: Should debug and info also have the "DEBUG" and "INFO" prefixes?

backend/src/test/createElection.test.ts
```diff
@@ -1,71 +1,72 @@
...
 
 afterEach(() => {
-    jest.clearAllMocks();
+  jest.clearAllMocks();
+  th.afterEach();
 });
 
 describe("Create Election", () => {
+  describe("Good Election data provided", () => {
+    var electionRes: Election;
```
ArendPeter, line 23: Nice! This looks a lot cleaner

ArendPeter (approved):
Looks good!

backend/src/test/createElection.test.ts
mikefranze: Agreed!

backend/src/Services/Logging/TestLoggerImpl.ts
scottPlusPlus: I just copied what's in the normal logger.

My general thought was that the debug and info logs would be the "normal" ones.  We would want warn and errors to stand out more.

However, if we're using new relic to filter things anyway, may make sense to add the prefix?

PR #123, opened 2022-07-18
Registration endpoint, sidebar, voter roll page

mikefranze:
Added registration endpoint that allows a voter to be submitted to election roll in registered state. Also added registration data to election roll db for the future. 

Added very basic sidebar, only appears if you have a role in the election. 

Added page to view voter rolls and approve, flag, invalidate them. Need to add logic to make these permission based.


scottPlusPlus (changes requested):
1. Can we get a test that demonstrates the functionality?

2. I'm a little lost as to what "registration" information is in this case?  What type of data are we expecting?

backend/src/Models/ElectionRolls.ts
```diff
@@ -44,14 +45,15 @@ class ElectionRollDB {
     }
 
 
-    submitElectionRoll(election_id: number, voter_ids: string[], submitted: Boolean, state: string, history: ElectionRollAction): Promise<boolean> {
+    submitElectionRoll(election_id: number, voter_ids: string[], submitted: Boolean, state: string, history: ElectionRollAction, registration: any): Promise<boolean> {
```
scottPlusPlus: I'm assuming the option to pass multiple voter_ids is ONLY used when setting up the election?

If so would consider breaking that into a separate method on the interface.

frontend/src/components/EditElectionRoll.tsx
```diff
@@ -6,15 +6,47 @@ import Typography from '@material-ui/core/Typography';
...
 import Container from '@material-ui/core/Container';
 import { Paper, Table, TableBody, TableCell, TableContainer, TableHead, TableRow, Tooltip } from "@material-ui/core";
+import useFetch from "../useFetch";
+import { ElectionRollState } from "../../../domain_model/ElectionRoll";
 
-const EditElectionRoll = ({ roll, onClose }) => {
+const EditElectionRoll = ({ roll, onClose, fetchRolls, id}) => {
     const [updatedRoll, setUpdatedRoll] = useState(roll)
+
+    const approve = useFetch(`/API/Election/${id}/rolls/approve`,'post')
```
scottPlusPlus, line 15: don't need to address for this diff but:

do we have (or want) an option for a voter to see whether or not they've been approved yet?  (and thus actually counted)?

frontend/src/components/EditElectionRoll.tsx
```diff
@@ -6,15 +6,47 @@ import Typography from '@material-ui/core/Typography';
...
+    const flag = useFetch(`/API/Election/${id}/rolls/flag`,'post')
+    const unflag = useFetch(`/API/Election/${id}/rolls/unflag`,'post')
+    const invalidate = useFetch(`/API/Election/${id}/rolls/invalidate`,'post')
+
+    const onApprove = async () => {
```
scottPlusPlus, line 20: theoretically can combine these into a single "request", with a type (validate, flag, unflag, etc) that determines the URL.  Then elsewhere in the page don't need to check for 4 different responses or errors, just of the active "request"

frontend/src/components/EditElectionRoll.tsx
mikefranze: Yes I think so. When the voter goes to the election home page it currently says if they've voted, we can add additional info there.

backend/src/Models/ElectionRolls.ts
mikefranze: Yea we really only do multiple when they're added in batch by the election host when setting up. Adding a separate method would be better for adding just one. Or I can refactor to input an array of the election rolls and move the mapping of voter ids to the controller.

mikefranze:
> 1. Can we get a test that demonstrates the functionality?
> 2. I'm a little lost as to what "registration" information is in this case?  What type of data are we expecting?

Yea I'll add a test. The registration is mostly just a placeholder right now. I'm thinking the election settings would hold info on the format of the registration data, like Name, Address, Zip Code, and API to send the data to like the registered voter database. Then the users would be prompted to enter Name, Address, etc and the data would be stored in the registration field. The credentialer would also be able to view this registration info and approve.

mikefranze:
Tests added

backend/src/Models/ElectionRolls.ts
mikefranze: Updated to just pass in list of voter rolls.

backend/src/test/testInputs.ts
```diff
@@ -51,6 +51,7 @@ export default {
         state: 'open',
         frontend_url: '',
         owner_id: '0',
+        credential_ids: ['Alice@email.com'],
```
scottPlusPlus, line 54: is this used somewhere?

scottPlusPlus (approved):
thanks for adding the test.

eventually will want to show that votes that are not approved will not count, and that an official could approve a registration before of after the user votes.  and ideally voter will have way to see if they are approved yet or not.

but don't want to hold a diff hostage because not everything is sorted yet, we can add eventually.

mikefranze:
> thanks for adding the test.
> 
> eventually will want to show that votes that are not approved will not count, and that an official could approve a registration before of after the user votes. and ideally voter will have way to see if they are approved yet or not.
> 
> but don't want to hold a diff hostage because not everything is sorted yet, we can add eventually.

Correct, and also need to add that actual logic lol. But I think it would make sense to wait until we do some general refactoring first.

backend/src/Models/ElectionRolls.ts
```diff
@@ -106,13 +108,13 @@ export default class ElectionRollDB {
 
     update(election_roll: ElectionRoll, ctx:ILoggingContext, reason:string): Promise<ElectionRoll | null> {
         Logger.debug(ctx, `ElectionRollDB.updateRoll`);
-        var sqlString = `UPDATE ${this._tableName} SET ballot_id=$1, submitted=$2, state=$3, history=$4  WHERE election_id = $5 AND voter_id=$6`;
+        var sqlString = `UPDATE ${this._tableName} SET ballot_id=$1, submitted=$2, state=$3, history=$4, registration=$5 WHERE election_id = $6 AND voter_id=$7`;
```
ArendPeter, line 111: This is outside the scope of this PR, but we should probably think about how to handle concurrent modifications down the line

Let's say there are currently 3 events in the history, and 2 clients try adding a 4th event at the same time. If the flow happened as follows, the second client could overwrite the addition on the first client added

Example
```
Client1: Downloads history (w/ 3 elements) 
Client2: Downloads history (same 3 elements)
Client1: Locally adds 4th event
Client2: Locally adds 4th event
Client1: Uploads their modified history
Client2: Uploads their modified history (thus overwriting client1's addition)
```

I've used optimistic locking ([explainer vid](https://www.youtube.com/watch?v=VxGKvqHhU5c)) for this in the past, but we can discuss other options

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -44,7 +44,14 @@ const addElectionRoll = async (req: any, res: any, next: any) => {
             actor: req.user.email,
             timestamp: Date.now(),
         }]
-        const newElectionRoll = await ElectionRollModel.submitElectionRoll(req.election.election_id, req.body.VoterIDList, false, ElectionRollState.approved, history, req, `User adding Election Roll??`);
+        const rolls: ElectionRoll[] = req.body.VoterIDList.map((id:string) => ({
+            election_id: req.election.election_id,
+            voter_id: id,
+            submitted: false,
+            state: ElectionRollState.approved,
+            history: history,
```
ArendPeter, line 52: How we store the history list could be worth more discussion. 

Currently looks like we're storing a separate history list for each user in the roll list (1 list per user/election)? Here's a list of some other options

- 1 list for everything (a global history)
- 1 list per election
- 1 list per election server (for the future when our architecture spreads the load across multiple servers)
- 1 list per ballot
- 1 list per voter roll (current approach)

I haven't decided which approach I like best yet. "1 list per election" will make the auditing process cleaner (otherwise we're combing a bunch of smaller lists), but it also makes concurrency issues more likely. "1 list per election tenant" will be ideal for the hashing, and "1 list per voter roll" approach is ideal for voters looking up their ballot info

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -60,7 +67,42 @@ const addElectionRoll = async (req: any, res: any, next: any) => {
         return responseErr(res, req, 500, msg);
     }
 }
-
+const registerVoter = async (req: any, res: any, next: any) => {
```
ArendPeter, line 70: In the future I could see a bulkRegisterVoter being handy (for example, if IPO wanted to use this API to register their full voter list)

ArendPeter (approved):
Looks great! All of my comments are more for future PRs so no changes needed

backend/src/Controllers/voterRolls.controller.ts
scottPlusPlus: Could you elaborate what you mean that "1 list per election will make auditing process cleaner?"

I'm imagining the two most common questions would be:
* is this voter approved
* why is this voter approved / not approved

which seem to be served by history per voter per election
(I'm also seeing "1 history per voter per election" the same as "1 list per ballot", but if there's a distinction here please LMK)

mikefranze: I think the use for 1 list per election for an auditor to easily grab all the data and see for example all the voters that were rejected. You might also want voters to be able to see this data (with any sensitive data removed). But it should be easy enough to query all the histories and concatenate them if needed. 

My concern with 1 list is that it makes your concurrent modifications concern more likely. 

Another approach is we have a separate table for histories and just have 1 row per modification, but that sounds like overkill.

ArendPeter: Sure thing

| Could you elaborate what you mean that "1 list per election will make auditing process cleaner?"

I was thinking the auditor would want to see a full log of everything that happened in the election in one place

| I'm also seeing "1 history per voter per election" the same as "1 list per ballot", but if there's a distinction here please LMK

That one is more a question of where we store the history. There are actions like "vote counted", or "vote rejected", that are initiated by an admin rather than the voter, so maybe they make more sense attached to the ballot rather than the voter? Then again, the registration history would exist before the ballot, so maybe roll is the better option between the 2?

mikefranze: I'm thinking voter roll history would have things like registration or changing state, ballot history would have things like if the ballot is invalidated or changed because the voter resubmitted or invalidated by an admin. I think we'll need to come up with a list of reasons why any of these would change/need history data and go from there.

scottPlusPlus: Ah, great thing to clarify:  are there reasons a vote would be rejected other than having to do with the voter registration?  Something about the ballot itself rather than about the voter?

What all information are we trying to capture in these histories?

For ballot, I assumed when the voter votes, and if they change their ballot.

For the registration, would want to see when the initial registration happened, when registration info is uploaded / changed (ex voter id etc)., as well as when approving / rejecting the voter.

What other events we need to capture?

backend/src/Controllers/voterRolls.controller.ts
mikefranze: We sort of have that with adding the voter id list. Registering marks them as registered which would then need approval. But if IPO is adding them then they would already be approved?

PR #126, opened 2022-07-27
using Service Locator to provide all DBs.  

scottPlusPlus:
using Service Locator to provide all DBs. 
Will be needed for controllers that interact with multiple DBs (so they are all working on the same DB instance)


ArendPeter (approved):
Nice work!

mikefranze (approved):
Looks good, thanks!

PR #127, opened 2022-07-28
Feat/controller refactor

scottPlusPlus:
Example of refactoring a controller into a single function  (rather than chained handlers)

Please note "//TODO - not sure what the intended response was here from voterRolls.addElectionRoll ??"
Wasn't entirely sure what create election was meant to return??  LMK and I'll fix that.

Also note this would need to be further fixed so that we write the Election and Roll to the DB atomically, but want to handle that after we have all the data that needs to be handled collected into a single flow.

### DIFF
--- a/backend/src/Controllers/controllerUtils.ts
+++ b/backend/src/Controllers/controllerUtils.ts
@@ -1,0 +1,39 @@
+import { IRequest, reqIdSuffix } from "../IRequest"
+import { Unauthorized } from "@curveball/http-errors/dist"
+import { Election, electionValidation } from "../../../domain_model/Election";
+import Logger from "../Services/Logging/Logger"
+import { BadRequest, InternalServerError } from "@curveball/http-errors";
+import { Request, Response } from 'express';
+
+var jwt = require('jsonwebtoken')
+
+export function expectUserFromRequest(req:IRequest ):any {
+    var user = jwt.decode(req.cookies.id_token);
+    if (user == null || user == {}){
+      throw new Unauthorized();
+    }
+    return user;
+  }
+
+export function expectValidElectionFromRequest(req:IRequest):Election {
+    const inputElection = req.body.Election;
+    const validationErr = electionValidation(inputElection);
+    if (validationErr) {
+        Logger.info(req, "Invalid Election: " + validationErr, inputElection);
+        throw new BadRequest("Invalid Election");
+    }
+    return inputElection;
+}
+
+export function catchAndRespondError(req:IRequest, res:Response, err:any):Response<any, Record<string, any>> {
+    var status = 500;
+    if (err.httpStatus) {
+        status = err.httpStatus;
+    }
+    var msg = "Error";
+    if (err.detail) {
+        msg = err.detail;
+    }
+    msg += reqIdSuffix(req);
+    return res.status(status).json({error:msg});
+}
--- a/backend/src/Routes/elections.routes.js
+++ b/backend/src/Routes/elections.routes.js
@@ -1,10 +1,12 @@
 const express = require('express');
 const router = express.Router();
 
+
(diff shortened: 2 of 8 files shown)


mikefranze:
> Example of refactoring a controller into a single function (rather than chained handlers)
Thanks Scott, this makes sense. Would there be different controllers for each route? 
> Please note "//TODO - not sure what the intended response was here from voterRolls.addElectionRoll ??" Wasn't entirely sure what create election was meant to return?? LMK and I'll fix that.
I'm thinking it would return the created election. The frontend can then use the ID to route to that election's home page. Also, I'm thinking we should separate the add election roll to a different route and make it another step in the election creation process. 
> Also note this would need to be further fixed so that we write the Election and Roll to the DB atomically, but want to handle that after we have all the data that needs to be handled collected into a single flow.
As I said above this might be split into two steps, but there's other areas where we need to write to two tables at the same time.

scottPlusPlus:
Could you expand on wanting a separate route for election roll?

mikefranze:
> Could you expand on wanting a separate route for election roll?

So we actually already have a post route "/Election/:id/rolls/" for adding rolls after the election is created. My thinking is that it might be easier on the user to submit one thing at a time instead of having to fill out everything at the start. Plus, because we want the ability to add rolls after election creation anyway, we might as well have that be one flow.

mikefranze (approved):

backend/src/Controllers/controllerUtils.ts
```diff
@@ -0,0 +1,39 @@
+import { IRequest, reqIdSuffix } from "../IRequest"
+import { Unauthorized } from "@curveball/http-errors/dist"
+import { Election, electionValidation } from "../../../domain_model/Election";
+import Logger from "../Services/Logging/Logger"
+import { BadRequest, InternalServerError } from "@curveball/http-errors";
+import { Request, Response } from 'express';
+
+var jwt = require('jsonwebtoken')
+
+export function expectUserFromRequest(req:IRequest ):any {
```
mikefranze, line 10: This should be a few functions I think. One to get user, one to expect user, and one to get user_auth. auth.controllers.ts has getUser which gets the user and user_auth, and has isLoggedIn. 

getUser (minus the part to get user_auth) could probably just be middleware that we add to the main app on all requests.

backend/src/Routes/elections.routes.js
```diff
@@ -67,10 +69,9 @@ router.get('/Elections',
     authController.getUser,
     electionController.getElections)
 router.post('/Elections/',
-    authController.getUser,
-    authController.isLoggedIn,
-    electionController.createElection,
```
ArendPeter, line 72: Should createElection() be removed from electionController now that it's implemented by createElectionController.createElectionController?

ArendPeter:
| Example of refactoring a controller into a single function (rather than chained handlers)

Could you clarify the rational for the single function approach vs the chained handlers? Is single function easier to maintain?

ArendPeter (approved):

mikefranze:
> | Example of refactoring a controller into a single function (rather than chained handlers)
> 
> Could you clarify the rational for the single function approach vs the chained handlers? Is single function easier to maintain?

The chained handlers works when each controller is just doing simple create, read, update, delete functions to one resource and the logic is pretty straightforward. This worked when we first started. But with more complexity it would be cleaner to move the common logic into functions and have one controller handle the more complex logic.

backend/src/Routes/elections.routes.js
scottPlusPlus: Yes!  Great catch, thanks

ArendPeter: Thanks for the update, looks good!

PR #134, opened 2022-07-31
refactoring voter roll and delete election controllers

mikefranze:
First pass at refactoring the voter roll controllers and the delete election controllers. They didn't require multiple databases, but I have a few TODOs for after Scott's PR is merged. I combined a few of them in the same file where it looks like it made sense.

### DIFF
--- a/backend/src/Controllers/getElectionRollController.ts
+++ b/backend/src/Controllers/getElectionRollController.ts
@@ -1,0 +1,65 @@
+import { ElectionRoll, ElectionRollState } from "../../../domain_model/ElectionRoll";
+import ServiceLocator from "../ServiceLocator";
+import Logger from "../Services/Logging/Logger";
+import { responseErr } from "../Util";
+import { hasPermission, permissions } from '../../../domain_model/permissions';
+
+const ElectionRollModel = ServiceLocator.electionRollDb();
+
+const className = "VoterRolls.Controllers";
+
+const getRollsByElectionID = async (req: any, res: any, next: any) => {
+    if (!hasPermission(req.user_auth.roles, permissions.canViewElectionRoll)) {
+        var msg = "Does not have permission";
+        Logger.info(req, msg);
+        return responseErr(res, req, 401, msg);
+    }
+    const electionId = req.election.election_id;
+    Logger.info(req, `${className}.getRollsByElectionID ${electionId}`);
+    //requires election data in req, adds entire election roll 
+    try {
+        const electionRoll = await ElectionRollModel.getRollsByElectionID(electionId, req);
+        if (!electionRoll) {
+            const msg = `Election roll for ${electionId} not found`;
+            Logger.info(req, msg);
+            return responseErr(res, req, 400, msg);
+        }
+
+        Logger.debug(req, `Got Election: ${req.params.id}`, electionRoll);
+        req.electionRoll = electionRoll
+        Logger.info(req, `${className}.returnRolls ${req.params.id}`);
+        res.json({ election: req.election, electionRoll: req.electionRoll });
+    } catch (err: any) {
+        const msg = `Could not retrieve election roll`;
+        Logger.error(req, `${msg}: ${err.message}`);
+        return responseErr(res, req, 500, msg);
+    }
+}
+
+const getByVoterID = async (req: any, res: any, next: any) => {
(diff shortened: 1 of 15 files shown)


ArendPeter:
I think the elections controller still needs to be deleted?

mikefranze:
> I think the elections controller still needs to be deleted?

Thanks, old middleware deleted.

backend/src/Controllers/voterRolls.controller.ts
```diff
@@ -30,11 +30,6 @@ const getRollsByElectionID = async (req: any, res: any, next: any) => {
     }
 }
 
-const returnRolls = async (req: any, res: any, next: any) => {
-    Logger.info(req, `${className}.returnRolls ${req.params.id}`);
-    res.json({ election: req.election, electionRoll: req.electionRoll });
-}
-
 const addElectionRoll = async (req: any, res: any, next: any) => {
```
ArendPeter, line 33: Should this snippet be deleted since it was replace by addElectionRollController.ts?

backend/src/Controllers/getElectionRollController.ts
```diff
@@ -0,0 +1,65 @@
...
+import ServiceLocator from "../ServiceLocator";
+import Logger from "../Services/Logging/Logger";
+import { responseErr } from "../Util";
+import { hasPermission, permissions } from '../../../domain_model/permissions';
+
+const ElectionRollModel = ServiceLocator.electionRollDb();
+
+const className = "VoterRolls.Controllers";
+
+const getRollsByElectionID = async (req: any, res: any, next: any) => {
```
ArendPeter, line 11: I think the corresponding function in backend/src/Controllers/voterRolls.controller.ts needs to be deleted?

ArendPeter (changes requested):
Thanks for adding the updates! I think there's still a few missing though

backend/src/Controllers/voterRolls.controller.ts
mikefranze: That is still used in the create election route which scott is working on.

backend/src/Controllers/getElectionRollController.ts
mikefranze: That is still used by the '/Election/:id/finalize' route which gets the rolls and sends the invitations.

ArendPeter (approved):
Looks good, thanks for the clarifications!

backend/src/Routes/elections.routes.js
```diff
@@ -6,65 +6,42 @@ const electionController = require('../Controllers/elections.controllers')
...
-    authController.hasPermission(permissions.canAddToElectionRoll),
-    voterRollController.addElectionRoll)
+router.get('/Election/:id/rolls', authController.getUser, getRollsByElectionID)
+router.get('/Election/:id/rolls/:voter_id', authController.getUser, getByVoterID)
+router.post('/Election/:id/rolls/', authController.getUser, editElectionRoll)
+router.post('/Election/:id/rolls/approve', authController.getUser, changeElectionRollStateController.approveElectionRoll)
+router.post('/Election/:id/rolls/flag', authController.getUser, changeElectionRollStateController.flagElectionRoll)
+router.post('/Election/:id/rolls/invalidate', authController.getUser, changeElectionRollStateController.invalidateElectionRoll)
+router.post('/Election/:id/rolls/unflag', authController.getUser, changeElectionRollStateController.uninvalidateElectionRoll)
+router.post('/Election/:id/rolls/', authController.getUser, addElectionRoll)
```
scottPlusPlus, line 44: In general, I'd like to see the controllers handle everything they need explicitly, rather than implicitly assuming that other methods have been chained in the route that stash data into the req.

scottPlusPlus: If it's really every route, then yes exactly we can use a middleware  (like you did for the catchError), and then add the user data explicitly to the IRequest (even if it's nullable).

If it's going to be conditional per route, I'd suggest putting it in a single function to handle the route, so that all the code required for the route is in one place.

mikefranze:
- Added async handler to the routes that will catch any errors and throw them to our custom error handler middleware. 
- Moved getUser to main app middleware and moved user_auth logic to getElectionByID middleware that runs whenever we have the id route param. 
- Added expect permission controller util, we can start to phase out where we use hasPermission functions.

scottPlusPlus:
LGTM!

PR #136, opened 2022-08-03
Error catching

mikefranze:
On top of #134 

Demo of an error catching middleware hoping to reduce the amount of try-catches in the code. With this we can just throw errors anywhere up to the error catcher using curveball/http-errors


scottPlusPlus (commented):
This middleware is awesome.

My only question is about the "failMsgPrfx"?  Is that something we want to show to users?

mikefranze:
> This middleware is awesome.
> 
> My only question is about the "failMsgPrfx"? Is that something we want to show to users?

Good point, I guess we only need it for the logger.

mikefranze:
Turns out this doesn't work with async functions. See #134 for better implementation.

PR #137, opened 2022-08-08
Election form refactor

mikefranze:
Moved election settings into an expandable section, added text field to quickly add candidates instead of the button which opened the more detailed form.

Also fixed error in election roll database that returned an empty array instead of null

https://user-images.githubusercontent.com/41272412/183326354-7c56ee2f-58dc-4343-a166-2e3a374c75be.mp4


ArendPeter (approved):
Nice! This looks like a much sleeker UX!

PR #139, opened 2022-08-12
changing ballot_id and election_id to be Uids

scottPlusPlus:
Updates ballot_id and election_id to be uuids (universally unique ids).  This means they can be generate by different processes without coordination.

THIS WILL BREAK production data, and should be accompanied by a wipe of the prod database.  (theoretically we could run a migration, but since we don't have any needed data yet, just nuke it)


mikefranze (approved):
Looks good, let me know when you plan to merge and I'll wipe the databases.

Also, did you test if this breaks anything in the front end?

PR #140, opened 2022-08-14
running npm audit, adding frontend debug page

scottPlusPlus:
It's been difficult running the front-end, due to an error in the create-react-app proxy.  This seems to fix it.

- Also added a debug page to the front-end, right now just lets you set a JWT token to bypass Keycloak if necessary.
- Also fixed display of election results with no votes.

I've still found some bugs (not able to vote in election with no voter roll).  I believe they likely already existed and would rather fix AFTER the controllers are refactored.

If someone could do a sanity check that this does not regress the frontend would be awesome.


mikefranze:
Do you remember what the exact error was with create-react-app?

PR #157, opened 2022-08-22
Refactor finalize route

mikefranze:
Creates finalize election controller and test. Currently test helper can't detect sent emails but can see sent emails when console logged. Tried importing service helper and importing mock service directly.

### DIFF
--- a/backend/src/Controllers/finalizeElectionController.ts
+++ b/backend/src/Controllers/finalizeElectionController.ts
@@ -12,8 +12,8 @@ var EmailService = ServiceLocator.emailService();
 const className = "election.Controllers";
 
 const finalizeElection = async (req: any, res: any, next: any) => {
-    expectPermission(req.user_auth.roles, permissions.canEditElectionState)
     Logger.info(req, `${className}.finalize ${req.election.election_id}`);
+    expectPermission(req.user_auth.roles, permissions.canEditElectionState)
     if (req.election.state !== 'draft') {
         var msg = "Election already finalized";
         Logger.info(req, msg);
@@ -39,7 +39,7 @@ const finalizeElection = async (req: any, res: any, next: any) => {
         Logger.info(req, `${className}.sendInvitations`, { election_id: req.election.election_id });
         try {
             const url = req.protocol + '://' + req.get('host')
-            const invites = Invites(req.election, req.electionRoll, url)
+            const invites = Invites(req.election, electionRoll, url)
             EmailService.sendEmails(invites)
         } catch (err: any) {
             const msg = `Could not send invitations`;
--- a/backend/src/Controllers/finalizeElectionController.ts
+++ b/backend/src/Controllers/finalizeElectionController.ts
@@ -1,0 +1,56 @@
+import ServiceLocator from '../ServiceLocator';
+import Logger from '../Services/Logging/Logger';
+import { permissions } from '../../../domain_model/permissions';
+import { expectPermission } from "./controllerUtils";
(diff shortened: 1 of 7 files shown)


backend/src/Controllers/finalizeElectionController.ts
```diff
@@ -0,0 +1,56 @@
...
+        throw new BadRequest(msg)
+    }
+    var failMsg = "Failed to update Election";
+    req.election.state = 'finalized'
+    const updatedElection = await ElectionsModel.updateElection(req.election, req, `Finalizing election`);
+    if (!updatedElection) {
+        Logger.info(req, failMsg);
+        throw new BadRequest(failMsg)
+    }
+    req.election = updatedElection
```
scottPlusPlus, line 29: what is the need to store the election back in the request object?  I would avoid this whenever possible as it becomes unclear where this data is used.

backend/src/Controllers/finalizeElectionController.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (req.election.settings.election_roll_type === 'Email') {
+        const electionId = req.election.election_id;
+        const electionRoll = await ElectionRollModel.getRollsByElectionID(electionId, req);
+        if (!electionRoll) {
+            const msg = `Election roll for ${electionId} not found`;
+            Logger.info(req, msg);
+            throw new BadRequest(msg)
```
scottPlusPlus, line 37: can we move this check above ElectionsModel.updateElection?  Basically want to do all checks, and then have all mutations at the end  (to minimize what could go wrong across different mutations).  If we refactor this to use event-driven-arch later, this point may be moot.

backend/src/test/TestHelper.ts
```diff
@@ -5,6 +5,7 @@ import { VoterAuth } from "../../../domain_model/VoterAuth";
 import makeApp from "../app";
 import Logger from "../Services/Logging/Logger";
 import { TestLoggerImpl } from "../Services/Logging/TestLoggerImpl";
+import ServiceLocator  from "../__mocks__/ServiceLocator"
```
scottPlusPlus, line 8: pull in the actual ServiceLocator instead of the mock to fix the problem with the test
import ServiceLocator  from "../ServiceLocator"

JEST is using some magic to swap out the ServiceLocator with the Mock.   As far as I can tell it creates a "mock-instance" of the Service Locator, which is different than when you pull in the mock SL directly...

I wouldn't mind to go in and make this explicit instead of relying on JEST voodoo, but not tonight :P

scottPlusPlus (changes requested):

backend/src/Controllers/finalizeElectionController.ts
mikefranze: Fixed, was from old code.

backend/src/Controllers/finalizeElectionController.ts
mikefranze: Fixed. This was a good idea, wouldn't want someone to forget to add the email roll and end up not being able to fix it.

backend/src/test/TestHelper.ts
mikefranze: This fixed it, thanks! I forgot about the Jest mock voodoo. The jest mocks don't behave quite like I'd expect so I wouldn't mind swapping them out somehow. I was thinking this would be a good case for dependency injection but you probably know some javascript voodoo to do it.

scottPlusPlus (approved):

PR #158, opened 2022-08-22
Account service

mikefranze:
Creates account service.
Moves get token axios call from app.js to service.
Creates getToken controller.
Creates account service mock, but is empty. Not sure how we want to use that yet.


backend/src/app.ts
```diff
@@ -53,66 +54,7 @@ const prodEndpoints : any = [
...
-        qs.stringify(params), 
-        {
-            headers: {
-                'Content-Type': 'application/x-www-form-urlencoded',
-                'Authorization': `Basic ${Buffer.from(`${authConfig.clientId}:${process.env.KEYCLOAK_SECRET}`).toString('base64')}`
-            }
-        }
-    )
-    .then(res => res.data)
-    .then(data => {Logger.debug(appInitContext, "success!"); res.json(data)})
```
scottPlusPlus, line 110: Here we used the Express Response object to send the token (data) back to the client. 

Does asyncHandler have some magic to do that based on the return value from getUserToken?

mikefranze: The response is sent in the same way in the getUserTokenController.

scottPlusPlus (approved):

PR #159, opened 2022-08-23
Feat/vote controller 2

scottPlusPlus:
refactors the Cast Vote controller.  Uses Postgres Transaction to update both tables simultaneously.  As discussed, will probably refactor this again to add the email confirmation and event-driven arch, but here's what's working for now :p


backend/src/Controllers/castVoteController.ts
```diff
@@ -0,0 +1,187 @@
...
+        const errMsg = "Invalid Ballot: "+ validationErr
+        Logger.info(req, errMsg);
+        throw new BadRequest(errMsg);
+    }
+
+    //TODO: currently we have both a value on the input Ballot, and the route param.
+    //do we want to keep both?  enforce that they match?
+    const targetElectionId = req.params.id;
+    inputBallot.election_id = targetElectionId;
+    const targetElection = await ElectionsModel.getElectionByID(targetElectionId, req);
```
mikefranze, line 34: Election data should already be available in req.election by this point from the param middleware in election.routes.js.

Probably wouldn't hurt to enforce that the ids match.

backend/src/Controllers/castVoteController.ts
```diff
@@ -0,0 +1,187 @@
...
+const ElectionsModel = ServiceLocator.electionsDb();
+const ElectionRollModel = ServiceLocator.electionRollDb();
+const BallotModel = ServiceLocator.ballotsDb();
+const CastVoteStore = ServiceLocator.castVoteStore();
+
+
+async function castVoteController(req: IRequest, res: any, next: any) {
+    Logger.info(req, "Cast Vote Controller");
+
+    const user = expectUserFromRequest(req);
```
mikefranze, line 21: User info would already be available in req.user. We get the user data if available on every request in app.ts 
`app.use(authController.getUser)`

We also don't need to assert this with every vote, just when the voter id is set to email.

backend/src/Controllers/castVoteController.ts
```diff
@@ -0,0 +1,187 @@
...
+
+function getVoterID(req:IRequest, voterIdType:string, user:any):string {
+    if (voterIdType === 'None') {
+        return "";
+    } else if (voterIdType === 'IP Address') {
+        Logger.debug(req, `ip=${String(req.ip)}`);
+        return String(req.ip)
+    } else if (voterIdType === 'Email') {
+        if (user == null){
+            throw new Unauthorized();
```
mikefranze, line 114: Currently the front end is expecting a response with info about what is required to vote. Its probably better to respond with unauthorized, but we should probably add info about why its unauthorized like "Log in required" or "Voter ID reqired" or "Already voted"

backend/src/Controllers/castVoteController.ts
```diff
@@ -0,0 +1,187 @@
...
+    } catch (err:any){
+        Logger.error(ctx, 'Failed to cast Ballot:  '+err.message);
+        throw err;
+    }
+
+    return savedBallot;
+}
+
+
+function getVoterID(req:IRequest, voterIdType:string, user:any):string {
```
mikefranze, line 106: getVoterID, getOrCreateElectionRoll, and assertVoterMayVote could be used in other controllers where the voter is checking if they can vote. We might want to move them to a util file or export them.

backend/src/Controllers/castVoteController.ts
scottPlusPlus: There are a bunch of tests where the ids don't match, so will leave enforcing this as a TODO for another diff.

backend/src/Controllers/castVoteController.ts
scottPlusPlus: I'm leaning towards leaving them here for now (probably out of laziness) and then whenever they need to be used elsewhere, go ahead and make the move to contollerUtils?

mikefranze (approved):
Looks good, just minor conflict to resolve

PR #160, opened 2022-08-29
more controller refactoring

mikefranze:
Moves more routes to individual files. Only remaining routes to be refactored include voterRollController.getVoterAuth which can be done after #159 is merged.


backend/src/Controllers/getBallotsByElectionIDController.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const BallotModel = ServiceLocator.ballotsDb();
+
+const getBallotsByElectionID = async (req: any, res: any, next: any) => {
+    var electionId = req.election.election_id;
+    Logger.debug(req, "getBallotsByElectionID: " + electionId);
+
+    expectPermission(req.user_auth.roles, permissions.canViewBallots)
+
+    const ballots = await BallotModel.getBallotsByElectionID(String(electionId), req);
```
scottPlusPlus, line 17: can we clarify when / why this would be null vs an empty array?
requesting ballots, and no one has voted, seems a legitimate case to handle and I wouldn't necessarily call that an error.

backend/src/Controllers/getBallotsByElectionIDController.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!ballots) {
+        const msg = `Ballots not found for Election ${electionId}`;
+        Logger.info(req, msg);
+        throw new BadRequest(msg)
+    }
+    Logger.debug(req, "ballots = ", ballots);
+    req.ballots = ballots;
```
scottPlusPlus, line 24: can we remove this?  (just put the ballots in the res.json)

backend/src/Controllers/getElectionResultsController.ts
```diff
@@ -0,0 +1,42 @@
+import ServiceLocator from "../ServiceLocator";
+import Logger from "../Services/Logging/Logger";
+import { BadRequest } from "@curveball/http-errors";
+import { Ballot } from '../../../domain_model/Ballot';
+import { Score } from '../../../domain_model/Score';
+
+const StarResults = require('../Tabulators/StarResults.js');
+const BallotModel = ServiceLocator.ballotsDb();
+
+const className = "VoterRolls.Controllers";
```
scottPlusPlus, line 10: update this i suppose

backend/src/Controllers/getElectionsController.ts
```diff
@@ -0,0 +1,22 @@
...
+const getElections = async (req: any, res: any, next: any) => {
+    Logger.info(req, `getElections`);
+    var filter = (req.query.filter == undefined) ? "" : req.query.filter;
+    const Elections = await ElectionsModel.getElections(filter, req);
+    if (!Elections) {
+        var msg = "Election does not exist";
+        Logger.info(req, msg);
+        throw new BadRequest(msg);
+    }
+    res.json(Elections);
```
scottPlusPlus, line 17: seems we normally return something like 
res.json( { elections : Elections })

so would be good to be consistent (unless specific need to deviate, if so maybe add comment)

scottPlusPlus (approved):
awesome!

backend/src/Controllers/getElectionsController.ts
mikefranze: Updated in recent commit

backend/src/Controllers/getElectionResultsController.ts
mikefranze: Updated in recent commit

backend/src/Controllers/getBallotsByElectionIDController.ts
mikefranze: Good catch, left over from old code. Updated in recent commit

backend/src/Controllers/getBallotsByElectionIDController.ts
mikefranze: Good question, currently the ballot model will only output empty array. Maybe we don't need to consider a null output.

PR #161, opened 2022-08-30
Adding ballot receipts

mikefranze:
Adds ballot receipt to voters who are logged in. 

Receipt email contains ballot ID and link to verify ballot in front end that currently does not exist, needs to be added.


backend/src/Controllers/castVoteController.ts
```diff
@@ -78,6 +79,12 @@ async function castVoteController(req: IRequest, res: any, next: any) {
     }
 
     const savedBallot = await persistBallotToStore(inputBallot, roll, req);
+
+    if (user.email){
+        const url = req.protocol + '://' + req.get('host')
+        const receipt = Receipt(targetElection, user.email,savedBallot, url)
+        EmailService.sendEmails([receipt])
```
scottPlusPlus, line 86: looks like sendEmails is async, so this should be awaited

scottPlusPlus (approved):

backend/src/Controllers/castVoteController.ts
mikefranze: Thanks

PR #163, opened 2022-09-10
event queue for CastVoteController

scottPlusPlus:
* pulled in pg-boss package  (event queue using postgres)
* added abstraction around it with mock
* implemented castVoteContoller using eventQueue

will want to add an async init method to service locator to make setup cleaner, but this works for now


mikefranze:
Awesome, it also looks like pg-boss has the option to set the datetime that a job should be run after. That would help us set up timed events like when the election starts/stops, email reminders, etc.

mikefranze (approved):

mikefranze:
Just needs a few conflicts to be resolved @scottPlusPlus

backend/src/ServiceLocator.ts
```diff
@@ -5,7 +5,11 @@ import ElectionRollDB from "./Models/ElectionRolls";
 import CastVoteStore from "./Models/CastVoteStore";
 import EmailService from "./Services/Email/EmailService";
 import { IBallotStore } from "./Models/IBallotStore";
+import { IEventQueue } from "./Services/EventQueue/IEventQueue";
+import PGBossEventQueue from "./Services/EventQueue/PGBossEventQueue";
+
 import AccountService from "./Services/Account/AccountService"
+import GlobalData from "./Services/GlobalData";
```
mikefranze, line 12: I think Services/GlobalData is missing from the commit

scottPlusPlus: whoops, great catch, thanks!

backend/src/Services/GlobalData.ts
```diff
@@ -0,0 +1,9 @@
+export default class GlobalData {
+
+  serviceUrl:string;
+
+  constructor() {
+    this.serviceUrl = "https://star-vote.herokuapp.com/";
```
mikefranze, line 6: Would it make sense for this to be an environment variable?

scottPlusPlus: Another great call :)

mikefranze:
Looks good, thanks!

PR #166, opened 2022-09-19
caching election results

scottPlusPlus:
Caches results of an election  for five minutes.

Still need to add some tests, but assume there may be some merge conflicts, so will hold until the tabulations refactor lands.

Just wanted to give heads up I was working on this


mikefranze:
This is a timely PR, we just had a related issue on star.vote that was causing the server to lock up.

ArendPeter:
I'm going to close some old PRs, we can always reopen them if we change our minds

PR #171, opened 2022-10-02
Front end cleanup

mikefranze:
- Fixing navigation after submitting election to route to new election page instead of home page.
- Disabling submit buttons while waiting for requests, displaying submitting message while waiting.
- Refreshes voter auth after voting so it correctly shows that voter has voted.
- Removes unused imports.
- Added data as output of usefetch.makerequest function in addition to setting data state so it can be used immediately without waiting for state change.
- Moved submit ballot button to top component, should make it easier to have multiple ballots.


scottPlusPlus (approved):
LGTM

PR #172, opened 2022-10-02
optional custom auth key for election

scottPlusPlus:
adds custom auth key field to an election.
when present, will use this key to authenticate users acting on the election, instead of the main app key

### DIFF
--- a/backend/src/Controllers/elections.controllers.ts
+++ b/backend/src/Controllers/elections.controllers.ts
@@ -7,10 +7,14 @@ import { roles } from "../../../domain_model/roles"
 
 
 var ElectionsModel =  ServiceLocator.electionsDb();
+var accountService = ServiceLocator.accountService();
 const className="Elections.Controllers";
 
 const getElectionByID = async (req: any, res: any, next: any) => {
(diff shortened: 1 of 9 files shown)


mikefranze (changes requested):
Looks good, but we might want to remove the auth key whenever we're returning the election in the response.

scottPlusPlus:
> Looks good, but we might want to remove the auth key whenever we're returning the election in the response.

Ah, great call.  Double checking: are there any situations where we would want to send that data back to the user?  (I could potentially see a need for election admin to update it, but not necessarily for them to view it)

scottPlusPlus:
hmm, and probably there are other fields we wouldn't want to show to all users, like the audit_ids...

mikefranze:
I don't think there's a need to ever send it in a response. Lets keep the other fields for now, I'm thinking we'd want information about auditors, admins, and credentialers to be public for transparency, but that does remind me of the issue that they are currently stored as emails instead of IDs. I'll make another issue ticket for that.

mikefranze (approved):

PR #174, opened 2022-10-08
Multi race election

mikefranze:
Built on top of #171 
Adds multiple race support
Some of it can be cleaned up a bit more but good enough for first cut. Currently all ballots are displayed on one page, should probably only show one at a time with buttons to flip through them.
Doesn't include precinct logic yet, will tackle that next.
Demo: https://www.youtube.com/watch?v=Bwb6OHTzkpY


backend/src/Controllers/getElectionResultsController.ts
```diff
@@ -22,31 +22,32 @@ const getElectionResults = async (req: any, res: any, next: any) => {
...
 
     const election = req.election
-    const candidateNames = election.races[0].candidates.map((Candidate: any) => (Candidate.candidate_name))
-    const cvr = ballots.map((ballot: Ballot) => (
-        ballot.votes[0].scores.map((score: Score) => (
-            score.score
+    let results = []
+    for (let race_index = 0; race_index < election.races.length; race_index++) {
+        const candidateNames = election.races[race_index].candidates.map((Candidate: any) => (Candidate.candidate_name))
+        const cvr = ballots.map((ballot: Ballot) => (
```
scottPlusPlus, line 28: what does cvr stand for?

scottPlusPlus (changes requested):
Can we get some testing on the backend for multi-election races?

Also add some validation when a vote is submitted that it contains the expected number of races.

backend/src/Controllers/getElectionResultsController.ts
mikefranze: Tests and validation added. 
CVR stands for cast vote record, was inherited from Jay's old code.

scottPlusPlus (approved):
awesome!

PR #176, opened 2022-10-12
Precincts

mikefranze:
Initial cut at adding precinct logic. There's a few more things to add but wanted to open PR now in case someone has a better idea on how to handle this.

-Adds voter's precinct to roll and ballot(so we can later look at results at a precinct level)
-Adds list of precincts to race, if no list is specified then its open to all voters
-Checks that voter's precinct is in all races voted for on ballot 
-Returns election with only races with voter's precinct if election is open
-Fixes bug in get election results controller, now when iterating through ballot it checks if ballot has a vote for current race being processed. There is probably a more efficient way to do this but this works for now.

Remaining todos: add tests, add ability to set precinct in addElectionRollController, add to frontend forms.


domain_model/Election.ts
```diff
@@ -39,6 +40,27 @@ export function electionValidation(obj:Election): string | null {
...
+export function removeHiddenFields(obj:Election, electionRoll: ElectionRoll|null):void {
   obj.auth_key = undefined;
+  if (obj.state==='open' && electionRoll?.precinct){
+    // If election is open and precinct is defined, remove races that don't include precinct
+    obj.races = getApprovedRaces(obj, electionRoll.precinct)
+    
+  }
+}
+// Where should this belong..
+export function getApprovedRaces(election:Election, precinct:string|undefined) {
```
scottPlusPlus, line 52: nit:  specify the argument is meant to be the voter's precinct.  type the output as list of Races

domain_model/Election.ts
```diff
@@ -39,6 +40,27 @@ export function electionValidation(obj:Election): string | null {
   return null;
 }
 
-export function removeHiddenFields(obj:Election):void {
+export function removeHiddenFields(obj:Election, electionRoll: ElectionRoll|null):void {
   obj.auth_key = undefined;
+  if (obj.state==='open' && electionRoll?.precinct){
```
scottPlusPlus, line 45: do we need to filter out precincts for every request?  Is this considered private information?  are we sure there aren't any routes that need this info?

domain_model/Ballot.ts
```diff
@@ -12,6 +13,7 @@ export interface Ballot {
     ip_address?: string; // ip address if once_per_ip is enabled
     votes: Vote[];         // One per poll
     history?: BallotAction[];
+    precinct?: string; // Precint of voter
```
scottPlusPlus, line 16: double checking there's no such thing as a voter being in two separate precincts?  (maybe a state-level precinct and a county-level precinct or some craziness?)

domain_model/Ballot.ts
```diff
@@ -37,13 +39,14 @@ export function ballotValidation(election: Election, obj:Ballot): string | null
     if (!obj.votes){
         return "Invalid Votes";
     }
-    const race_ids = election.races.map((race: Race) => race.race_id)
-    const ballot_race_ids = obj.votes.map((vote: Vote) => vote.race_id)
-    const hasDuplicates = new Set(ballot_race_ids).size !== ballot_race_ids.length
+    const approvedRaces = getApprovedRaces(election, obj.precinct)
+    const approvedRaceIds = approvedRaces.map((race: Race) => race.race_id)
+    const ballotRaceIds = obj.votes.map((vote: Vote) => vote.race_id)
+    const hasDuplicates = new Set(ballotRaceIds).size !== ballotRaceIds.length
```
scottPlusPlus, line 45: nit:  may be worth making a "hasDuplicates" util method for an array

domain_model/Ballot.ts
```diff
@@ -37,13 +39,14 @@ export function ballotValidation(election: Election, obj:Ballot): string | null
...
     if (hasDuplicates) {
         return "Duplicate votes";
     }
-    const validIds = ballot_race_ids.every(id => race_ids.includes(id))
+    const validIds = ballotRaceIds.every(id => approvedRaceIds.includes(id))
     if (!validIds) {
         return "Invalid IDs";
```
scottPlusPlus, line 51: can we make this error more clear?  Which ids?

scottPlusPlus (commented):
Solid approach.  A few questions just to double check the requirements, but looks great.

domain_model/Election.ts
mikefranze: I think this might end up in a separate route just for ballot data. The intention was that when the voter requests the ballot to vote they only receive races that include their precinct, but only when the election is open so the other routes should still work. Another approach would be to include the list of approved races in the voter auth data and filter it client side.

domain_model/Ballot.ts
mikefranze: The way I've been looking at it is the precinct is the lowest level group that voters can be divided into, meaning everyone in that precinct gets the same ballot. You can have situations where the state senate, state assembly, counties, cities, etc., don't share the same district borders. So your neighbor could be in the same state assembly district but different senate district for example. 

When I was looking through the Oregon precinct database it looks like they had to further divide their precincts into sub-precincts which was confusing, but for simpler elections this should cover most use cases.

mikefranze:
Added tests and precinct support to frontend. I ended up removing the election roll input to the create election route, now adding the roll is done just with its own dedicated route. So after creating the election the admin can go to a dedicated page to add and view rolls.

I also changed the precinct list from a set to an array. Didn't realize sets don't work with JSON so we'd have to convert back and forth anyway. A set lookup is O(1) but converting to set from an array is O(n) so might as well stick with arrays. If we end up having issues with larger elections we could change it to be an object with the precinct names as properties which would be O(1).

PR #178, opened 2022-10-17
More front end cleanup

mikefranze:
Built on top of #176 

Cleans up the component file structures a bit
Adds buttons to flip through different races instead of having them all on one page.
Changed indexing in star ballot to be consistent with scores. Before 0 mean nothing selected and 1-6 mapped to 0-5. Now null means nothing selected and the rest maps as expected.

https://user-images.githubusercontent.com/41272412/196080820-d32efccf-80d7-4fa5-a96f-defcf40e2640.mp4


backend/src/Models/CastVoteStore.ts
```diff
@@ -29,11 +29,12 @@ export default class CastVoteStore {
             ballot.date_submitted,
             ballot.ip_address,
             JSON.stringify(ballot.votes),
-            JSON.stringify(ballot.history),
+            JSON.stringify(ballot.history),,
```
MagnusMuscle, line 32: Is this double comma intentional? Seems wrong to me.

mikefranze: Good catch, was a typo

mikefranze:
Added a few more big updates that basically touched every file, I don't expect people to really dig through the code for review so I'll include screenshots of the final product. Still needs some styling but this should help make it easier in the future.

Added hooks for local storage so the election forms now persist if you close the window.

Added a random "tempID" to local storage. If the user isn't logged in it will use that as the user ID. The intention is to allow you to submit a quick poll if you're not logged in and also be able to (in a future update) see all polls created by that ID.

Added a quick poll form to the landing page with a button to log in for more features. This uses the "tempID if you're not logged in and sets the election state to "open".
image

Upgraded the MUI styling library to the latest version which changed a lot of import lines.

Updated the header to be collapsible and responsive
image

Some styling updates to the election form and election landing page
image
image

And most importantly: Share buttons!
image

scottPlusPlus (commented):
firefox_aWUWz3M00r

Seems like the logged out version isn't quite working for me?

Also the log-in doesn't seem to be working for me either?  I can get through the keycloak, but then once it redirects me, I see all the sign in data in the url (ex: http://localhost:3000/?session_state=0ca8cbbc-bae2-4e8a-87a3-705cbf4839ed&code=4a4e1e84-dbbe-451c-a515-395d6914bb14.0ca8cbbc-bae2-4e8a-87a3-705cbf4839ed.7804660f-ad64-4892-b022-4229716d1107), but the top-right nav still says "Login"

I assume this is my user-error, but LMK

mikefranze:
Found the issue for the quick poll missing the vote button. I had moved it recently under a condition for when the election is open but forgot about another condition. 

For logged in issue try refreshing the page. Currently the log in checks are inside a class object being passed around and when that changes it doesn't trigger a rerender. I'm going to tackle that another day.

scottPlusPlus:
The refresh still didn't work for me.  But anyway this is probably a net win.  Let's go ahead and land the refactor and then can work on fixing any issues one at a time.

mikefranze:
You might be missing the keycloak public key

PR #199, opened 2022-12-19
Default to Mock Email Service without SendGrid API key

scottPlusPlus:
ConEmu64_rznxwDBmGi

Current bottleneck to load-testing is the email service.   Adding option to default to mock Email service when no SendGrid API key.

And when doing that, remove the eternally growing copy of every email sent :P


mikefranze:
Is this for load testing or the Jest unit testing?

scottPlusPlus:
for load testing.  Currently with the Jest unit testing we use the mock ServiceLocator which uses the  mock EmailService anyway.

mikefranze (approved):

backend/src/Services/Email/EmailService.ts
```diff
@@ -1,10 +1,11 @@
...
-export default class EmailService {
+export default class EmailService implements IEmailService {
 
   sgMail;
 
-  constructor() {
+  constructor(sendGrigApiKey:string) {
     this.sgMail = require('@sendgrid/mail')
     require('dotenv').config()
     this.sgMail.setApiKey(process.env.SENDGRID_API_KEY)
```
ArendPeter, line 11: I think this should use sendGridApiKey from the instructor instead of referencing process.env?

ArendPeter:
I'm going to close some old PRs, we can always reopen them if we change our minds

PR #210, opened 2023-01-04
New score tiebreaker protocol

mikefranze:
We have updated our score tiebreaker protocol again, flowchart [here](https://docs.google.com/drawings/d/1L-tIDY6-8OeKGzW0w1eh2EUjlHbarEcsfvhsegMT8qg/edit?usp=sharing).

Before we would look for any weak Condorcet winner, if none existed we'd look for a candidate with the most five star votes, if none existed we'd pick a random winner.

Now we eliminate candidates with the most head to head losses until one can advance. If that doesn't work we look for a candidate with the most five star votes. If that doesn't work we eliminate candidates with the least five star votes. Finally if that doesn't work we pick a random winner. 

While more complex this should let us select better winners, cover most weird edge cases, and avoid random winners more often.

### DIFF
--- a/backend/src/Tabulators/Star.ts
+++ b/backend/src/Tabulators/Star.ts
@@ -2,6 +2,14 @@ import { ballot, candidate, fiveStarCount, results, roundResults, summaryData, t
 
 import { IparsedData } from './ParseData'
 const ParseData = require("./ParseData");
+declare namespace Intl {
+  class ListFormat {
+    constructor(locales?: string | string[], options?: {});
+    public format: (items: string[]) => string;
+  }
+}
+// converts list of strings to string with correct grammar ([a,b,c] => 'a, b, and c')
+const formatter = new Intl.ListFormat('en', { style: 'long', type: 'conjunction' });
 
 export function Star(candidates: string[], votes: ballot[], nWinners = 1, breakTiesRandomly = true, enablefiveStarTiebreaker = true) {
   // Determines STAR winners for given election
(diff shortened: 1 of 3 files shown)


backend/src/Tabulators/Star.ts
```diff
@@ -171,62 +171,83 @@ function runStarRound(summaryData: summaryData, remainingCandidates: candidate[]
...
-      if (fiveStarWinners.length === 2 && runoffCandidates.length === 0) {
-        // Two candidates win five-star tiebreaker, both advance to runoff
-        runoffCandidates.push(...fiveStarWinners)
-        roundResults.logs.push(`${fiveStarWinners[0].name} and ${fiveStarWinners[1].name} win five-star tiebreaker and advances to the runoff round.`)
-        continue
+      // All tied candidates have the same number of head to head lossess
+      if (nCandidatesNeeded === 2 && tiedCandidates.length === 2) {
+        // Tie between two candidates, but both can advance to runoff
+        runoffCandidates.push(...tiedCandidates)
+        roundResults.logs.push(`${tiedCandidates[0].name} and ${tiedCandidates[1].name} advance to the runoff round.`)
```
ArendPeter, line 207: nit, maybe this log can be worded differently to distinguish it from line 188. That way we can tell if the tie was passed through during the tiebreaker phase instead of prior. Maybe "advanced to runoff round from tiebreaker"?

backend/src/Tabulators/Star.ts
```diff
@@ -171,62 +171,83 @@ function runStarRound(summaryData: summaryData, remainingCandidates: candidate[]
...
+        tiedCandidates = tiedCandidates.filter(c => !headToHeadLosers.includes(c))
+        continue tieLoop
       }
-      if (fiveStarWinners.length === 2 && runoffCandidates.length === 0) {
-        // Two candidates win five-star tiebreaker, both advance to runoff
-        runoffCandidates.push(...fiveStarWinners)
-        roundResults.logs.push(`${fiveStarWinners[0].name} and ${fiveStarWinners[1].name} win five-star tiebreaker and advances to the runoff round.`)
-        continue
+      // All tied candidates have the same number of head to head lossess
+      if (nCandidatesNeeded === 2 && tiedCandidates.length === 2) {
```
ArendPeter, line 204: nit, I think this would read slightly clearer if this check was at the top of the tie loop. I was trying to understand how this was related tot eh headToHeadLosers check, but I think it's unrelated

ArendPeter (approved):
Great work!

backend/src/Tabulators/Star.ts
```diff
@@ -32,14 +32,14 @@ export function Star(candidates: string[], votes: ballot[], nWinners = 1, breakT
   var remainingCandidates = [...summaryData.candidates]
```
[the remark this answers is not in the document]
mikefranze: Technically it still applies in the runoff round so I made a small clarification.

backend/src/Tabulators/Star.ts
mikefranze: Done

backend/src/Tabulators/Star.ts
mikefranze: It is there for a reason, if you start with three tied candidates and have two candidates remaining, you could have one with a loss and one without, so you'd want to want to check first if you can eliminate a candidate.

mikefranze:
Added a list formatter that uses correct grammar 
['a','b'] => 'a and b'
['a','b','c'] => 'a, b, and c'

PR #227, opened 2023-02-17
Adding basic single winner IRV tabulator for running comparison elections

mikefranze:
Adding basic single winner IRV tabulator for running elections comparing different voting methods, not to be recommended for general use.

### DIFF
--- a/backend/src/Tabulators/IRV.ts
+++ b/backend/src/Tabulators/IRV.ts
@@ -87,7 +87,7 @@ export function IRV(candidates: string[], votes: ballot[], nWinners = 1, breakTi
         // get max number of votes
         let remainingCandidatesIndexes = remainingCandidates.map(c => c.index)
         let maxVotes = Math.max(...roundVoteCounts.filter((c, i) => remainingCandidatesIndexes.includes(i)))
-        if (maxVotes >= voteThreshold) {
+        if (maxVotes > voteThreshold || remainingCandidates.length <= 2) {
             // candidate meets the threshold
             // get index of winning candidate
             let winningCandidateIndex = roundVoteCounts.indexOf(maxVotes)
--- a/backend/src/Tabulators/IRV.ts
+++ b/backend/src/Tabulators/IRV.ts
@@ -1,0 +1,241 @@
+import { ballot, candidate, irvResults, irvSummaryData, totalScore } from "./ITabulators";
+
+import { IparsedData } from './ParseData'
+const ParseData = require("./ParseData");
+
+export function IRV(candidates: string[], votes: ballot[], nWinners = 1, breakTiesRandomly = true) {
+    // Determines Instant Runoff winners for given election
+    // Parameters: 
+    // candidates: Array of candidate names
+    // votes: Array of votes, size nVoters x Candidates, zeros indicate no rank
+    // nWiners: Number of winners in election, defaulted to 1 (only supports 1 at the moment)
+    // breakTiesRandomly: In the event of a true tie, should a winner be selected at random, defaulted to true
+
+    // Parse the votes for valid, invalid, and undervotes, and identifies bullet votes
+    const parsedData = ParseData(votes, getIRVBallotValidity)
+
+    const summaryData = getSummaryData(candidates, parsedData)
+
+    // Initialize output data structure
+    const results: irvResults = {
+        elected: [],
+        tied: [],
+        other: [],
+        summaryData: summaryData,
+        logs: [],
+        voteCounts: [],
+        exhaustedVoteCounts: [],
+        overVoteCounts: []
+    }
+
+    let remainingCandidates = [...summaryData.candidates]
+    let activeVotes = parsedData.scores
+    let exhaustedVotes: ballot[] = []
+    let exhaustedVoteCount = 0
+    let overVoteCount = 0
+    while (results.elected.length < 1) { //TODO: multiwinner STV
+        let roundVoteCounts = Array(summaryData.candidates.length).fill(0)
+        for (let i = activeVotes.length - 1; i >= 0; i--) {
+            // loop backwards over the ballots so exhausted votes can be removed
+            let ballot = activeVotes[i]
+            let topRemainingRank: number = 0
+            let topRemainingRankIndex: number = 0
+            let isOverVote = false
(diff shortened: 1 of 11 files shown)


backend/src/Tabulators/IRV.ts
```diff
@@ -0,0 +1,241 @@
...
+                        topRemainingRank = ballot[candidate.index]
+                        isOverVote = false
+                    }
+                    else if (ballot[candidate.index] < ballot[topRemainingRankIndex]) {
+                        // set new top rank
+                        topRemainingRankIndex = candidate.index
+                        topRemainingRank = ballot[candidate.index]
+                        isOverVote = false
+                    }
+                    else if (ballot[candidate.index] === ballot[topRemainingRankIndex]) {
```
ArendPeter, line 62: I think this logic can skip some overvotes. What if ballot[topRemainingRankIndex] is set to 1 on the first iteration and then all remaining ranks are set to 2? This should be an overvote (I think?) but this comparison will evaluate to 1 === 2 for each iteration

backend/src/Tabulators/IRV.ts
```diff
@@ -0,0 +1,241 @@
...
+        }
+        results.voteCounts.push(roundVoteCounts)
+        results.exhaustedVoteCounts.push(exhaustedVoteCount)
+        results.overVoteCounts.push(overVoteCount)
+        let voteThreshold = activeVotes.length / 2
+        
+        // get max number of votes
+        let remainingCandidatesIndexes = remainingCandidates.map(c => c.index)
+        let maxVotes = Math.max(...roundVoteCounts.filter((c, i) => remainingCandidatesIndexes.includes(i)))
+        if (maxVotes >= voteThreshold) {
```
ArendPeter, line 90: I don't think this accounts for tie situations. Here's an example

```
A: 50
B: 30
C: 20
```

Here A has >= voteThreshold, but the remaining votes still need to be distributed to determine if A gets a win or if it's a tie between A and B

backend/src/Tabulators/IRV.ts
```diff
@@ -0,0 +1,241 @@
...
+            // get index of winning candidate
+            let winningCandidateIndex = roundVoteCounts.indexOf(maxVotes)
+            // add winner, remove from remaining candidates
+            results.elected.push(summaryData.candidates[winningCandidateIndex])
+            remainingCandidates = remainingCandidates.filter(c => !results.elected.includes(c))
+            // TODO: Some STV stuff
+        }
+        else {
+            // find candidate with least votes and remove from remaining candidates
+            let lastPlaceCandidateIndex = roundVoteCounts.indexOf(Math.min(...roundVoteCounts.filter((c, i) => remainingCandidatesIndexes.includes(i))))
```
ArendPeter, line 101: What if multiple candidates are tied for least votes? I'm not sure what the standard practice is there, should we eliminate all tied non-first place candidates?

backend/src/Tabulators/IRV.ts
```diff
@@ -0,0 +1,241 @@
...
+            remainingCandidates = remainingCandidates.filter(c => c !== summaryData.candidates[lastPlaceCandidateIndex])
+        }
+    }
+    // Sort data in order of candidate placements
+    // results.summaryData = sortData(summaryData, results.elected.concat(results.tied).concat(results.other))
+    return results
+}
+
+function getIRVBallotValidity(ballot: ballot) {
```
ArendPeter, line 110: The front end could use this information too when people are submitting their ballots. Would it be better to expose this api for consistency, or to reimplement this on the front end?

backend/src/Tabulators/IRV.ts
```diff
@@ -0,0 +1,241 @@
...
+                        // multiple top ranks, mark as over vote
+                        isOverVote = true
+                    }
+                }
+            })
+            if (isOverVote) {
+                // ballot is overvote
+                overVoteCount += 1
+            }
+            if (topRemainingRank === 0 || isOverVote) {
```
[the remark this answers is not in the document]
mikefranze: I looked at [this](https://legislature.maine.gov/statutes/21-A/title21-Asec723-A.html) as it was the first thing to pop up. "Exhausted ballot" means a ballot that does not rank any continuing candidate, contains an overvote at the highest continuing ranking or contains 2 or more sequential skipped rankings before its highest continuing ranking.  
So it is not invalid or exhausted until the overvoted candidates are your top remaining ranked candidates. I did not bother implementing the "sequential skipped rankings" part.

backend/src/Tabulators/IRV.ts
mikefranze: As said in other comment, overvote is only when its between your top remaining candidates. So if you rank candidate A 1, and rank both candidates B and C 2, it is not an overvote unless A is eliminated. Maybe I'm wrong though.

backend/src/Tabulators/IRV.ts
mikefranze: Is the threshold 50%+1?

backend/src/Tabulators/IRV.ts
mikefranze: According to [this](https://legislature.maine.gov/statutes/21-A/title21-Asec723-A.html) its determined by lottery. We could hook up the random candidate selector but I'm not all that concerned about getting this 100% correct at the moment as its just for demos.

backend/src/Tabulators/IRV.ts
mikefranze: Easiest would be to reimplement it. Plus you could add the "Potential over vote" warning to it.

backend/src/Tabulators/IRV.ts
ArendPeter: I think so, that makes the most sense to me

This source says 50+1, https://ballotpedia.org/Ranked-choice_voting_(RCV) but I'm not sure if that's the consensus

ArendPeter (approved):

backend/src/Tabulators/IRV.ts
mikefranze: Changed to needing to be greater than 50% threshold, also now selects the candidate with the most votes if your have 2 or less candidates which will handle ties.

ArendPeter (approved):

PR #228, opened 2023-02-21
Add plurality ballot support

ArendPeter:
⚠️NOTE: Highly recommend reviewing the individual commits instead of the code diff. They're much easier to follow

This commit does the following
 * Moves the StarBallotView into a GenericBallotView
 * Recreates StarBallotView to leverage GenericBallotView
 * Creates PluralityBallotView which also leverages GenericBallotView

I've got more more PRs coming for IRV and approval ;) 

image
image

### DIFF
--- a/frontend/src/components/Election/Voting/GenericBallotView.js
+++ b/frontend/src/components/Election/Voting/GenericBallotView.js
@@ -58,7 +58,7 @@ const Row = ({ rowIndex, candidate, score, onClick, columns }) => {
   return (
     <>
       <Grid container alignItems="center" style={{ backgroundColor: rowColor }}>
-        <Grid item xs={1} style={{paddingLeft: (columns.length == 1)? '25%' : '0%'}}>
+        <Grid item xs={1} >
           {!expanded && hasExpandedData &&
             <IconButton aria-label="Home" onClick={() => { setExpanded(true) }}>
               <ExpandMore />
@@ -69,7 +69,7 @@ const Row = ({ rowIndex, candidate, score, onClick, columns }) => {
             </IconButton>}
         </Grid>
         <Grid item xs={4}>
-          <Typography wrap sx={{wordwrap: "break-word"}} align='left' variant="h6" component="h6">
+          <Typography wrap sx={{wordwrap: "break-word"}} align={columns.length==1? 'right' : 'left'} variant="h6" component="h6">
             {candidate.candidate_name}
           </Typography>
         </Grid>
@@ -80,7 +80,6 @@ const Row = ({ rowIndex, candidate, score, onClick, columns }) => {
           onClick={onClick}
           columns={columns}
         />
-
       </Grid>
       {expanded && hasExpandedData &&
       <>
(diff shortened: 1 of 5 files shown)


frontend/src/components/Election/Voting/GenericBallotView.js
```diff
@@ -0,0 +1,221 @@
...
+          </Grid>
+
+          <ColumnHeadings columns={columns}/>
+          <Divider style={{ width: '100%' }} />
+          <Rows candidates={candidates} scores={scores} onClick={onClick} columns={columns}/>
+
+          <Grid item xs={10} style={{ padding: '0.4cm 0cm' }}>
+            {race.num_winners == 1 && race.voting_method == 'STAR' &&
+              <Typography align='center' component="p">
+                This election uses STAR Voting and will elect 1 winner. In STAR Voting the two highest scoring candidates are finalists and the finalist preferred by more voters wins.
```
mikefranze, line 208: We should move this to StarBallotView and add an input for the tabulation explanation to generic ballot view.

frontend/src/components/Election/Voting/GenericBallotView.js
```diff
@@ -0,0 +1,221 @@
...
+    <Grid item xs={1}>
+      <Typography align='center' variant="h6" component="h6">
+        {columns.length == 1? ' ' : columnValue}
+      </Typography>
+    </Grid>
+  ));
+
+
+
+// Renders a complete RCV ballot for a single race
```
mikefranze, line 173: nit: remove rcv references in comments in this file and in other ballot views. Not sure why it was there in the first place.

mikefranze:
Looks good, though I'm wondering if we could reduce the whitespace on the right side.

frontend/src/components/Election/Voting/GenericBallotView.js
ArendPeter: Sounds good, will address this in the next PR as well

frontend/src/components/Election/Voting/GenericBallotView.js
ArendPeter: Sounds good, I've got a fix for this ready in the next PR

ArendPeter:
> Looks good, though I'm wondering if we could reduce the whitespace on the right side.

Makes sense, I looked into this a little bit but I was having trouble figuring out which part of the styling to edit. When I inspected the element the color coding didn't match up with margin or padding?

image

mikefranze:
How does that padding look with the expanded ballot button? You can enable that by adding a bio to a candidate. Another way you could do it is setting xs={4} or some other number. That controls how many columns in the grid the component will take up. For the star ballot the expand button is 1 column, the name is 4, and each of the six circles is 1.

ArendPeter:
image
I just pushed a change making the text alignment depend on the ballot type, I think that'll be good for now?

I was struggling with the grid, my second preferred approach is to play more the padding, but that'll give problems for longer candidate names

mikefranze (approved):

PR #229, opened 2023-02-21
Fix edge case in plurality tabulator

ArendPeter:
I was getting a 500 error when viewing the results for a plurality election. Here's the edge case that was triggered

The election had candidates [A, B, C, D], there was only 1 vote cast, and it was for candidate A. It caused 2 problems with the tabulator

1. sortedScores always selected results.elected.length. In this case it would select A on the first iteration, and then select B on all future iterations. The index should probably have been sortedScores.length - remainingCandidates.length? But that wouldn't have accounted for the second issue
2. The tie breaker would selected a random candidate. This means the candidates aren't always eliminated in the sortedScore order, and this caused an infinite loop where remainingCandidates could have candidates that are never selected in scoreWinners

### DIFF
--- a/backend/src/Tabulators/Plurality.ts
+++ b/backend/src/Tabulators/Plurality.ts
@@ -13,24 +13,26 @@ export function Plurality(candidates: string[], votes: ballot[], nWinners = 1, b
     roundResults: [],
     summaryData: summaryData,
   }
-  const sortedScores = summaryData.totalScores.sort((a: totalScore, b: totalScore) => {
+  let sortedScores = summaryData.totalScores.sort((a: totalScore, b: totalScore) => {
     if (a.score > b.score) return -1
     if (a.score < b.score) return 1
     return 0
   })
-  
+
   var remainingCandidates = [...summaryData.candidates]
+  let prevRemainingCount = remainingCandidates.length;
   while (remainingCandidates.length>0) {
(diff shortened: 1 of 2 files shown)


mikefranze:
A few comments: 
Originally the sortedScores array was a reference to the summaryData.totalScores array that holds the total results we want to report, so I didn't want to edit it. But we can get around that the way we do with the candidates:
`var remainingCandidates = [...summaryData.candidates]`
That will prevent us from accidentally changing the summaryData. 

When we select a winner we can just filter it out of the array, then the next winner will just be the first index and we can do a quick check for ties easier. 

Next, if breakTiesRandomly is enabled we can make this a lot simpler. We can handle the random selection by building it into the sort function. 
`const sortedScores = summaryData.totalScores.sort((a: totalScore, b: totalScore) => {`
`    if (a.score > b.score) return -1`
`    if (a.score < b.score) return 1`
`    return 0.5 - Math.random()` <--Random sort for tied candidates
 ` })`

Then we just have to grab the first nWinners indexes and put them in the elected array and push the rest into the others array.

If breakTiesRandomly is not enabled, then we'd have to look for tied candidates, check if there's enough seats for them, and if so elect them all, if not put them into the tied array.

ArendPeter:
Closing since this was fixed here https://github.com/Equal-Vote/star-server/pull/241

PR #231, opened 2023-02-26
Add stars to the ballot

ArendPeter:
Tweaked ballot to be closer to the target ballot

# Updated Ballot
Screenshot_20230225_092906

# Reference Ballot
image

### DIFF
--- a/frontend/src/components/Election/Voting/GenericBallotView.js
+++ b/frontend/src/components/Election/Voting/GenericBallotView.js
@@ -178,8 +178,8 @@ const ColumnHeadings = ({starHeadings, columns, leftTitle, rightTitle, headingPr
 
 const ScoreIcon = ({color, value}) => (
   <div align='center'>
-    <FaStar style={{color: color, fontSize: '2.1em', transform:'translate(0%,15%)'}}/>
-    <Typography style={{position: 'absolute', transform:'translate(155%,-110%)'}} variant="h6" component="h6" className="scoreColumnHeading">
+    <FaStar style={{color: color}} className="starIcon"/>
+    <Typography variant="h6" component="h6" className="scoreColumnHeading">
       {value}
     </Typography>
   </div>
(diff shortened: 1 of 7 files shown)


frontend/src/components/Election/Voting/GenericBallotView.js
```diff
@@ -144,26 +142,56 @@ const Rows = ({ candidates, scores, onClick, columns }) =>
...
 // Represents the list of column headings for all possible scores
-const ColumnHeadings = ({columns}) => (
+const ColumnHeadings = ({starHeadings, columns, leftTitle, rightTitle, headingPrefix}) => (
   <>
-    <Grid container alignItems="stretch">
+  { leftTitle != '' &&
+    <Grid container alignItems="stretch" >
       <Grid item xs={5}></Grid>
-      <ScoreColumnHeadings columns={columns}/>
+      <Typography align='center' className="columnDescriptor">
```
mikefranze, line 155: Should put the leftTitle and rightTitle typeographies in a <Grid item xs={1}>, should help align with the other rows below.

frontend/src/components/Election/Voting/GenericBallotView.js
```diff
@@ -144,26 +142,56 @@ const Rows = ({ candidates, scores, onClick, columns }) =>
...
+      }
     </Grid>
   </>
 );
 
-const ScoreColumnHeadings = ({columns}) =>
+const ScoreIcon = ({color, value}) => (
+  <div align='center'>
+    <FaStar style={{color: color, fontSize: '2.1em', transform:'translate(0%,15%)'}}/>
+    <Typography style={{position: 'absolute', transform:'translate(155%,-110%)'}} variant="h6" component="h6" className="scoreColumnHeading">
```
mikefranze, line 182: We should try to make this more responsive on smaller screens. The .circle css used for the ballot bubbles might be a good reference. 

https://user-images.githubusercontent.com/41272412/221441804-984d19b1-c42d-46e9-be03-b624f40fab99.mp4

ArendPeter: Good catch!

.circle seems different since it's manipulating the border to be a circle, whereas for the star I'm trying to render multiple elements on top of each other

I made a change that I think fixes it? That said css is still kind of black magic to me so I just keep trying random things until it works 😅

mikefranze (approved):

PR #234, opened 2023-02-27
Add Ranked Ballots and Approval Ballots

ArendPeter:
Add ranked ballots and approval ballots

image
image

### DIFF
--- a/frontend/src/components/Election/Voting/BallotSelector.js
+++ b/frontend/src/components/Election/Voting/BallotSelector.js
@@ -2,6 +2,7 @@ import React, { useState } from "react";
 import StarBallotView from "./StarBallotView";
 import PluralityBallotView from "./PluralityBallotView.js";
 import RankedBallotView from "./RankedBallotView.js";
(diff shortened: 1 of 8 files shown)


frontend/src/components/Election/Voting/BallotSelector.js
```diff
@@ -27,6 +29,22 @@ export default function BallotSelector({
         onUpdate={onUpdate}
         />
     }
+    {(race.voting_method == 'Ranked-Robin' || race.voting_method == "IRV") &&
```
mikefranze, line 32: We'll probably need separate ballot view for RCV and Ranked Robin as the instructions and tabulation description will be different.

ArendPeter: I'm happy to do this in the future, but for the Ranked ballot view will update it's description depending on which method is being used

mikefranze: Ah, didn't see that.

mikefranze (approved):

PR #235, opened 2023-02-28
Results pages and tabulator bug fixes

mikefranze:
Some quick fixes to some tabulator bugs so that they'll work, but still need some more refactoring and unit tests.

Adding some basic results pages for other voting methods.

Approval
image

Plurality
image

IRV
image


ArendPeter (approved):
Looks good!

PR #241, opened 2023-03-11
More plurality and approval bug fixes, Ranked Robin name fix

mikefranze:
Plurality and Approval are still buggy regarding ties. This randomizes the order of tied candidates and grabs the first tied candidate if random tiebreakers are enabled. Won't work if random tiebreakers are not enabled so it will need to be refactored some time in the future, but this temp fix will work in the meantime.

Changed Ranked Robin validity check to count 0 as valid. It already checks for 0 ranks later in the code and considers them as last place ranks.

Also changed Ranked-Robin to RankedRobin in the frontend. The '-' doesn't work with the VotingMethodSelector in the backend.


ArendPeter (approved):
Looks good!

PR #244, opened 2023-03-12
Log scores correctly in runoff tiebreaker

mikefranze:
Currently logs objects: "Cat defeats Dog in runoff tiebreaker with a score of [object Object] to [object Object]"


ArendPeter (approved):
Good catch!

PR #249, opened 2023-03-19
Voting method radio button and results page crop fix

mikefranze:
Uses radio buttons for selecting voting method
https://files.slack.com/files-pri/T9TFU145P-F04VB54EJQY/image.png

Fixes issue with the edges of the results page getting cropped on mobile. I switched to using the MUI table component and set a max table width based on screen size. So now if the table is too big it creates a scrollable table that fits on the screen.

Original view:
image

Fixed view:
image


ArendPeter (approved):
Looks good! Thanks for the update

PR #254, opened 2023-03-23
Moving frontend keycloak url to env variable 

mikefranze:
Figured out that you can use env variables in react apps that get embedded at build time. Figured the keycloak url would be a good use for that.
Variables have to start with "REACT_APP_" which is why I can't use the existing KEYCLOAK_URL variable used in the backend.
Worked locally, creating PR to test in heroku.


mikefranze:
Worked on heroku

ArendPeter (approved):
Looks good!

PR #255, opened 2023-03-25
Moving tests to dev dependencies

mikefranze:
Moving tests to dev dependencies, trying to reduce size of node modules in production
Also we don't use the star_tabulators package anymore. Maybe in the future we can move our current tabulators back to the star core repo and rerelease the updated npm package.


ArendPeter (approved):
Looks good!

PR #264, opened 2023-04-29
Add election page support for temp-users

ArendPeter:
Temp users are anyone who hasn't logged in. They are able to create elections using quick poll, but they don't have a way to view elections they've created or vote in public elections

This PR updates the elections page to be friendlier to users who haven't logged in

The "elections you manage" section will now filter elections using their temp id, and there's also a new open elections section

image

If they open one of their previously created elections it'll take them to the voting page

### DIFF
--- a/backend/src/Services/Account/AccountService.ts
+++ b/backend/src/Services/Account/AccountService.ts
@@ -82,6 +82,13 @@ export default class AccountService {
             const key = customKey ? customKey : this.publicKey;
             return AccountServiceUtils.extractUserFromRequest(req, token, key);
         }
+        const tempId = req.cookies.temp_id;
+        if(tempId){
+            return {
+                'typ': 'TEMP_ID', // keycloak populates typ with ID, so I'm trying to follow a similar convention
+                'sub': tempId
+            }
+        }
         return null
     }
 }
--- a/backend/src/Controllers/controllerUtils.ts
+++ b/backend/src/Controllers/controllerUtils.ts
@@ -10,15 +10,6 @@ import ServiceLocator from "../ServiceLocator";
 
 const accountService = ServiceLocator.accountService();
 
-export function expectUserFromRequest(req:IRequest ):any {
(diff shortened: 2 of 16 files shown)


mikefranze:
Looks good, I can't test at the moment but from the admin page are you able to access voter or ballot data? Since there's no authentication on the temp id we don't want that to be accessible.

ArendPeter:
> Looks good, I can't test at the moment but from the admin page are you able to access voter or ballot data? Since there's no authentication on the temp id we don't want that to be accessible.

Good point, will fix

ArendPeter:
When applying the changes I realized it doesn't make sense to expose the admin page to temp users at all

I updated the backend to not give permissions to temp users. Temp users will still see their elections in the list, but the link will take them to the voting page instead of the admin page

backend/src/Services/Account/AccountService.ts
```diff
@@ -82,6 +82,13 @@ export default class AccountService {
...
             return AccountServiceUtils.extractUserFromRequest(req, token, key);
         }
+        const tempId = req.cookies.temp_id;
+        if(tempId){
+            return {
+                'typ': 'TEMP_ID', // keycloak populates typ with ID, so I'm trying to follow a similar convention
+                'sub': tempId
+            }
+        }
         return null
```
mikefranze, line 92: One thing to look out for here is places where we are expecting to return null in the user is not logged in, like the expectUserFromRequest function in controllerUtils.ts. 

You might also want to add the temp_id cookie to the backend test requests as a default to see if that breaks anything.

ArendPeter: Good point

controllerUtils.ts doesn't seem to be used so I removed it to avoid confusion. I didn't see any similar assumptions with other calls to extractUserFromRequest

I also added a bunch of tests checking for temp user scenarios

mikefranze (approved):

PR #268, opened 2023-05-22
Adding Typescript version of STAR-PR with tests

mikefranze:
Updating STAR-PR (Allocated Score) to typescript with a few basic tests.

Currently the random tiebreaker works, but there are some cases where numerical errors can prevent ties from being found (https://github.com/Equal-Vote/starpy/issues/23). This isn't a huge deal at the moment since the ties will be broken randomly and the numerical errors are probably random, but it'll be best to be able to at least identify ties. 

I suspect it is related to or exacerbated by the scores normalization, so removing that will be the first step.


backend/src/Tabulators/AllocatedScore.ts
```diff
@@ -0,0 +1,369 @@
...
+import { IparsedData } from './ParseData'
+const ParseData = require("./ParseData");
+declare namespace Intl {
+    class ListFormat {
+        constructor(locales?: string | string[], options?: {});
+        public format: (items: string[]) => string;
+    }
+}
+// converts list of strings to string with correct grammar ([a,b,c] => 'a, b, and c')
+const formatter = new Intl.ListFormat('en', { style: 'long', type: 'conjunction' });
```
ArendPeter, line 12: Is this variable used? The logic seems like a front end problem, so I was surprised to see it here

ArendPeter (commented):
Looks good overall

I just did a quick scan over the implementation, but if the tests pass then I trust it

backend/src/Tabulators/AllocatedScore.test.ts
```diff
@@ -0,0 +1,92 @@
...
+            [0, 0, 4, 5]]
+        const results = AllocatedScore(candidates, votes, 2, false, false)
+        expect(results.elected.length).toBe(2);
+        expect(results.elected[0].name).toBe('Allison');
+        expect(results.elected[1].name).toBe('Doug');
+        expect(results.summaryData.weightedScoresByRound[0]).toStrictEqual([25, 24, 24, 23]);  
+        expect(results.summaryData.weightedScoresByRound[1]).toStrictEqual([0, 0, 16, 23]);  
+    })
+    test("Fractional surplus", () => {
+        // Two winners, two main parties, Allison wins first round with highest score, Allison has 8 highest level supporters, more than the quota of 6 voters
```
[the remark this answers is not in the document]
mikefranze: Good question, it is handled slightly different. Added extra test case.

backend/src/Tabulators/AllocatedScore.ts
mikefranze: It was accidently copied over from the STAR tabulator which uses it to write human readable logs. It could possibly be used in all the tabulators but removed it for now.

ArendPeter (approved):
Thanks for adding the test, approved!

PR #269, opened 2023-05-26
Ballot data csv

mikefranze:
Adds link in ballot page to download csv of ballot data. Data includes just ballot ids and scores. Currently only works for just one race.

### DIFF
--- a/frontend/src/components/Election/Admin/ViewBallots.tsx
+++ b/frontend/src/components/Election/Admin/ViewBallots.tsx
@@ -7,7 +7,7 @@ import Container from '@mui/material/Container';
 import { Paper, Table, TableBody, TableCell, TableContainer, TableHead, TableRow, Tooltip, Typography } from "@mui/material";
 import PermissionHandler from "../../PermissionHandler";
 import ViewBallot from "./ViewBallot";
-import { CSVLink, CSVDownload } from "react-csv";
+import { CSVLink } from "react-csv";
 
 const ViewBallots = ({ election, permissions }) => {
     const { id } = useParams();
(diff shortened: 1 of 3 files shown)


frontend/src/components/Election/Admin/ViewBallots.tsx
```diff
@@ -31,6 +34,26 @@ const ViewBallots = ({ election, permissions }) => {
         return event.toLocaleString();
     }
 
+    const buildCsvData = () => {
+        let header = [
+            { label: 'ballot_id', key: 'ballot_id' },
+            ...data.election.races[0].candidates.map((c) => ({ label: c.candidate_name, key: c.candidate_id }))
```
ArendPeter, line 40: I'm curious, when we extend this to multiple races, which approach do you think would work better?

All races in 1 CSV: we add more columns to the csv to account for all ballots variations
Separate CSVs per race: When user downloads data there's a page to select which race they want to load (or maybe it's all in a single zip)

I'm assuming we'll go with the "Separate CSVs per race" approach? I'm not sure what if there's a precedent for existing elections

ArendPeter:
For now this is only accessible from the admin page, but do you think we should add this to the view results page as well? (in a future PR) 

There can be a setting to keep the ballot data private, but I think it could be good to make it public by default

ArendPeter (approved):
Looks good!

frontend/src/components/Election/Admin/ViewBallots.tsx
mikefranze: I've been thinking about that as well. I think the best would be one file per race. That would allow you to add race configuration details to a header that a parser can read as well. I don't know of any precedent either. 
We could also have a link to download the raw json data and not worry about formatting.

mikefranze:
> For now this is only accessible from the admin page, but do you think we should add this to the view results page as well? (in a future PR)

Yea I think we should have that be an option.

PR #270, opened 2023-05-26
Random candidate order

mikefranze:
Randomizes order of candidates on ballot by default, will add setting in the future.

This is done by adding the candidate's scores to the candidate object instead of being held in a separate array. Then the array of candidates gets shuffled. When the ballot is submitted the scores get resorted into the correct order for tabulation.

This also opens up the potential to allow voters to sort candidates by score and get a better visual sense of how they've scored everyone.

### DIFF
--- a/frontend/src/components/Election/Voting/ApprovalBallotView.js
+++ b/frontend/src/components/Election/Voting/ApprovalBallotView.js
@@ -6,7 +6,6 @@ import Typography from '@mui/material/Typography';
 export default function ApprovalBallotView({
   race,
   candidates,
-  scores,
   onUpdate
 }) {
   const instructions = (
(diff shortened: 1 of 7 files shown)


frontend/src/components/Election/Voting/ApprovalBallotView.js
```diff
@@ -24,11 +23,10 @@ export default function ApprovalBallotView({
       key="approvalBallot"
       race={race}
       candidates={candidates}
-      scores={scores}
       columns={[1]}
       instructions={instructions}
       onClick={(row, score) => {
-        const newScores = [...scores];
```
ArendPeter, line 31: Good find, I like how this cleans up the BallotView calls

ArendPeter (approved):

PR #271, opened 2023-05-27
Adding time zone selector

mikefranze:
Adds time zone selector to election form. Exact placement could change.

image

Uses the luxon library to handle changes, but final datetimes are still saved as JS dates.

List of some common timeszone names is saved in Timezones.ts and input to a selector, luxon Datetime can parse these and figure out the conversions to UTC. Currently just saving the time zone to local storage but might be useful to add it the the election data object as well?

image


frontend/src/components/ElectionForm/ElectionDetails.tsx
```diff
@@ -3,25 +3,23 @@ import { useState } from "react"
...
-    }
+
     const [errors, setErrors] = useState({
         title: '',
         description: '',
         startTime: '',
         endTime: '',
     })
+
+    const [timeZone, setTimeZone] = useLocalStorage('timezone',DateTime.now().zone.name)
```
ArendPeter, line 21: Interesting that we're using localStorage here. So every time they create an election it'll remember what timezone they used, and then that'll be the default for the next election?

I think always using the device timezone would be a better default since a non-local timezone is likely to be a one-off, but I'm conflicted on that one

frontend/src/components/ElectionForm/TimeZones.ts
```diff
@@ -0,0 +1,137 @@
+export const timeZones = [
+    { ID: 'Pacific/Midway', name: 'Midway Island, Samoa' },
```
ArendPeter, line 2: This list could get hard to managed over time

I'm surpised Luxon doesn't have an API for getting a list of timezones, but there are some workarounds listed here https://github.com/moment/luxon/issues/353

mikefranze: I read through that thread a bit before. The issue is that I don't want to have a list of ALL timezones, just the most common ones that will allow 99.9% of users to find one so users don't have to scroll through 300 time zones. I should also note that the default time zone the browser uses will be selected first and doesn't have to be in that list.

This list is what Zoom uses, so I'm fairly comfortable using it. But getting a managed library would probably be better.

frontend/src/components/ElectionForm/ElectionDetails.tsx
mikefranze: Hmm, so I had just used state here originally but ran in the issue where if you go to the next page and then back it would reset the time zone to the device's which was annoying. Maybe the least annoying thing would be to add it to a parameter in the settings object and default it to your local browser. Then it would persist just for that election across all users/devices.

frontend/src/components/ElectionForm/TimeZones.ts
ArendPeter: Sounds good, I was worried we'd be signing up to keep this in sync as new timezones are modified or added but if it's a subset then that's easier

mikefranze: Yea definitely just a subset

frontend/src/components/ElectionForm/ElectionDetails.tsx
```diff
@@ -127,28 +125,48 @@ export default function ElectionDetails({ election, applyElectionUpdate, getStyl
...
+                    <Select
+                        labelId="demo-simple-select-label"
+                        id="demo-simple-select"
+                        value={timeZone}
+                        label="Time Zone"
+                        onChange={(e) => setTimeZone(e.target.value)}
+                    >
+                        <MenuItem value={timeZone}>{timeZone}</MenuItem>
+                        {timeZones.map(tz =>
+                            <MenuItem value={tz.ID}>{tz.name}</MenuItem>
```
ArendPeter, line 141: Can we include the timezone offset in the listing? That way if someone doesn't see their location in the list, they can pick the one that matches their timezone offset

Zoom does this as well
image

I could see this being complicated since the offsets change with day light savings, so I'm also ok with making an issue for now and leaving it for a future PR

frontend/src/components/ElectionForm/ElectionDetails.tsx
ArendPeter: Yeah I like that, I could see that information being handy in the future

frontend/src/components/ElectionForm/ElectionDetails.tsx
mikefranze: Hmm, I originally didn't do that because they change but I can see the utility of it. Lets make an issue for it for now.

PR #276, opened 2023-06-03
More typescript

mikefranze:
Initial pass at adding more typescript to front end app starting with higher level components and hooks. Other PRs soon will be built off this. I probably won't merge this on its own, but adding for easier reviewing.


ArendPeter (approved):
Nice work!

PR #277, opened 2023-06-03
Admin home rework

mikefranze:
Built off of #276. Updates to admin home page to reduce repeated code and makes recommended changes from Sara along with more typescript updates. Screen captures [here ](https://docs.google.com/presentation/d/15uccXaYIh1J4j63yYYyZONJDh1cx8qi6KQ3FhFziPBQ/edit#slide=id.p)

PR #278, opened 2023-06-04
Election form updates

mikefranze:
Built off of #277 

Adds checkbox to enable start and end times
Some styling changes to race page and button to hide some voting methods

Screen captures [here](https://docs.google.com/presentation/d/15uccXaYIh1J4j63yYYyZONJDh1cx8qi6KQ3FhFziPBQ/edit#slide=id.g24e81a767ec_1_8)

### DIFF
--- a/frontend/src/components/ElectionForm/ElectionDetails.tsx
+++ b/frontend/src/components/ElectionForm/ElectionDetails.tsx
@@ -3,7 +3,7 @@ import { useState } from "react"
 import Grid from "@mui/material/Grid";
 import TextField from "@mui/material/TextField";
 import FormControl from "@mui/material/FormControl";
-import { Checkbox, FormControlLabel, FormGroup, FormHelperText, InputLabel, MenuItem, Select } from "@mui/material"
+import { Checkbox, Divider, FormControlLabel, FormGroup, FormHelperText, InputLabel, MenuItem, Select } from "@mui/material"
 import { StyledButton } from '../styles';
 import { Input } from '@mui/material';
 import { DateTime } from 'luxon'
@@ -197,7 +197,8 @@ export default function ElectionDetails({ election, applyElectionUpdate, getStyl
                                 label="Time Zone"
                                 onChange={(e) => applyElectionUpdate(election => { election.settings.time_zone = e.target.value })}
                             >
-                                <MenuItem value={timeZone}>{timeZone}</MenuItem>
+                                <MenuItem value={DateTime.now().zone.name}>{DateTime.now().zone.name}</MenuItem>
+                                <Divider />
                                 {timeZones.map(tz =>
                                     <MenuItem value={tz.ID}>{tz.name}</MenuItem>
                                 )}
--- a/frontend/src/components/ElectionForm/ElectionDetails.tsx
+++ b/frontend/src/components/ElectionForm/ElectionDetails.tsx
@@ -8,9 +8,19 @@ import { StyledButton } from '../styles';
 import { Input } from '@mui/material';
 import { DateTime } from 'luxon'
 import { timeZones } from './TimeZones'
+import { Election } from '../../../../domain_model/Election';
 
-export default function ElectionDetails({ election, applyElectionUpdate, getStyle, onBack, onNext }) {
-    const dateToLocalLuxonDate = (date: Date|string, timeZone: string) => {
+
+type Props = {
+    election: Election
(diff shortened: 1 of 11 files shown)


frontend/src/components/ElectionForm/ElectionDetails.tsx
```diff
@@ -54,10 +69,6 @@ export default function ElectionDetails({ election, applyElectionUpdate, getStyl
                 newErrors.startTime = 'Invalid date';
                 isValid = 0;
             }
-            else if (election.start_time < new Date()) {
-                newErrors.startTime = 'Start date must be in the future';
```
ArendPeter, line 58: I'm curious why this was removed, did we identity scenarios where users would want to set a start date in the past?

frontend/src/components/ElectionForm/ElectionDetails.tsx
```diff
@@ -140,74 +151,111 @@ export default function ElectionDetails({ election, applyElectionUpdate, getStyl
...
+                                onChange={(e) => {
+                                    setEnableStartEndTime(e.target.checked)
+                                    if (e.target.checked) {
+                                        applyElectionUpdate(election => {
+                                            election.start_time = DateTime.now().setZone(timeZone, { keepLocalTime: true }).toJSDate()
+                                            election.end_time = DateTime.now().plus({ days: 1 }).setZone(timeZone, { keepLocalTime: true }).toJSDate()
+                                        })
+                                    }
+                                    else {
+                                        applyElectionUpdate(election => {
```
ArendPeter, line 172: nit

Looks like unchecking and checking the box clears the start and end time? I feel like it would be a bad experience if someone accidentally unchecked the box and then lost their dates. Ideally it should remember the previously set date when someone rechecks the box

Fixing this would add some complexity since we'd need a to add a boolean to the election settings for whether start/end times are enabled, so I think this is fine for now, but I wanted to bring it up anyway

frontend/src/components/ElectionForm/ElectionDetails.tsx
```diff
@@ -140,74 +151,111 @@ export default function ElectionDetails({ election, applyElectionUpdate, getStyl
...
+                        <FormControl fullWidth>
+                            <InputLabel id="demo-simple-select-label">Time Zone</InputLabel>
+                            <Select
+                                labelId="demo-simple-select-label"
+                                id="demo-simple-select"
+                                value={timeZone}
+                                label="Time Zone"
+                                onChange={(e) => applyElectionUpdate(election => { election.settings.time_zone = e.target.value })}
+                            >
+                                <MenuItem value={timeZone}>{timeZone}</MenuItem>
```
ArendPeter, line 198: nit

If timezone is in the timezones list would there be duplicate timezones in the list? 

I think that's probably ok. When I'm entering my country in a form I never know if they'll have USA on top for convenience or if I should search for it alphabetically. So I think having duplicates might be a better experience

That said, is it possible to add a divider to the list? That way we can communicate that the first item in the list has some significance

ArendPeter (approved):
Looks good! just a few nit picks

frontend/src/components/ElectionForm/ElectionDetails.tsx
mikefranze: I wasn't sure if this really helps anything. One thing I wanted with the enable checkbox was to auto-populate with default times so its clear what needs to be entered. The current time seemed like the least annoying option for people that just want to set an end time, but that would potentially trip this error. Worst case the election just starts immediately.

frontend/src/components/ElectionForm/ElectionDetails.tsx
mikefranze: I think if we wanted that I would just put the values in a state hook, then when enabled it would use that state as a default. I'll think on this a bit.

frontend/src/components/ElectionForm/ElectionDetails.tsx
mikefranze: Yes, whatever is currently selected will be at the top of the list. But thinking about that more I think the value at the top of the list should be the local time zone. Or we could have both current and local at the top. A divider is a good idea, I'll need to see if that is possible.

mikefranze: Changed top of list to local time zone and added divider.

ArendPeter (approved):
Thanks for making the change, approved!

PR #273, opened 2023-06-02
Fake star.vote page to prompt users

ArendPeter:
This PR is an intermediate phase for when we're transitioning off of the original star.vote.

Ideally we can add a "Try the Beta!!" button on star.vote but it's tricky for us to update that page. Instead I propse the following

Phase 1 - Optional Beta:
 * Point the star.vote dns to the star.vote 2.0 website
 * Point the classic.star.vote dns to the original star.vote
 * New users for star.vote 2.0 will be shown a fake copy of the original star.vote with a prompt to try the beta. (This gives the impression that they're still on the original star.vote site)
 * star.vote 2.0 site will have a button to try classic.star.vote

Phase 2 - Remove prompt:
* Once we leave the beta phase we'll remove the popup

Here's how the prompt looks with this PR

image

fixes #430 

### The original PR from months ago for reference
image

PR #279, opened 2023-06-04
Archive election

mikefranze:
Built on top of #278 

Adds ability to archive elections. This changes the state to archived and hides it in the my election table and prevents further changes or ballots to be submitted. 

Added a button in the my elections table as well as a button in the admin page. Both prompt the user to confirm that they want to archive.

### DIFF
--- a/frontend/src/components/Election/Admin/AdminHome.tsx
+++ b/frontend/src/components/Election/Admin/AdminHome.tsx
@@ -309,6 +309,38 @@ const TogglePublicResultsSection = ({ election, permissions, togglePublicResults
     />
 }
 
+const ArchiveElectionSection = ({ election, permissions, archiveElection }: { election: Election, permissions: string[], archiveElection: Function }) => {
+    return <Section
+        Description={
+            (<>
+                <Typography variant="h5">
+                    Archive
+                </Typography>
+                <Typography variant="body1" sx={{ pl: 2 }}>
+                    Achives election, preventing future changes and hiding it from on the elections page
+                </Typography>
+                {!hasPermission(permissions, 'canEditElectionState') &&
+                    <Typography variant="body1" sx={{ color: 'error.main', pl: 2 }}>
+                        You do not have the correct permissions for this action
+                    </Typography>
+                }
+            </>)}
+        Button={(<>
+            <StyledButton
+                type='button'
+                variant='contained'
+                disabled={!hasPermission(permissions, 'canEditElectionState')}
+                fullwidth
+                onClick={() => archiveElection()}
+            >
+                Archive
+            </StyledButton>
+
+        </>)}
+    />
+}
+
+
 const ShareSection = ({ election, permissions }: { election: Election, permissions: string[] }) => {
     return <Section
         Description={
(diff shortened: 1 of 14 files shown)


frontend/src/components/Election/Admin/AdminHome.tsx
```diff
@@ -341,6 +374,20 @@ const AdminHome = ({ election, permissions, fetchElection }: Props) => {
             console.log(err)
         }
     }
+
+    const archiveElection = async () => {
+        console.log("archiving election")
+        if (window.confirm('Are you sure you wish to archive this election? This action cannot be undone')){
```
ArendPeter, line 380: nit

Using the default popup is fine for now, but we should probably implement our own in the future

Maybe I can iterate on my popup in [this pr](https://github.com/Equal-Vote/star-server/pull/273) to make it a reusable component

frontend/src/components/Elections.tsx
```diff
@@ -79,15 +89,16 @@ const Elections = ({ authSession }) => {
                                     <TableCell > {election.start_time ? new Date(election.start_time).toLocaleString() : ''}</TableCell>
                                     <TableCell >{election.end_time ? new Date(election.end_time).toLocaleString() : ''}</TableCell>
                                     <TableCell >{limit(election.description, 30) || ''}</TableCell>
-                                    <TableCell ><Button variant='outlined' href={`/Election/${String(election.election_id)}${authSession.isLoggedIn()? '/admin' : ''}`} > View </Button></TableCell>
+                                    <TableCell ><Button variant='outlined' href={`/Election/${String(election.election_id)}${authSession.isLoggedIn() ? '/admin' : ''}`} > View </Button></TableCell>
```
ArendPeter, line 92: I don't think the archival button needs to be on the elections page. It makes sense for us since we're frequently bulk creating and archiving elections, but not for the common user

frontend/src/components/Elections.tsx
```diff
@@ -101,7 +112,7 @@ const Elections = ({ authSession }) => {
                             <TableCell> View </TableCell>
                         </TableHead>
                         <TableBody>
-                            {data?.elections_as_voter?.map((election: Election) => (
+                            {data?.elections_as_voter?.filter(election => election.state != 'archived').map((election: Election) => (
```
ArendPeter, line 115: nit

we can probably filter this on the backend within the getElections api, but this is fine for now

ArendPeter (changes requested):
I left some comments, I think we should consider removing the archive button from the elections page

frontend/src/components/Election/Admin/AdminHome.tsx
mikefranze: Yea I looked at some options for a pop up but it was more than I wanted for this PR. Having a reusable component would be nice.

frontend/src/components/Elections.tsx
mikefranze: Fair, I really only added it so I could declutter my test elections faster.

frontend/src/components/Elections.tsx
mikefranze: One thing I'd like to do is add the ability to filter, sort, and add pagination to these tables. I'm thinking the part about filtering out archived elections would just be a preset to those filters.

frontend/src/components/Elections.tsx
mikefranze: Done

ArendPeter (approved):
Thanks for making the change, approved!

PR #287, opened 2023-06-10
Adding Footer

mikefranze:
Adds basic footer, recommendations for what else to put are welcome. 
Some potentials would be site navigation, GitHub contribution info, copyright info, feedback options.

Also changes STAR Voting Elections Hosting to STAR Elections, will likely change again later.

image
image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -39,25 +39,25 @@ const App = () => {
             </Alert>
           </Snackbar>
           <CssBaseline />
-          <Header authSession={authSession} />
-          <Box
-            sx={{
-              width: '100%',
-              minHeight: `calc(100vh - 255px - 64px)` //Sets min height of content to 100% minus footer and header in order to force footer to the bottom
-            }}>
-            <Routes>
-              <Route path='/' element={<LandingPage authSession={authSession} />} />
-              <Route path='/Elections' element={<Elections authSession={authSession} />} />
-              <Route path='/Login' element={<Login />} />
-              <Route path='/Debug' element={<DebugPage authSession={authSession} />} />
-              <Route path='/CreateElection' element={<AddElection authSession={authSession} />} />
-              <Route path='/Election/:id/*' element={<Election authSession={authSession} />} />
-              <Route path='/DuplicateElection/:id' element={<DuplicateElection authSession={authSession} />} />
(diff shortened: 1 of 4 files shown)


frontend/src/App.tsx
```diff
@@ -39,16 +40,24 @@ const App = () => {
...
-            <Route path='/Debug' element={<DebugPage authSession={authSession} />} />
-            <Route path='/CreateElection' element={<AddElection authSession={authSession} />} />
-            <Route path='/Election/:id/*' element={<Election authSession={authSession} />} />
-            <Route path='/DuplicateElection/:id' element={<DuplicateElection authSession={authSession} />} />
-            <Route path='/Sandbox' element={<Sandbox />} />
-          </Routes>
+          <Box
+            sx={{
+              width: '100%',
+              minHeight: `calc(100vh - 255px - 64px)` //Sets min height of content to 100% minus footer and header in order to force footer to the bottom
```
ArendPeter, line 46: This height logic will probably break with the new responsive header that can take multiple rows

Maybe you can put the header, middle, and footer all in a flex box and set the middle to flex grow?

frontend/src/App.tsx
```diff
@@ -39,16 +40,24 @@ const App = () => {
...
+              <Route path='/Login' element={<Login />} />
+              <Route path='/Debug' element={<DebugPage authSession={authSession} />} />
+              <Route path='/CreateElection' element={<AddElection authSession={authSession} />} />
+              <Route path='/Election/:id/*' element={<Election authSession={authSession} />} />
+              <Route path='/DuplicateElection/:id' element={<DuplicateElection authSession={authSession} />} />
+              <Route path='/Sandbox' element={<Sandbox />} />
+            </Routes>
+
+          </Box>
+          <Footer />
```
ArendPeter, line 60: I was thinking about if we wanted to do a call to action for contributions somewhere in the footer, but I don't think we need one. I elaborated on my thoughts [here](https://github.com/Equal-Vote/star-server/issues/289)

ArendPeter (changes requested):
Nice work! Just left a comment about the minHeight calculation

frontend/src/App.tsx
mikefranze: Done, but by setting footer margin top to auto.

Issue #281, opened 2023-06-04
Ballot formatting and font sizing


SaraWolk:
I think the poll title is bigger than it needs to be. Also some fine tuning on the size and centering of the numbers in the stars on the ballot is needed to get it fitted and centered so the numbers are as big as they can be without touching the stars and with a nice white border around them for clarity. 
<img width="1511" alt="Screenshot 2023-06-04 at 1 27 45 PM" src="https://github.com/Equal-Vote/star-server/assets/75465271/bc5aee9f-58bd-43e6-b040-dc0f9a6f9b38">

PR #288, opened 2023-06-11
Ballot updates

mikefranze:
Addresses #281 
- Reduces size of ballot title font and increases padding.
- Fixes issues with stars numbers shifting to the right on smaller screens and attempts to center them more. The centering is a little iffy, might need to reduce font size on smaller screens.
- Centers ballot circles/numbers a little better and aligns with stars. 
- Makes title/description/instructions/ballot spacing more even and makes instructions more responsive

Before
image

After
image


ArendPeter (approved):
Looks good!

PR #292, opened 2023-06-13
Fix bugs with yaxis cutoff and sorting candidates

### LINKED ISSUE #291: Results page funnies
Did a quick test and found some issues with new results page, can see here:
https://star-vote-staging.herokuapp.com/Election/7c4a5443-31b2-4212-91df-4d0e02d64475/results

This uses the dev dB for debugging.

First it looks like the names are clipping on the left.
Second, it shows leia and yoda going to the runnoff in the score round but actually leia and luke. This might be due to the fact that the candidate order is not by score but by what order they'd win if you kept running the election.
Third the sections of the pie chart don't line up with the votes in the runoff and chart.

Maybe the fix for this is how we sort the data in the results, or in how it's displayed. Not sure.

image

image

ArendPeter:
looks like sorting fixed it

image

I also added some automatic scrolling logic, and removed some excess padding from the results (it was bothering me that I had to scroll to see the full results on small elections)

https://github.com/Equal-Vote/star-server/assets/9289903/38aff004-406f-42da-a4b7-06de8ab02f1b

fixes #291

### DIFF
--- a/frontend/src/components/Election/Results/DetailedResultsExpander.tsx
+++ b/frontend/src/components/Election/Results/DetailedResultsExpander.tsx
@@ -7,8 +7,42 @@ const DetailedResultsExpander = ({children, defaultSelectedIndex}) => {
     const [viewDetails, setViewDetails] = useState(false);
     const [widgetIndex, setWidgetIndex] = useState(defaultSelectedIndex);
 
+
+    function scrollToElement(e){
+        setTimeout(() => {
+            // TODO: I feel like there's got to be an easier way to do this
+            const openedSection = (e as HTMLElement);
+
+            const elemTop = document.documentElement.scrollTop + openedSection.getBoundingClientRect().top;
+            const elemBottom = elemTop + openedSection.scrollHeight;
+            const windowTop = document.documentElement.scrollTop
+            const windowBottom = windowTop + window.innerHeight;
+
+            // scroll down if the element is below the window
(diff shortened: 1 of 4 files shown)


frontend/src/components/Election/Results/DetailedResultsExpander.tsx
```diff
@@ -7,16 +7,55 @@ const DetailedResultsExpander = ({children, defaultSelectedIndex}) => {
     const [viewDetails, setViewDetails] = useState(false);
     const [widgetIndex, setWidgetIndex] = useState(defaultSelectedIndex);
 
+
+    function scrollToElement(e){
```
mikefranze, line 11: What does this do?

mikefranze:
I think this will work for a single winner results widget, but we'll want to think about multiwinner as well

ArendPeter:
When you expand the detailed sections it'll automatically scroll to make
sure the section is fully in view

I added a video in the description but it's probably hard to tell which
parts of the scrolling is automatic

mikefranze:
Gotcha, sorry overlooked that part of the description.

mikefranze (approved):

ArendPeter:
> I think this will work for a single winner results widget, but we'll want to think about multiwinner as well

Makes sense, I'll start on the STAR PR dashboard soon, and I'll make sure the final solution supports both

PR #293, opened 2023-06-15
Thanks page spacing and responsivity

mikefranze:
Adjusting thanks page spacing and responsivity. Goal was to have more evenly spaced components. At the bottom we can have from 1 to 3 buttons depending on the election so there was some jankiness to get them to be centered and evenly spaced but also switch to columns on xs screens.

image

image


ArendPeter:
Looks good!

ArendPeter (approved):
Looks good! 

I like that you included both desktop and mobile screenshots, I should get into that habit as well. I think I'll try making a pull request template to help us enforce good habits 🤔

PR #294, opened 2023-06-16
Add pull request template

ArendPeter:
This adds a template so future PRs will have default text (at least I think that's how it works)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,7 +3,7 @@
 
 ## Screenshots / Videos (frontend only) 
 
-> Be sure to include both desktop and mobile views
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+## Description
+
+
+## Screenshots / Videos (frontend only) 
+
+> Be sure to include both desktop and mobile views
```
mikefranze, line 6: Should specify smallest mobile screen size is 320 pixels if possible. Some pages we could assume desktop view though.

ArendPeter: Added a note

I felt like it got too wordy when mentioning that some pages would only need desktop. I figure people can deviate if that's the case since it's just meant to be a template

mikefranze (approved):

PR #295, opened 2023-06-17
Email invites rework with job manager

mikefranze:
This addresses a couple a couple concerns I had with sending bulk email invites and is probably overkill. 

Some concerns were:
- hitting the 1000 email per request cap. to get around this we'd have to send multiple requests, but then we'd have to track which emails were successful in case some of the batches failed to send.
- not saving the response status of individual emails. Before if any of the email API responses were bad (status>400) we'd just log it and throw an error, but that's not very visible or useful for the user.
- resending invitations after finalizing the election/sending invitations to anyone who's invite failed to send.

What these changes do is 
- creates an email invite job queue for sending email invitations. 
- When the election is finalized all invitations will be added to the queue
- The pg-boss job manager will grab a couple of these at a time and make requests to the email service API.
- For each email, the response will be saved to the voter's election roll dB and a history entry will be added.
- Adds another API route that will look for voters who haven't been emailed successfully and add them to the queue.

I ran a couple tests with a temp job queue where each job took 100ms and the server is able to process several hundred a minute so while this does require a lot more work I think it's more robust.

A few TODOs
- [x] Email invite wordsmithing and styling. There's also an HTML tag that we need to add so users can unsubscribe which will help keep us out of the spam box.
- [x] Sendgrid has a "sandbox" option in their API that will respond with a valid status without actually sending the email or costing anything. I'm thinking of adding a flag that will let us test the system with a few thousand voters without actually sending emails.
- [x] Some basic tests using real emails


mikefranze:
Update: using sandbox mode was able to send 2000 invites to fake emails and save all responses in ~10 minutes

mikefranze:
New invitation desktop view
image

Description and start/end times are only shown in specified. Times also only shown if time zone defined.

mikefranze:
Mobile view 

image

ArendPeter (approved):
Looks great!

PR #299, opened 2023-06-19
Candidate photo positioning

ArendPeter:
## Description
I made an improvement to the candidate photo positioning on the election form

https://github.com/Equal-Vote/star-server/assets/9289903/649a62a2-21a8-4842-ba3d-3a197ac5622e

### DIFF
--- a/frontend/src/components/ElectionForm/AddCandidate.tsx
+++ b/frontend/src/components/ElectionForm/AddCandidate.tsx
@@ -81,6 +81,7 @@ const AddCandidate = ({ onEditCandidate, candidate, index }: CandidateProps) => 
         }
     }
 
+
     return (
         <>
             <Grid item xs={10} sx={{ display: "flex", alignItems: "center", m: 0, p: 1 }}>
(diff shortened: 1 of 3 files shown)


ArendPeter:
TODO: I need to move the "Select Photo" button below the picture on mobile since mobile can't hover

mikefranze:
Looks good, only comment is the alignment of the text fields on the right with the save button, but making them shorter wouldn't help. Maybe we could make them shorter and move bio below both the photo and other fields and make it full width since it will potentially be longer and multi line.

ArendPeter:
> Looks good, only comment is the alignment of the text fields on the right with the save button, but making them shorter wouldn't help. Maybe we could make them shorter and move bio below both the photo and other fields and make it full width since it will potentially be longer and multi line.

Pushed a change with some updates
 * I tried tweaking the bio, but I had trouble making it look good, so I opted to make it taller to communicate that it can be longer
 * Personally I don't think it matters too much for the buttons to align with the buttons, since they're separate columns with no direct relation
 * I moved the candidate photo title within the image area, and moved the select file button outside the image once the photo has been selected 
 
 **with image and hover**
 
 
image

**no image**
image

**mobile**
image

mikefranze (approved):
Looks good!

PR #300, opened 2023-06-19
Expandable race results

ArendPeter:
## Description

This updates the race results page to put each race in a expandable area. This also adds scrolling logic to the election form when errors are made in the form

https://github.com/Equal-Vote/star-server/assets/9289903/0c89341c-64e7-4827-bde1-25b66ea1a937

Note: The bug at the end of the video is also fixed in this PR


mikefranze (approved):
Looks good!

PR #301, opened 2023-06-20
Block result tabs

ArendPeter:
## Description
Add round tabs to block star results

> NOTE: The vader star is cut off in a lot of these screenshots, but I fixed that in the PR

> NOTE2: These results also show the same numbers across rounds, this is fixed in the PR as well

**Desktop tabs**
image

**Mobile tabs**
image

**Example with many rounds**
image

**Single winner star, no tabs**
image


frontend/src/components/Election/Results/STARResultSummaryWidget.tsx
```diff
@@ -22,29 +22,37 @@ const renderCustomizedLabel = ({ cx, cy, midAngle, innerRadius, outerRadius, per
...
+        .map(winner => winner.index);
+
+    const histData = results.summaryData.candidates
+        .map((c, n) => ({
+            name: c.name,
+            votes: results.summaryData.totalScores[n].score,
+            // vvvv HACK to get the bars to fill the whole width, this is useful if we want to test the graph padding
+            votesBig: results.summaryData.totalScores[n].score*10000 
+        }))
+        .filter((_, i) => !prevWinners.includes(i));
```
mikefranze, line 39: I'm wondering if instead of filtering out candidates we should just mark previous winners or grey out their scores.

frontend/src/components/Election/Results/RoundResultTabs.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const [currentTab, setCurrentTab] = useState(0);
+
+    if(rounds == 1) return <>{children}</>
+
+    let i = 0;
+    const roundIndexes = Array.from({length: rounds}, () => i++);
+
+    return <>
+        {rounds >= 5 && 
+            <Typography variant="h6">Rounds</Typography>
```
mikefranze, line 15: We might want to use a different word than Rounds here. While we use it in the code as one winner per round some people might get it confused with score/runoff rounds.

ArendPeter: Agreed, changed it to winner in the latest update

frontend/src/components/Election/Results/STARResultSummaryWidget.tsx
ArendPeter: We discussed this in a meeting, and decided to leave it as is. Greying out candidates could create confusion at what the stars mean in the scoring round

ArendPeter:
Here it is with the latest updates

image

image

mikefranze (approved):

PR #303, opened 2023-06-23
Adding fraction math to allocated score

mikefranze:
## Description
Adds fractional math to allocated score. Because of normalized scores and ballot weights there's a lot of non integer math involved where precision can be an issue and ties can be hard to determine. This converts most of the internal data types to fractions which gets around floating point precision issues. All outputs are converted back to floating point.

## Related Issues
https://github.com/Equal-Vote/starpy/issues/23


ArendPeter (approved):
Looks good!

I didn't review super closely but if the tests passed then I trust it

PR #304, opened 2023-06-23
Email updates

mikefranze:
## Description
Adds route for sending individual invite 
Displays email invite status on individual voter page. If email failed will display debug info.
Adds button on individual voter page to send invite. This will send even if previous invite was successful. 

image
image

image
image

## Related Issues
none


ArendPeter (approved):
Looks great! I can see how this would be really useful for the administrators

PR #307, opened 2023-06-25
Make useFetch types generic, move API  hooks to one file

mikefranze:
## Description
In order to make the front end more type safe this makes the useFetch hook type generic and allows us to specify the expected request body and response types. This PR also moves all api requests to our server to a new file useAPI.ts and defines all the expected types (the election results being the exception and is just type any for now).

This will allow us to have our API defined in one place, minimize redundant code, and clean up our components that use the API hooks.

Hopefully in the future this will also help us write some dedicated tests for these as well.


ArendPeter (approved):
Looks good! I like how this documents the expected return information in code, since that's usually a trial and error process for me

PR #308, opened 2023-06-26
Ballot update from feedback

mikefranze:
## Description
From Sara's feedback, adjusts some font sizing and spacing on ballot and changes order of footer sections and about us wording.

Footer
image
image

Ballot images to come soon, dev DB is not starting


ArendPeter (approved):
Looks good!

PR #334, opened 2023-07-05
Make time consistent with timezone

ArendPeter:
## Description

More date format into a shared function so it can be consistent

image

## Related Issues

#fixes 326

### DIFF
--- a/frontend/src/components/util.js
+++ b/frontend/src/components/util.js
@@ -3,6 +3,7 @@ import ExpandMore from '@mui/icons-material/ExpandMore';
 import { Grid, IconButton, Paper, Typography } from '@mui/material'
 import React, { useState, useRef, useEffect }  from 'react'
 import _uniqueId from 'lodash/uniqueId';
+import { DateTime } from 'luxon'
 
 export function scrollToElement(e){
     setTimeout(() => {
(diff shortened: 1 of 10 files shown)


frontend/src/components/util.js
```diff
@@ -94,4 +94,8 @@ export const DetailExpanderGroup = ({children, defaultSelectedIndex}) => {
             </Paper>
         ))}
     </div>;
+}
+
+export const formatDate = (time) => {
+    return new Date(time).toLocaleString(navigator.language, {timeZoneName: 'short'});
```
mikefranze, line 100: I think we should use the luxon library here and use the time zone defined by the user if possible. Its pretty customizable. Example uses of it are in the election details component and in the email invite template:

`DateTime.fromJSDate(new Date(election.start_time)).setZone(election.settings.time_zone).toLocaleString(DateTime.DATETIME_FULL)`

Only exception would be comparing start/stop times of different elections in the election tables, it would be confusing to sort and compare elections with different time zones. Maybe time zone can be an optional input to this function and we can leave it out in that case.

mikefranze (changes requested):

frontend/src/components/util.js
ArendPeter: Done!

PR #336, opened 2023-07-05
Add more snacks for button feedback

ArendPeter:
## Description

Add more snacks to improve button feedback

image

### DIFF
--- a/frontend/src/components/Election/Admin/EditElectionRoll.tsx
+++ b/frontend/src/components/Election/Admin/EditElectionRoll.tsx
@@ -7,12 +7,9 @@ import Container from '@mui/material/Container';
 import { Paper, Table, TableBody, TableCell, TableContainer, TableHead, TableRow, Tooltip } from "@mui/material";
 import PermissionHandler from "../../PermissionHandler";
 import { useApproveRoll, useFlagRoll, useInvalidateRoll, useSendInvite, useUnflagRoll } from "../../../hooks/useAPI";
-import useFetch from "../../../hooks/useFetch";
-import { SnackbarContext } from "../../SnackbarContext";
 
 const EditElectionRoll = ({ roll, onClose, fetchRolls, id, permissions }) => {
     const [updatedRoll, setUpdatedRoll] = useState(roll)
-    const { snack, setSnack } = useContext(SnackbarContext)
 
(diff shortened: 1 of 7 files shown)


frontend/src/components/Election/Admin/EditElectionRoll.tsx
```diff
@@ -36,6 +38,14 @@ const EditElectionRoll = ({ roll, onClose, fetchRolls, id, permissions }) => {
     }
     const onSendInvite = async () => {
         if (!await sendInvite.makeRequest()) { return }
+
+        setSnack({
```
mikefranze, line 42: The useFetch hook actually has this snack feature built in, you can set an optional success message on a successful request. I never got around to using it for anything.

mikefranze (changes requested):

frontend/src/components/Election/Admin/EditElectionRoll.tsx
ArendPeter: Good point, updated!

frontend/src/components/Election/Voting/Thanks.tsx
```diff
@@ -46,7 +46,7 @@ const Thanks = ({ election }) => {
                         }
                         {election.settings.voter_access !== 'closed' &&
                             <Box sx={{ width: '100%', p: 1, px:{xs: 5, sm: 1}  }}>
-                                <ShareButton url={`${window.location.origin}/Election/${election.election_id}`} text={'Invite'} />
```
mikefranze, line 49: I'm not sure about this one, I think we made a point to change this to say 'Invite' in some places to be clear about what was being shared.

ArendPeter: Ah ok

When I read Invite I thought it was confusing that it was inconsistent from all the other Share buttons. Invite also has a different meaning to me, and made me think that it wanted me to invite someone specific rather than share it to a general group

I guess "Share Election" would be the most clear, but that's a little long

mikefranze: We can stick with share for now, might revisit it later.

mikefranze (approved):

PR #338, opened 2023-07-05
Kyesly integration

mikefranze:
## Description
Built on #337, will update once that PR is merged to clear this PR up. 

This adds the kysely library which is a type safe query builder. I wanted to practice using this library for another project and figured this would help us. 

A few notes:
- Creation of the database tables is now in the new migrations folder. The npm script migrate:latest can be run to update the database to the latest version. It runs migrations in alphabetical order so use the year_month_day date format. We'll want to think about best practices for handling migrations.
- The library will look at the interfaces to determine how to create queries, so we need to make sure the interface definitions stay in sync with the schemas defined in the migration files.
- A kysely dev pointed me to a plugin that will automatically stringify any objects when inserting or updating, so we don't have to handle that anymore. 
- The query building API is amazing, makes adding, updating, and reading queries so much easier. It also uses a lot of typescript magic that will prevent you from making a lot of mistakes. 

## Related Issues

None, just general improvements.


ArendPeter:
Now that the previous PR is pushed, can you rebase this one to make the diff cleaner?

mikefranze:
> Now that the previous PR is pushed, can you rebase this one to make the diff cleaner?

@ArendPeter done

ArendPeter (approved):
Nice! This looks much cleaner!

PR #339, opened 2023-07-05
Election creation prompts

ArendPeter:
## Description
This breaks some of the longer prompts at the end of the election creation process into bullet points

## Before
image

## After
image


mikefranze (approved):
Looks good!

PR #341, opened 2023-07-06
Election home tweaks

ArendPeter:
## Description

This adds a message when the election is in draft mode and disables the voter id box if it's already been given

## Screenshots / Videos

image

### DIFF
--- a/frontend/src/components/Election/VoterAuth.tsx
+++ b/frontend/src/components/Election/VoterAuth.tsx
@@ -21,8 +21,9 @@ type Props = {
 
 const VoterAuth = ({ authSession, electionData, fetchElection }: Props) => {
   const { voter_id } = useParams();
+  console.log('useParam')
   // TODO: maybe we should reconsider useCookie here? this has the potential of inserting the voter id on a different election
-  const [voterID, setVoterID] = useCookie('voter_id', voter_id ? voter_id : null, 1)
+  const [voterID, setVoterID] = useCookie('paramVoterID', voter_id ? voter_id : null, 1)
 
   useEffect(() => {
     setVoterID(voter_id)
@@ -50,6 +51,7 @@ const VoterAuth = ({ authSession, electionData, fetchElection }: Props) => {
   const missingVoterID = electionData.voterAuth?.required === "Voter ID Required"
 
 
+  console.log(voter_id)
   return (
     <Box sx={{ p: 1, flexGrow: 1 }}>
       {isOpen &&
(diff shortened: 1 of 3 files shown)


frontend/src/components/Election/VoterAuth.tsx
```diff
@@ -67,7 +67,7 @@ const VoterAuth = ({ authSession, electionData, fetchElection }: Props) => {
               </Typography>
             </Box>
           }
-          {voterIdRequired &&
```
ArendPeter, line 70: This disables the voter id box if it was provided via a url parameter

I'm curious to hear feedback on this idea. I'm suggesting it because people generally seem confused to see a prefilled box when they open the ballot, but I could also see an argument that having more context is better

I'm also not sure if the voter needs to be aware of the voter id at all. The election admins should be aware, and voters should have a way to learn about it, but I don't think it needs to be shown during the voting process

frontend/src/components/Election/VoterAuth.tsx
```diff
@@ -21,6 +21,7 @@ type Props = {
 
 const VoterAuth = ({ authSession, electionData, fetchElection }: Props) => {
   const { voter_id } = useParams();
+  // TODO: maybe we should reconsider useCookie here? this has the potential of inserting the voter id on a different election
```
ArendPeter, line 24: I think we should consider removing the voter cookie in a future PR, curious to hear feedback on this

mikefranze: I'd be interested in alternatives. Cookies are a convenient place to store the voter_id.

frontend/src/components/Election/VoterAuth.tsx
mikefranze: This is a tough one. I think our goals should be to communicate clearly that they've been given a unique link and are authorized to vote. Currently we don't do that but hiding the box doesn't fix the issue and might make people think the opposite. We should definitely add a line in the email that this is a unique link and not to share it.

What I've seen on other forms is a brief empty form, then a filled-out form, some confirmation, then transition to the page you're trying to go to. Maybe we could put in a loading animation and then a green check or something. 

The clear button can probably be removed if we transition away from storing the voter_id in a cookie. 

You should also add a "&& isAuthorized" to this check. I think if an invalid voter_id is submitted that would be false and we'd want to render the box and error. We should also consider refactoring the voter_auth data sent from the server. It wasn't well thought out so it's a bit janky. 

For near term testing this is ok but we'll want to revisit it.

mikefranze (approved):

frontend/src/components/Election/VoterAuth.tsx
ArendPeter: my bad, I forgot that we needed to keep the voter id for the scope of the whole voting session. Cookie is fine for that case

And as long as we're clearing the cookie when id isn't in the url (which it seems like it does), then that's fine

frontend/src/components/Election/VoterAuth.tsx
ArendPeter: Hmm that is tricky

For now I'll add the && isAuthorized and adjust the language on the email to clarify that it's unique

 Also, in the future I think we should move the voter id from the url. Sensitive data in the url can be intercepted. https://docs.guardrails.io/docs/vulnerability-classes/insecure-access-control/sensitive-data-in-urls
 
Maybe we can try using post using this method https://stackoverflow.com/questions/3915917/make-a-link-use-post-instead-of-get

ArendPeter: Here's another idea, for when the parameter is specified?

image

PR #342, opened 2023-07-06
Voter auth hot fix

ArendPeter:
## Description

My previous PR broke the voter id login experience, this one should fix it


ArendPeter:
I'll go ahead an merge this since I need it for tomorrow's playtest

PR #352, opened 2023-07-15
Results types

mikefranze:
## Description
Adds types to results on frontend, changes results response format slightly to make it easier for typescript to distinguish the different race results types. No functional changes.

## Related Issues
None, but will make it easier to update results pages in the future.


ArendPeter (changes requested):
Looks good!

ArendPeter (approved):
Looks good! (the "request changes" was an accident)

Issue #330, opened 2023-07-05
Include election you administer in the elections you manage section (not just elections you own)


ArendPeter:
Adam detailed the issue here: https://docs.google.com/document/d/1OsvgLCS1VLgdW4kJH7V2g-imB2RyI86YSrnnDI1d_sM/edit

mikefranze:
This _should_ already be a feature, but I haven't tested it in a while if there's tests showing it doesn't work. It's tough because the the admin ids are stored in a json object which is harder to query.

masiarek:
Correct link:
https://docs.google.com/document/d/1OsvgLCS1VLgdW4kJH7V2g-imB2RyI86YSrnnDI1d_sM/edit

mikefranze:
@masiarek Ok, so a couple issues that I've found from your [video ](https://www.youtube.com/watch?v=c0lZ17ZJR7M)

1. When clicking submit in the add elections roles page it doesn't update the version in the browser. Normally after making a request we'll make a second request to get the latest version. So when you submit and then immediately check it doesn't appear like they were added. If you refresh the page you will see the admins. You could see later on when you went back that the field was populated.
2. When you log in from the "my elections" page, the request to fetch the elections isn't getting sent, or is getting sent before the user cookie is saved, or something. So it looks like your other users don't have any elections to admin. If you refresh you should see them. 
3. When you went to the election url while logged in you said you "can see it but not admin it". Note that you do have the ability to edit the election, add voters (if it was the type of election where you can add voters), and access the side bar. The other actions (adding admins, finalizing, archiving) are restricted to the owner of the elections. If we want to change those permissions for different roles we can do that. Making it clearer what permissions the user does have might help. After finalizing the election admins will be able to view voters/ballots/results. 

When I go to the election not as an admin I get all red. 
![image](https://github.com/Equal-Vote/star-server/assets/41272412/077643cf-cdfc-44b8-9146-1f40fe00ef09)

In my own test I added mike@equal.vote as an admin
![image](https://github.com/Equal-Vote/star-server/assets/41272412/fcfcd33a-8274-4f99-b18d-2753b17d3ba5)

And can see it in my list after refreshing
![image](https://github.com/Equal-Vote/star-server/assets/41272412/61478c47-be73-4005-b930-2c99b8a21357)

Then after finalizing can do more 
![image](https://github.com/Equal-Vote/star-server/assets/41272412/552e5d6a-6cc5-455b-8dd1-40d53ed16e62)

[...]

PR #353, opened 2023-07-16
refetch elections on login

mikefranze:
## Description
After logging in from the my elections page, the fetch request will be made before the login logic is finished, so the "election I manage" table will be empty and requires a page refresh. This makes the authSession.isLoggedIn() output a depenency to the fetch elections useEffect hook, so if the login status chages it will make the request again. Two requests are still made, but that shouldn't be an issue.

## Related Issues
#330


ArendPeter:
Looks good!

PR #364, opened 2023-08-24
Add feature flags to environment variables

### LINKED ISSUE #335: Stop Random Display Order of Candidates (on the ballot)
Election Configuration - add Radio-button "Display Order of Candidates (on the ballot)" to allow Election Administrator to chose "FIXED order of Candidates versus RANDOM order of Candidates" on the ballot.

## As Is:
- see video https://youtu.be/kYkmK0GA2gw
- Current behavior - Candidates are *always* RANDOMIZED on the ballot 
- Missing option to chose "Display Order of Candidates (on the ballot)" 
- Missing setting "Fixed order" of Candidates on the Ballot

## Should Be:
Add radio button "Display Order (Candidates) on the Ballot" with two Options:
- "Fixed order" (default)  
- "Random order" - the sequence of Candidates is randomized for each generated (printed, displayed) Ballot

Election administrator should be allowed to chose the "Candidates Display Order on the Ballot" before Finalizing election.

The default should be "Fixed Order" (the same as entered in the Election Configuration - typically the order of Candidates is decided ahead of time by drawing lots.

Only by activating checkbox "Random Order" during Election configuration - Candidates should be randomized on the ballot.

Another benefit:
In a demo mode - we want to cast 5 ballots with the same pattern (predefined test case with predefined scores for each candidate).
Currently - with the random order - the presenter must be careful when scoring the ballot - it is easy to make mistake (Candidates change position with each ballot).

### Solution - option 1:
- Make the default "Fixed Order of Candidates = True.
- Keep current screen layout 'as-is': hide the Option "Display Order (Candidates)"  when Election is in CREATE mode
- In Election CHANGE mode - make option "Display Order (Candidates)" visible and allow  activating "Random order" = True

PROS:
- Initial CREATE mode is simple 

CONS:
- we are hiding some features that can be changed only in CHANGE mode

### Solution - option 2:
- Make the default "Fixed Order of Candidates" = True.
- User can change the option "Display Order of Candidates" both in CREATE mode and in CHANGE (until election is Finalized) 

PROS:
- same view in CREATE mode and in CHANGE mode

CONS:
- initial Election creation screen gets more complicated

# Legal / Laws
https://electionlab.mit.edu/research/ballot-order-effects 
All states have laws that mandate the order in which candidates appear on the ballot. 

If the order in which candidates appear on the ballot influences election outcomes, then these laws will advantage some candidates over others, whether this advantage is due to incumbency, party, or alphabetical order. 

Concern over the advantage presumably conveyed by “ballot-order effects” has led some states to require names to be randomly listed on the ballot (Q: by drawing lots ahead of time? - before printing ballots???).

many states list candidates on the ballot randomly, although there does not appear to be a comprehensive listing of which states randomize ballot names and how the randomization is implemented.

# Proposed User Documentation:
The Candidates listed first on a ballot often receive more votes simply because of their position on the ballot, especially if voters aren't familiar with the choices. This is called the "Ballot Order Effect".
To avoid the Ballot Order Effect - activate Random Order option (checkbox?) for displaying choices (Candidates) in a different order on each voter's ballot.

### LINKED ISSUE #355: Add feature flag support
- [ ] Roles: Admins, auditors, credentialers (auditor for example)
 - [ ] Voting methods, one for each
   - [ ] STAR-PR
   - [ ] Ranked-Robin
   - [ ] Approval
   - [ ] Plurality
   - [ ] Ranked Choice
 - [ ] Candidate Details
 - [ ] Multi Race Support
 - [ ] Multi Winner Support

ArendPeter:
## Description
This adds support for a bunch of feature flags in the environment variables. These are all features that we could potentially delay in favor of getting beta out sooner.

 - [x] Roles: Admins, auditors, credentialers (auditor for example)
 - [x] Voting methods, one for each
   - [x] STAR-PR
   - [x] Ranked-Robin
   - [x] Approval
   - [x] Ranked Choice
 - [x] Candidate Details
 - [x] Candidate Photos
 - [x] Precinct Support
 - [x] Multi Race Support
 - [x] Multi Winner Support
 - [x] Custom registration
 - [x] Voter flagging
 
I omitted plurality from the feature flags since I figured that would be the easiest to support, and I didn't want to add extra logic for removing the voting method selector. 

image

## Related Issues

fixes #355


frontend/sample.env
```diff
@@ -1,3 +1,13 @@
...
+
+FF_ELECTION_ROLES=false
+FF_MULTI_WINNER=false
+FF_METHOD_STAR_PR=false
+FF_METHOD_RANKED_ROBIN=false
+FF_METHOD_APPROVAL=false
+FF_METHOD_PLURALITY=false
+FF_METHOD_RANKED_CHOICE=false
+FF_CANDIDATE_DETAILS=false
+FF_MULTI_RACE=false
```
mikefranze, line 13: Did this work in the build process? I think React env variables have to be prefixed with "REACT_APP_" https://create-react-app.dev/docs/adding-custom-environment-variables/

frontend/src/components/ElectionForm/AddCandidate.tsx
```diff
@@ -114,11 +114,13 @@ const AddCandidate = ({ onEditCandidate, candidate, index }: CandidateProps) =>
                 </Grid>
                 :
                 <Grid item xs={2} sx={{ display: "flex", alignItems: "center" }}>
-                    <Button
-                        onClick={() => setEditCandidate(true)}
-                    >
-                        <Typography variant="h6" component="h6"> Add Details </Typography>
-                    </Button>
+                    {(process.env.FF_CANDIDATE_DETAILS || false) && <>
```
mikefranze, line 117: Does " || false" do anything here? Seems like its added as some kind of protection.

frontend/sample.env
ArendPeter: Updated to use REACT_APP_

frontend/src/components/ElectionForm/AddCandidate.tsx
ArendPeter: good point, I wanted it to default to false if it didn't exist, but in that case it would be undefined so the false is unnecessary

I ended up changing it to === 'true', I think that's the only way to handle booleans in environment variables

mikefranze (approved):
Looks good, I opened a PR to add these variables to the infrastructure so once that's merged we can merge this.

mikefranze:
Vars added to the server, so you're good to merge

PR #372, opened 2023-10-06
About goes to About page now instead of Starvoting.org

Elitemoni:
## Description
I updated the About button such that instead of going to the Star.org website, it goes to an About page. I added some placeholder text (figure out what should go here)

## Related Issues


ArendPeter (approved):
Looks good!

Issue #344, opened 2023-07-06
Add more confirmation prompts


ArendPeter:
Here's some scenarios we need to account for
 - confirmation on submit
 - confirmation when finalizing election
 - prompt for race 1 when adding the second race (we might find other solutions for this one)
 - prompt to try the beta when initially opening the site

ArendPeter:
![Image](https://github.com/Equal-Vote/star-server/assets/9289903/c8756753-b39d-4576-85eb-fb66a5bef8d5)

ArendPeter:
The popup has been designed, we just need to add the confirmations to all the places that need them

ArendPeter:
I'll label this as good-first-issue once I've found an example PR and clarified the spots that need a prompt

ArendPeter:
Actually these I'll mostly implemented, I'll just resolve this one

PR #373, opened 2023-10-16
Adding confirmation dialog provider

mikefranze:
## Description
Made this as part of a larger task. Creates a context provider and hook for accessing confirmation dialog. The current dialog is just the MUI dialog, we can update the styling using #344. Basically, it adds a single dialog component at the top level of the app and uses a context provider to allow other components to open it, update the title and message, and get the response.

Built based on this tutorial https://akashhamirwasia.com/blog/building-expressive-confirm-dialog-api-in-react/ 
Uses some useRef and useCallback optimizations that are explained in there.

image
image

### DIFF
--- a/frontend/src/components/ConfirmationDialogProvider.tsx
+++ b/frontend/src/components/ConfirmationDialogProvider.tsx
@@ -1,0 +1,84 @@
+import React, {
+    createContext,
+    useCallback,
+    useContext,
+    useRef,
+    useState,
+} from 'react';
+import { Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle } from '@mui/material'
+import { StyledButton } from './styles';
+// Built from this example buth with MUI dialogs: https://akashhamirwasia.com/blog/building-expressive-confirm-dialog-api-in-react/
+// Uses a context provider to allow any component to access a confirm dialog component using the useConfirm hook
+// Example: 
+// import useConfirm from '../../ConfirmationDialogProvider';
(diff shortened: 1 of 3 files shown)


frontend/src/components/ConfirmationDialogProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type confirmContext = (data: ConfirmData) => Promise<boolean>
+
+const ConfirmDialog = createContext<confirmContext>(null);
+
+export function ConfirmDialogProvider({ children }) {
+
+
+    const [state, setState] = useState({ isOpen: false, title: '', message: '' });
```
ArendPeter, line 33: nit

One thing we need to be careful of with this pattern is, manipulating state (like state.isOpen = true), won't trigger a refresh. Only reinstantiating the object with trigger a refresh

Looks like that's what's happening here, and I see that it makes the code cleaner, but I wanted to mention it anyway so that we're careful with it in the future

ArendPeter (approved):
Looks great! Thanks for adding this

frontend/src/components/ConfirmationDialogProvider.tsx
mikefranze: Using the setState function in the callback should reinstate the object though right?

ArendPeter: Yep, it's being reinstantiated in the callback, so it's fine

PR #374, opened 2023-10-18
Adding election context provider

mikefranze:
## Description
Part of the larger admin page work. Was getting messy passing election data and functions around between components so I made another context provider. Right now this only is used at the top level and some components I'm working on. Next will need to add the useContext hooks to the subcomponents.

None
## Related Issues
None


ArendPeter (approved):
Looks good!

PR #375, opened 2023-10-21
Block out landing page elements

ArendPeter:
## Description

I copied the landing page elements from figma: https://www.figma.com/file/Z6HA3YZz56bwjwYihmJ1Mv/STAR-Mock-Ups-W-MUI?type=design&node-id=6472-77151&mode=design&t=dKOeio58jJfkLJpg-0

https://github.com/Equal-Vote/star-server/assets/9289903/e88bf56a-648d-46bb-a35e-bbae4fdc40fd

The responsiveness is 90% working, but there's a sliver that appears on the 320px view that I'm having trouble figuring out


mikefranze:
For the sliver of white on the side, most likely one of the components is having its width or min width set manually to something larger than 320. You might be able to inspect the elements to see which one is poking out.

frontend/src/components/LandingPage/FeaturedElection.tsx
```diff
@@ -0,0 +1,17 @@
+import { Paper, Typography } from "@mui/material";
+import React from 'react';
+
+export default ({electionId}) => {
+    return <Paper className='featuredElection' elevation={8} sx={{
```
mikefranze, line 5: Try out the Material UI card component for these. It uses the paper component but with extra features.
https://mui.com/material-ui/react-card/

ArendPeter:
> For the sliver of white on the side, most likely one of the components is having its width or min width set manually to something larger than 320. You might be able to inspect the elements to see which one is poking out.

Yep it was the trashcan icon. The quick poll component was adding extra padding that was making everything a tad too wide

I also switch Paper over to Card

mikefranze (approved):

PR #376, opened 2023-10-23
Not another admin page rework

mikefranze:
## Description
Reworks the election creation process. Now the create election page shows a list of election templates to choose from that are tied to the different voter authentication/access options. Still needs styling and wordsmithing. Clicking one will now create the election and direct the voter to the admin page where the user can edit the election and save in steps.

The election details section will be open automatically the first time you go to the page so its clear it needs to be filled out.

The race section now shows a list of races and a green add button. Adding or editing a race will open a pop up dialog window with the race form. Saving will push the update to the server. There is now an option to delete races as well.

For candidates, before we had a list of text fields for candidate names that users could edit with an additional text field for adding new candidates. Now, existing candidates are shown in a box with their name, edit and delete buttons. I felt like this made it feel more locked in maybe? Curious on feedback on that. Also added arrow keys to rearrange the candidates but still thinking of better ways of implementing that. 

Settings option works the same as above. I didn't include the edit voter authentication settings in this PR, felt it was getting to big already.

Some other notes:
I moved some of the related components into their own folders. I also broke up some components into parts so they could be reused easily. For example the election details form now has a form component, a hook for handling all the form logic, and an inline form and pop up dialog form that ingest those to use them in different ways. I did this so if we still wanted a set up wizard it could easily use those components.

I changed the candidates and race IDs to be UUIDs. I still need to test that that didn't break anything.

Some styling and responsivity still needs work.

https://youtu.be/8KxPmN02DW8


ArendPeter (approved):
Looks great! The UX on the new flow looks a lot more clear!

Here's a suggestion for a future PR: 
I feel like seeing all the templates is still a lot of information. Maybe we can divide them into 2 minimized accordion groups. One for public elections and one for private

PR #380, opened 2023-11-06
Admin page cleanup

mikefranze:
## Description
Disables some edit buttons after draft phase, requires election title and 1 race to enable finalize button, refactors the top component so we're not calling the same components over and over. 

Adds new function to duplicate election in draft phase instead of opening a duplicate election in old form component. 

image

image
image


ArendPeter (approved):
Looks good!

PR #381, opened 2023-11-12
Fix scores order

mikefranze:
## Description
How the approval and plurality scores were displayed was incorrect. The scores array is sorted by score while and candidates array is in its original order. Since the scores array includes the candidate indexes, this gets the index of the candidate and retrieves the candidate name from the candidate array.

Before: 
image

After:
image

## Related Issues

none


ArendPeter (approved):
Looks good!

PR #385, opened 2023-11-20
Random candidate order and voter instruction confirmations

### LINKED ISSUE #335: Stop Random Display Order of Candidates (on the ballot)

mikefranze:
## Description
Adds setting to randomize candidate orders on the ballots. Order was already randomized but this allows you to disable it. Randomizing order is enabled by default. 
Also adds setting for requiring voter confirm they've read ballot instructions. Hooks are in to enable/disable it, but keeping it locked in the settings form for now until we get the checkbox added to the ballot.

Settings page
image
With randomization disabled
image
With randomization enabled
image

## Related Issues

closes #335 
provides hook for #363

### DIFF
--- a/domain_model/ElectionSettings.ts
+++ b/domain_model/ElectionSettings.ts
@@ -23,4 +23,5 @@ export interface ElectionSettings {
     public_results?:	    boolean; //		allows public to view results
     time_zone?:           string; // Time zone for displaying election start/end times 
     random_candidate_order?: boolean; // Randomize order of candidates on the ballot
+    require_instruction_confirmation?: boolean // Require voter to confirm that they've read the instructions in order to vote
   }
--- a/domain_model/ElectionSettings.ts
+++ b/domain_model/ElectionSettings.ts
@@ -22,4 +22,5 @@ export interface ElectionSettings {
     ballot_updates?:	    boolean; //		allows voters to update their ballots before election ends
     public_results?:	    boolean; //		allows public to view results
     time_zone?:           string; // Time zone for displaying election start/end times 
+    random_candidate_order?: boolean; // Randomize order of candidates on the ballot
   }
--- a/frontend/src/components/Election/Election.tsx
+++ b/frontend/src/components/Election/Election.tsx
@@ -31,7 +31,7 @@ const TempWrapper = ({ authSession }: Props) => {
       <Grid xs={12} sm={8}>
         <Routes>
           <Route path='/' element={<ElectionHome authSession={authSession} electionData={data} fetchElection={refreshElection} />} />
-          <Route path='/vote' element={<VotePage election={data.election} fetchElection={refreshElection} />} />
+          <Route path='/vote' element={<VotePage />} />
           <Route path='/thanks' element={<Thanks election={data.election} />} />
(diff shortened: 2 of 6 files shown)


ArendPeter (approved):
Looks good! Thanks for adding this, the useElection context looks really handy

mikefranze:
> Looks good! Thanks for adding this, the useElection context looks really handy

Thanks! It helps simplify the interfaces a lot.

PR #388, opened 2023-11-22
Use quick poll data if it exists in CreateElection

### LINKED ISSUE #318: Improve transition from quick poll to advanced poll
[from Sara over slack]

It would be good to have an idea of what features basic mode includes and doesn't on the first create election page, or if you do fill out the first page and then click more settings can we have it autofill with what's already been filled out?

 * I'm not sure how to do this best, but maybe the buttons should be "create a quick STAR Voting Poll" and "Create Election".
 * Or .... maybe it's best to not separate quick polls out from other elections, but rather allow people to not get into the weeds unless they want to? If we went this route the homepage could have the basic poll screen like it does now, but then the only button would say "Create Election", That would take you to the "your election" type page, but it would have an new "Election Details" button or section that shows you a summary of the options selected (1 winner, star voting, open to all, etc.) and it would have a "More Options" button to allow you to customize the election more or less as you would with the "Explore More Settings" button from the homepage.
 * On the homepage we could add a section that has a bullet point feature list, so people looking at the tool will know off the top what options are on the table. Maybe that would be a chunk of the homepage, or maybe it could be a button that says "Explore Features"

mikefranze:
## Description
Adds generic types to useLocalStorage 
In Create Election page, will use quick poll title if it exists and any candidates that were entered. 

Quick poll page with title and candidates entered
image

Clicking explore more settings and selecting a template will bring you here
image

Race view
image

## Related Issues

closes #318


frontend/src/components/ElectionForm/QuickPoll.tsx
```diff
@@ -11,29 +11,22 @@ import { useLocalStorage } from '../../hooks/useLocalStorage';
...
-            })
-        if ((!newElection)) {
-            throw Error("Error submitting election");
-        }
-        localStorage.removeItem('Election')
-        navigate(`/Election/${newElection.election.election_id}`)
-    }
 
-    const QuickPollTemplate = {
+    const QuickPollTemplate: Election = {
```
ArendPeter, line 21: Looks like the quick poll reset every time?

I think this is usually the correct behavior, but for the following flow it would be nice to have it remember the previous info

1. User partially fills out the quick poll, but realizes they want to specify more things
2. User is not logged in, so they click "Login for more settings"
3. When they come back they should be able to continue where they left off 

I see 2 implementation options
1. The QuickPoll component checks if session_state is in the url, and only reads localStorage if that's true (kind of hacky, but I think it would work)
2. Clicking "login for more settings" directs the user to the admin page, and then the admin page will redirect to the login flow if the user isn't logged in. This way they'll be redirected to admin page with more settings after login

Option 2 requires fewer clicks from the user, but seeing the admin page might be confusing to them, not sure which is better

ArendPeter (commented):
Left some comments

frontend/src/components/ElectionForm/QuickPoll.tsx
mikefranze: QuickPoll doesn't get reset. This input to the useLocalStorage hook is just the default value if it doesn't exist in local storage already.

ArendPeter: Ok great! Then there are 2 other questions

1. What happens if someone partially fills out the quick poll, and then creates a new election using the nav bar? Will the election creation know to ignore the quick poll info?
2. (nit) Should we have an expiration setup so that people don't to see their partial quick poll days/weeks into the future?

ArendPeter: hmm, I feel like it should ignore it, but I don't feel too strongly

Everything else looks good so I'll approve. We can create tasks for those items if you think they're worth addressing in the future

PR #406, opened 2023-12-05
Make Elections header button center clickable

mjpauly:
## Description

Make Elections header button center clickable.


ArendPeter (approved):
Looks good!

PR #410, opened 2023-12-08
Add checkbox to ballot for reading instructions

### LINKED ISSUE #405: Implement ballot checkbox

ArendPeter:
## Description
 * Added checkbox to ballot
 * Blur instructions when box isn't checked
 * Changed bubbles to ovals
 * Created a BallotContext

https://github.com/Equal-Vote/star-server/assets/9289903/97398af9-67cf-423f-a669-6542878675bc

## Related Issues

fixes #405

### DIFF
--- a/frontend/src/components/Election/Voting/BallotPageSelector.js
+++ b/frontend/src/components/Election/Voting/BallotPageSelector.js
@@ -9,100 +9,46 @@ import StarPRBallotView from "./StarPRBallotView";
 import { DetailExpander } from '../../util';
 
 export default function BallotPageSelector({page, races, onUpdate}) {
-  var race, candidates; 
-  if(page.type == "ballot"){
-    race = races[page.race_index];
-    candidates = page.candidates;
-  }
+  let race = races[page.race_index];
+  let candidates = page.candidates;
   // TODO: it would be more scalable if we selected the class from a dictionary, but I'm not sure how to do that in react
   return (
     <>
-    {page.type == 'info' &&
-      <Box border={2} sx={{ mt: 5, ml: 0, mr: 0, width: '100%', padding: '10%', minHeight: '500px' }} className="ballot">
-        {page.voting_method == 'STAR' &&
-          <>
-          The next race uses STAR Voting, here's how it works
-          <img style={{maxWidth: '100%'}} src="/images/star_info_vertical_top.png"/>
-          <DetailExpander title='Learn More'>
-            <img style={{maxWidth: '100%'}} src="/images/star_info_vertical_bottom.png"/>
-            <hr/>
-            <iframe width="480" height="270" src="https://www.youtube.com/embed/3-mOeUXAkV0" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
-          </DetailExpander>
-          </>
-        }
-        {page.voting_method == 'STAR_PR' &&
-          <>
-          The next race uses Proportional STAR Voting, here's how it works
-          <img style={{maxWidth: '100%'}} src="/images/star_pr_info.png"/>
-          </>
-        }
-        {page.voting_method == 'Plurality' &&
-          <>
-          The next race uses Plurality Voting, just fill in 1 bubble
-          <hr/>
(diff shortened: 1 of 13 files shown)


frontend/src/components/Election/Voting/VotePage.tsx
```diff
@@ -45,35 +47,26 @@ const VotePage = () => {
     let pages = election.races.map((race, i) => {
       let candidates = race.candidates.map(c=> ({...c, score: null}))
       return {
-        type: "ballot",
+        instructionsRead: false,
```
mikefranze, line 50: I added a setting in #385 "require_instruction_confirmation", we should only show this when that is set. We also need to enable that in the settings page in this PR.

frontend/src/components/Election/Voting/VotePage.tsx
```diff
@@ -21,6 +21,8 @@ const CHECKED_BOX = "M 19 3 H 5 c -1.11 0 -2 0.9 -2 2 v 14 c 0 1.1 0.89 2 2 2 h
 //const UNCHECKED_BOX = "M 19 5 v 14 H 5 V 5 h 14 m 0 -2 H 5 c -1.1 0 -2 0.9 -2 2 v 14 c 0 1.1 0.9 2 2 2 h 14 c 1.1 0 2 -0.9 2 -2 V 5 c 0 -1.1 -0.9 -2 -2 -2 Z"
 const DOT_ICON = "M12 6c3.31 0 6 2.69 6 6s-2.69 6-6 6-6-2.69-6-6 2.69-6 6-6m0-2c-4.42 0-8 3.58-8 8s3.58 8 8 8 8-3.58 8-8-3.58-8-8-8z"
 
+export const BallotContext = createContext({})
```
mikefranze, line 24: Could you add types? Makes it easier to determine types in the components that use it.

frontend/src/components/Election/Voting/BallotPageSelector.js
```diff
@@ -8,101 +8,15 @@ import ApprovalBallotView from "./ApprovalBallotView.js";
...
-            onUpdate={onUpdate}
-            />
-        }
-      </>
-    }
+      {votingMethod == 'STAR' && <StarBallotView/>}
+      {votingMethod == 'STAR_PR' && <StarPRBallotView/>}
+      {votingMethod == 'Plurality' && <PluralityBallotView/>}
+      {(votingMethod == 'RankedRobin' || votingMethod == 'IRV') && <RankedBallotView/>}
+      {votingMethod == 'Approval' && <ApprovalBallotView/>}
```
mikefranze, line 19: Nice, looks much cleaner

mikefranze (changes requested):

frontend/src/components/Election/Voting/VotePage.tsx
ArendPeter: Sure, how deep should I show the nesting? Is it ok to only give the first level of keys, or do those need to be fully defined?

mikefranze: Just the top level keys, should be able to import the race and candidate types so should be simple. 

What type do the components currently see this as? An any type?

frontend/src/components/Election/Voting/VotePage.tsx
```diff
@@ -45,35 +57,26 @@ const VotePage = () => {
...
     return pages
   }
+
   const { id } = useParams();
   const [pages, setPages] = useState(makePages())
   const navigate = useNavigate();
   const [currentPage, setCurrentPage] = useState(0)
+
+  const setInstructionsRead = () => {
+    pages[currentPage].instructionsRead = true;
```
mikefranze, line 75: How does this work? pages should be immutable as a state

ArendPeter: Yeah, that operation a lone won't trigger a refresh so that's why I call setPages

It's a bit hacky, but the alternative is to have a separate useState where I represent instructionsRead as an array and I'd need to do the same trick anyway

frontend/src/components/Election/Voting/VotePage.tsx
ArendPeter: Done!

frontend/src/components/Election/Voting/VotePage.tsx
ArendPeter: Done! (we figured out the format over slack)

mikefranze (approved):
Looks good!

Issue #415, opened 2023-12-13
Make tie breaker scenarios more clear in results (for approval)


ArendPeter:

ArendPeter:
Approval is a 1.0 feature (possibly 2.0), will push back the milestone

PR #416, opened 2023-12-13
Add tie result for approval

ArendPeter:
## Description
This updates the approval results page to sometimes indicate that the election is tied instead of showing a winner

Here are the specific updates
* Add election setting for breaking ties randomly
* Adjust results title to show ties
* Refactor tabulator to avoid managing offsets on an array
* Refactor tabulator to always sort candidates and break ties randomly (the break ties randomly setting is only used by the frontend)
* Refactor tabulator with SingleWinnerApproval sub-method

Here's some follow up tasks
 * filter random tie break out of the detailed steps if random tiebreaker isn't enabled
 * add support for random tie breaker in the settings (it's currently locked to disabled)

image

## Related Issues

related #415

### DIFF
--- a/backend/src/Tabulators/AllocatedScore.ts
+++ b/backend/src/Tabulators/AllocatedScore.ts
@@ -48,6 +48,7 @@ export function AllocatedScore(candidates: string[], votes: ballot[], nWinners =
         other: [],
         roundResults: [],
         summaryData: summaryData,
+        tieBreakType: 'none',
     }
     var remainingCandidates = [...summaryData.candidates]
     // Run election rounds until there are no remaining candidates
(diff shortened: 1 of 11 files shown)


mikefranze:
Looks good, but I'm wondering if we should also include ties that occur in each round in the round_results. That should also be an array since you could have multiple ties in STAR.

ArendPeter:
> Looks good, but I'm wondering if we should also include ties that occur in each round in the round_results. That should also be an array since you could have multiple ties in STAR.

Agreed, those can be in round_results

How would you feel about renaming round_results to details or logs?

ArendPeter:
> Looks good, but I'm wondering if we should also include ties that occur in each round in the round_results. That should also be an array since you could have multiple ties in STAR.

I added the ties to the logs within the round results, but not as an additional field. I'd be open exposing it's own field in the future, but in general I think it's better to keep the output simpler and limit the fields to those that we expect the frontend to consume.

mikefranze (approved):
Looks good

PR #417, opened 2023-12-15
Add dynamic placeholder text when entering voter data

ArendPeter:
## Description

Add dynamic placeholder text to make the input format more clear

https://github.com/Equal-Vote/star-server/assets/9289903/1e5d67a7-44b0-4fa9-a679-20c95eec5535


mikefranze (approved):
Looks good, been meaning to add something like this for a while. Good idea with the snackbar for errors.

PR #418, opened 2023-12-16
Use paths to clean up domain_model references

ArendPeter:
## Description

This cleans up references to the domain_model

It's only applied to the frontend, backend is harder since there are multiple domain_model folders that are used there


mikefranze (approved):
Didn't know about paths, very useful!

PR #419, opened 2023-12-16
Get voter limit from shared config

ArendPeter:
## Description
Update create-react-app to let us use shared files

image

image

image

### DIFF
--- a/shared/SharedConfig.ts
+++ b/shared/SharedConfig.ts
@@ -1,6 +1,8 @@
 export const sharedConfig = {
     FREE_TIER_PRIVATE_VOTER_LIMIT: 2,
     ELECTION_VOTER_LIMIT_OVERRIDES: {
-        'ee948c52-f79e-4449-acb1-1296debc0884': 10
-    }
+    },
+    ACCOUNT_CUSTOM_REGISTRATION_OVERRIDES: [
+        '84724136-bb4d-41be-8202-3373a1a1c934', // Arend's Keycloak sub
+    ]
 };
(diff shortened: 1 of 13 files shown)


mikefranze:
Looks good, but I'd like to see that it works in the back end too in our deployment process before approving.

mikefranze:
Have you tried building the docker image? I think you'll need to copy the shared folder like the domain folder.

ArendPeter:
> Have you tried building the docker image? I think you'll need to copy the shared folder like the domain folder.

Done! I added the copied commands and tested building the image. It works

ArendPeter:
Rebased, it should be a clean merge now

shared/SharedConfig.ts
```diff
@@ -0,0 +1,9 @@
+export const sharedConfig = {
+    FREE_TIER_PRIVATE_VOTER_LIMIT: 2, // making really low at first so it's easy to test
```
mikefranze, line 2: Can we increase this to 20? Would like to be able to test with more voters in private elections. We can lower it locally for testing.

mikefranze: This is still only applied in the front end right? Where does the backend use this to reject the request?

ArendPeter: You're right I messed up the rebase and lost the most recent commits. Should be fixed now

(more specifically I did the rebase on my desktop, but the latest version of the branch was from my laptop, so I overwrite the laptop changes with the force push)

mikefranze (approved):

PR #420, opened 2023-12-16
Results types

mikefranze:
## Description
Moves the tabulator results types to the domain model folder and adds the types to the front end so the results page can be fully typed.
Adds some generic types that can be extended for specific voting method types.


domain_model/ITabulators.ts
```diff
@@ -0,0 +1,130 @@
...
+    pairwiseMatrix: pairwiseMatrix,
+}
+
+export interface allocatedScoreSummaryData extends starSummaryData {
+    splitPoints: number[],
+    spentAboves: number[],
+    weight_on_splits: number[],
+    weightedScoresByRound: number[][]
+}
+export interface approvalSummaryData extends genericSummaryData { }
```
ArendPeter, line 55: much cleaner :D

domain_model/ITabulators.ts
```diff
@@ -0,0 +1,130 @@
...
+    summaryData: pluralitySummaryData,
+}
+
+export interface rankedRobinResults extends genericResults {
+    summaryData: rankedRobinSummaryData,
+}
+
+export interface irvResults extends Omit<genericResults, 'roundResults'> {
+    summaryData: rankedRobinSummaryData,
+    logs: string[],
```
ArendPeter, line 104: Not for this PR, but in the future should we remove logs in favor of putting it in roundResults?

ArendPeter (approved):
Looks great!

domain_model/ITabulators.ts
mikefranze: Probably a good idea.

Issue #325, opened 2023-07-05
Add more sandbox validation


ArendPeter:
Adam detailed the issue here, https://docs.google.com/document/d/1hHoPLDHh7Y6-zOTuXtbMQ9584polrH8WKpa7fshg-ZY/edit

Issue #314, opened 2023-06-27
Sandbox page improvements


mikefranze:
- Sandbox page currently fills up the whole window, could use some paddding
- Need checks that number of candidates match number of scores given
- As mentioned in #132 we could add live updates when form inputs change
- Maybe add ability to have side by side windows to compare results?

mikefranze:
Hi @Morning1139Angel, we'd really appreciate your help! Let us know if you have any questions or need help setting up locally.

Issue #132, opened 2022-07-29
Add live updates to sandbox


ArendPeter:
Suggested by Sass during 7/28/2022 happy hour meeting

For sandbox mode it would be nice if the results were updating as the csv is updated. Currently users need to hit "Get Results" whenever they want the results to refresh

mikefranze:
This would be a good first React issue. Would involve using hooks to run functions when the data changes, checking that the data is valid(don't want to fetch if user is in the middle of typing and the input is not in the right format), and fetching results if valid.

PR #421, opened 2023-12-17
Sandbox updates

### LINKED ISSUE #367: incorrect result - B wins (correct result A wins)
incorrect result - B wins (correct result A wins)
    A,B,C,D
21: 5,4,3,2
12: 2,3,5,4
5:  3,2,4,5
12: 3,5,2,4

See details at: https://docs.google.com/document/d/1qNzGzW4gRdQ7_vtGJqBgWknmgfp8dANMc3tmfhh-B-0/edit

mikefranze:
## Description
Some updates to the sandbox to prevent errors that break the page.
Adds API hook for sandbox.
Adds useEffect to refetch data after inputs change.
Adds some input validation to check that the correct number of  votes are added per line and filters out spaces, empty strings, and NaNs. This could probably be done better with some regex but I need relearn some of that.
Sends back number of winners, voting method, and candidates with results, and only render results with data sent back. This prevents errors where user would change the number of winners or voting method that wouldn't work with the results currently loaded and crash the page.

## Related Issues
Fixes #367, #325, #314, #132

### DIFF
--- a/backend/src/Controllers/sandboxController.ts
+++ b/backend/src/Controllers/sandboxController.ts
@@ -23,7 +23,9 @@ const getSandboxResults = async (req: Request, res: Response, next: NextFunction
 
     res.json(
         {
-            results: results
+            results: results,
+            nWinners: num_winners,
+            candidates: candidateNames,
         }
     );
 }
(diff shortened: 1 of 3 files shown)


ArendPeter (approved):
Looks good!

PR #423, opened 2024-01-02
fix: remove file extensions from relative imports

JonBlauvelt:
## Description
Removes .js file extension from relative imports that were breaking dev server (nodemon) since it is looking at the files in src (.ts) and not in build dir (.js).

## Related Issues
None listed though this seems to be a recurring sticking point when folks are following the dev onboarding which suggests using `npm run dev`. It's not strictly necessary since folks can use `npm run build && npm run start`, but might be nice to be able to directly pickup backend changes during dev without restarting the server.


ArendPeter (approved):
Looks good!

Issue #425, opened 2024-01-02
Improve onboarding flow


ArendPeter:
Here's some improvements that came up when onboarding Jon

https://equal-vote.github.io/star-server/contributions/1_local_setup.html

- [x] Onboarding documentation still has some references to the script it was copied from (which was here: https://contribute.freecodecamp.org/#/how-to-setup-freecodecamp-locally )
- [x] Fix the start script (either update the documentation to show npm run start, or fix npm run dev)
- [x] Make it more clear that credentials are required
- [x] Add instructions for hosting the database locally (and add instructions as needed, but also create scripts where possible)
- [x] Add calendar links, and explain weekly vs monthly meetings: starvoting.us/calendar
- [x] Add instructions for copying frontend sample.env (or update the dev script to copy sample.env to .env when .env isn't present)

JonBlauvelt:
Created slack workflow for onbaording.

ArendPeter:
I've added myself

I'll take a look at the database instructions, I think that's the only task left on here

ArendPeter:
spin up docker image locally

ArendPeter:
Assigning to Mike since he has more context on the database stuff

PR #436, opened 2024-01-11
fix: update dev setup docs and default env vars

JonBlauvelt:
## Description
After completing onboarding, making some changes to doc to reflect current setup process.  I made some assumptions here:
* We don't need/want any legacy (pre-azure) env vars
* We don't need to get keycloak creds directly since we're using the shared dev creds from the doc
## Related issues
https://github.com/Equal-Vote/star-server/issues/425

### DIFF
--- a/docs/contributions/1_local_setup.md
+++ b/docs/contributions/1_local_setup.md
@@ -83,7 +83,7 @@ This is essential, as it allows you to work on your own copy of star.vote on Git
 
 ## Clone your fork from GitHub
 
-[Cloning](https://help.github.com/articles/cloning-a-repository/) is where you **download** a copy of a repository from a `remote` location that is either owned by you or by someone else. In your case, this remote location is your `fork` of freeCodeCamp's repository that should be available at `https://github.com/YOUR_USER_NAME/star-server`. (`YOUR_USER_NAME` would be replaced with your GitHub user name.)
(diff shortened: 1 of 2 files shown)


docs/contributions/1_local_setup.md
```diff
@@ -97,7 +97,7 @@ Run these commands on your local machine:
    git clone --depth=1 https://github.com/YOUR_USER_NAME/star-server.git
    ```
 
-This will download the entire freeCodeCamp repository to your projects directory.
```
ArendPeter, line 100: Good catch!

docs/contributions/1_local_setup.md
```diff
@@ -202,60 +186,26 @@ copy sample.env .env
...
 Launch backend 
 
 ```bash
 cd backend
+npm install
 npm run dev
 ```
 
-Launch frontend
+Launch frontend. If you want to just run the web client without setting up the server you can change the proxy under frontend/package.json to "proxy": "dev.star.vote".
```
ArendPeter, line 201: Were you able to get the proxy to work? 

I used to use this trick when we were on heroku, but I haven't had success with the dev.star.vote endpoint yet

ArendPeter (approved):
Looks good! Thanks for making these fixes

docs/contributions/1_local_setup.md
JonBlauvelt: I didn't try it - assumed it was still working 😬 
I was able to get it to look at other local port here.

PR #440, opened 2024-01-12
Enable irv and fix quick poll for demo

ArendPeter:

## Related Issues


ArendPeter:
Merging for demo

Issue #449, opened 2024-01-19
Handle confusing experience with ip locked elections


ArendPeter:
I was confused by an election where I appeared in the voter log 3 times. Also 2 of them were submitted and 1 of them was unsubmitted, despite this being a public election

This was because it was an ip-locked election, but there some actions items we should take on to make it less confusing

![Image](https://github.com/Equal-Vote/star-server/assets/9289903/a6f16e62-21f9-4141-9eda-1e104c9ed517)

## Problems

1. IP address shouldn't be visible to admins
2. Since this is a public election it shouldn't be possible to be added to the roll unless you vote. Currently it seems like I get added to the roll when I open the ballot?
3. For the admin experience, it's confusing seeing multiple votes from the same email address (even if they're from different IPs).  
4. The elections currently default to ip locked, this should default to false and be deliberately enabled
5. For the voter experience, it's confusing seeing "ballot submitted" for an election that I haven't voted in yet (for example if someone else voted on the same IP) 
6. Showing the email address in the Admin view of a public election is problematic. Someone could make a public election, and then farm their email addresses. Email addresses should only be visible if the list specified by the admin

## Solutions

1. Remove email & ip from public election tables. Email should only be shown if you're hosting a private election using an email list  (this also needs to be obscured on the API level so that people can't retrieve the information) (fixes problems 1, 3, 6) 
2.  Currently voters get added to the election roll as a side effect of getElection. So everyone who opens the homepage will automatically be "registered" to all  the featured elections. IMO voters in public elections shouldn't be added to the voter roll until they vote (fixes problem 2 )
3. Allow admins to specify the voter authentication after creating the election (ip, device cache, or email login) (fixes problem 4)

I'm not sure how to solve problem #5. Messaging "a ballot has already been submitted for this IP" would be the best user experience, but it encourages users to seek out different IPs. Maybe this is a good excuse to disable IP locking for beta

ArendPeter:
Here's the columns we decided on in most recent meeting

Standard Field
 - Voter ID
 - Create Time

Public Local Storage
 - voter id, vote status
Public IP Locked
 - voter id, vote status
Public Email Verification
 - voter id, vote status
Private Standard
 - voter id, (email), (email invite), has voted, state
Elections w/ Email
 Standard + Email
Elections w/ Email Invite
 Standard + Email Invite

ArendPeter:
Resolving in favor of subtasks

Issue #447, opened 2024-01-16
Add profile pic to navbar


ArendPeter:
This would be pulled from google if they used login with google

ArendPeter:
I followed some links

https://stackoverflow.com/questions/71372878/keycloak-i-cant-add-custom-user-attribute-into-the-token
https://keycloak.discourse.group/t/how-to-get-google-profile-picture/16820

but I couldn't get it to work. Also one of the comments says user attribute mappers is a bad approach since it's limited to 255 characters, and image urls will often be longer than that


There are extensions that add this feature to keycloak, but we can leave that for later
https://github.com/thomasdarimont/keycloak-avatar-minio-extension

ArendPeter:
This might not be worth the effort at all, but for now I've moved it to milestone 3.0 for evaluation later

PR #450, opened 2024-01-19
Add view account to navbar and separate invitations and previous elections pages 

### LINKED ISSUE #446: Add "Your Account" link to navbar
This will take people to keycloak so that they can configure their profile

ArendPeter:
## Description

* Update getElections to seperate unsubmitted and submitted votes
* Add "your account" button that takes you to the keycloak account console

image

## Not covered

* Avatars: I looked into this but hit some complications. It's probably not worth it pre-beta (more details #447)
* Voter Roll Logic: Current voters get added to rolls as a side effect of getElection. As a result your ip get's readded to the featured elections every time your on a new device. This causes the unsubmitted ballot and submitted ballot views to have identical entries even though the queries are correct (more details #449)

## Related Issues

fixes #446


mikefranze (approved):
Looks good!

PR #455, opened 2024-01-27
Documenting starting up DB with docker

mikefranze:
## Description
Adds some documentation on how to spin up postgres database and keycloak server locally.
For a follow up task, it's possible to export a keycloak configuration as a file, we should export a version with the main setup steps already done to minimize setup time.

I have been having issues getting the local database to connect to the local server on my desktop, but seems to work on my laptop following the same steps. So test it out and let me know if it works for you.

## Related Issues
#425


ArendPeter (approved):

Issue #451, opened 2024-01-19
Add new columns to voter and ballot databases


ArendPeter:
Add the following columns to the election database
 - [x] claim_key_hash
 - [x] is_public (or is_unlisted? or visibility? visiblility could be an enum including private)
 - [x] create date

Votes
 - [x] ip_hash (replacing the ip column)

Ballot DB
 - check if there are more columns to add here

PR #456, opened 2024-01-28
Election creation dates, IP hashes, migration scripts

mikefranze:
## Description
Built on top of #455 

Updates election DB to include creation date, public election flag, and claim key. Existing elections in DB at time of migration will be given creation time based on current time.

Changes IP addresses in voter roll and ballots to IP hashes instead, though I don't think we're currently saving that for ballots.

Adds scripts to migrate the database up and down to help with development. 

Adds a NewElection type for the forms in the frontend that just removes the election id and creation date since they're assigned in the server. 

## Related Issues
#451


backend/src/Controllers/controllerUtils.ts
```diff
@@ -38,4 +38,8 @@ export function expectPermission(roles:roles[],permission:permission):any {
         if (!roles.some( (role) => permission.includes(role))){
             throw new Unauthorized("Does not have permission")
       }
+}
+
+export function hashString(inputString: string) {
+    return createHash('sha1').update(inputString).digest('hex')
```
ArendPeter, line 44: sha256 might be a better option

sounds like sha1 is still pretty secure overall, but if we're committing to one I figure we may as well go with the stronger option

https://www.keycdn.com/support/sha1-vs-sha256#:~:text=SHA1%20is%20no%20longer%20considered,should%20also%20switch%20to%20SHA256.

https://crypto.stackexchange.com/questions/48289/how-secure-is-sha1-what-are-the-chances-of-a-real-exploit

backend/src/Controllers/voterRollUtils.ts
```diff
@@ -46,12 +47,12 @@ export async function getOrCreateElectionRoll(req: IRequest, election: Election,
             election_id: String(election.election_id),
             email: req.user?.email ? req.user.email : undefined,
             voter_id: new_voter_id,
-            ip_address: req.ip,
+            ip_hash: hashString(req.ip),
```
ArendPeter, line 50: nit

Could we use the ip_hash variable here?

ArendPeter (approved):
I left some comments, it it looks great! Those migrator functions look useful

backend/src/Controllers/voterRollUtils.ts
mikefranze: Updated to reuse the variable, but changed the logic some to preserve the behavior. I'm not sure if this is what we want, but currently when we submit a new voter roll there we always include the IP address even if it isn't required for authentication. The ip_hash variable was being set to null if it isn't required so that the PG query knows to ignore it. I updated the ip_hash to be its own variable and the one sent to the the PG querry to require_ip_hash

ArendPeter (commented):
Looks good!

Issue #441, opened 2024-01-16
Add support for deleting an election (stronger version of archive)


ArendPeter:

ArendPeter:
Moving to 2.0, there's a larger story there for improving election state management

masiarek:
"Head" field: https://docs.google.com/document/d/1vJ65Wh9d-iKakIa8J4jzWHtCR02MchfvEAfL-wsz7Ng/edit?tab=t.0

PR #457, opened 2024-01-30
Database version histories

### LINKED ISSUE #453: Add versioning to databases
We're updating the elections database and voter roll database to support versioning. This will be done by making the primary key a union between the version and the id

We'll also need to update all the corresponding queries to sort for the latest version

mikefranze:
## Description
Updates database tables to allow multiple versions of each object. This was done by adding an "update_date" column and adding it to the primary key. Another column "head" was added to indicate which version is the latest for faster queries. I did some tests ordering the data by update_date and grabbing the latest, which performed ok but when grabbing multiple elections, ballots, or rolls it would require subqueries in a loop which slowed down the queries quite a bit. Using a head column and indexing it (just for the elections table) improved the performance by about 15x. 

The head is kept up to date by using transactions on all update queries. Transactions allow you to make multiple queries in one request and guarantees that they all occur or none of them do. So these transactions unflag the current head and push the new head. This is pretty simple in kysely.

In one test I went through the election creation flow and then retrieved the head and update times for all rows in the db with the election id
```
[
  { head: false, update_date: '1706628690445' },
  { head: false, update_date: '1706628795105' },
  { head: false, update_date: '1706628812315' },
  { head: false, update_date: '1706628965036' },
  { head: true, update_date: '1706628965691' }
]
```
Other updates:

- Updates Kysely to latest version to allow updating primary keys. This included updating queries that had been deprecated.
- Adds a new "Delete All Election Data" transaction query that will delete all rows in all tables with a given election id. The isn't hooked up anywhere yet but would be useful later.
- Adds a NewBallot type that doesn't require fields assigned by the server
- Adds a database sandbox script for making queries and doing tests on the local database. Not really organized in any way.

## Related Issues
closes #453 
starts on #441


backend/src/Models/Elections.ts
```diff
@@ -72,15 +85,16 @@ export default class ElectionsDB {
...
+
         if (id !== '' || email !== '') {
-            querry = querry.where(({ or, cmpr }) =>
-                or([
-                    cmpr('owner_id', '=', id),
-                    cmpr(sql`admin_ids::jsonb`, '?', email),
-                    cmpr(sql`audit_ids::jsonb`, '?', email),
-                    cmpr(sql`credential_ids::jsonb`, '?', email)
-                ]))
+            querry = querry.where(({ eb }) =>
```
ArendPeter, line 92: typo? s/querry/query

domain_model/Election.ts
```diff
@@ -20,13 +20,15 @@ export interface Election {
     settings:       ElectionSettings;
     auth_key?:      string;
     claim_key_hash?: string;
-    is_public?:      Boolean;
-    create_date:    Date | string;
```
ArendPeter, line 24: Does Date | string mean there's multiple ways this can be stored in the DB? 

I would be cleaner if the format could be consistent, but I have no idea how much work that would be so feel free to ignore

ArendPeter (approved):

Also good work doing the research on the performance

backend/src/Models/Elections.ts
mikefranze: Ah, I kept making that mistake. I'll do a search to make sure I didn't do that anywhere else.

domain_model/Election.ts
mikefranze: The reason for it is that dates get converted to string when converted to JSON. I think they're only being stored as date objects in the front end to interface with the date logic better.

backend/src/Models/Elections.ts
mikefranze: Fixed

PR #463, opened 2024-02-06
Fixing bug in start up script

mikefranze:
## Description
Moved migration script, didn't fix the top level script start script that calls it.


ArendPeter (approved):
Good find!

PR #466, opened 2024-02-16
Tweak feature flags & Super quick tie support

ArendPeter:
## Description

Tweak feature flags to disable public elections page and add super quick tie support


ArendPeter:
Merging since I want these features for a demo

PR #467, opened 2024-02-17
Fix end time

ArendPeter:
## Description

Ends at was showing the wrong time


ArendPeter:
Merging since it's necessary for demo

PR #468, opened 2024-02-18
Add max width for results yaxis

ArendPeter:
## Description

I noticed the mobile view on some elections cut off the names, so I set a max on the y axis width

image


ArendPeter:
Merging since this is relevant for a demo

PR #469, opened 2024-02-18
Admin columns

### LINKED ISSUE #459: Adjust which voter columns are displayed to the admin
see parent issue: https://github.com/Equal-Vote/star-server/issues/449

ArendPeter:
## Description

* Combined EnhancedElection into EnhancedTable
* Moved roll columns into enhanced table
* Disable voter roll view for public elections * (both frontend and backend)
* Also cleaned up election invitations table

\* = after scraping sensitive fields, there was only voter id and voter state left, and no way to identify the voters. So I didn't feel it was useful for public elections and it was easier to just disable the view

## Related Issues

fixes #459


mikefranze (approved):
Looks good!

PR #481, opened 2024-02-26
Add new id scheme for the election id

ArendPeter:
## Description

Here's what's new

 * Election IDs use a 6 digit random string instead of a uuid
 * The election routes are now at the root
 * Some extra logic future proofing slugs
 * New election ids will be tested against the new DB and classic star.vote 
 * If a slug matches a classic.star.vote slug we'll redirect
 
 

https://github.com/Equal-Vote/star-server/assets/9289903/2ba72252-6b1c-4e4a-8061-0ca43f8ad35c


ArendPeter:
I'm going to go ahead and merge it, we had an informal review during a meeting a few weeks ago, and I'd like to have this change in for the Eugene demo election

PR #510, opened 2024-03-06
Hotfix how end time is displayed post election

ArendPeter:
## Description

Hotfixing a bug where the wrong date is displayed post-election


ArendPeter:
Merging since this is relevant for a live demo

PR #516, opened 2024-03-09
Remove "My invitations" - https://github.com/Equal-Vote/star-server/issues/490

### LINKED ISSUE #490: My Election invitations - menu item - remove
Typically folks will not register at our website - hence I suggest that we remove this menu item.

Voters will simply use their Notifications (email, SMS, etc) to cast a vote.

image

We obviously need a very mature and feature rich “Dashboard” with statuses related to “sent emails” monitoring - what is in the queue, what bounced back, etc - but this is a different concept and approach.

In case we decide to keep - we need to fix the bug:
- video 1 - https://youtu.be/S_GYVQ55wBA
- video 2 - https://youtu.be/zxOXvQektso

masiarek:
## Description
Fixes #490

### DIFF
--- a/backend/src/Controllers/auth.controllers.ts
+++ b/backend/src/Controllers/auth.controllers.ts
@@ -13,7 +13,6 @@ const getUser = (req: any, res: any, next: any) => {
   Logger.info(req, `${className}.getUser`);
   const user = accountService.extractUserFromRequest(req);
   if (user){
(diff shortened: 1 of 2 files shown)


ArendPeter:
Looks good! We agreed to remove this for now, and we can always add it back if we want it later

ArendPeter (approved):
Approved

PR #533, opened 2024-03-17
Make results prettier

ArendPeter:
## Description
 - Replace "No Preference" with "Equal Preference" every in the front end
 - Count Equal Preference distribution support in the tabulator
 - Add view results button when opening the ballot
 - Add landscape support
 - Setup shared Widget / WidgetContainer system
 - Add landscape support
 - Adjust subcomponents layout on results
 - Set limit on how many candidates are shown in top level results

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5067,6 +5067,14 @@
         "node": ">=10.17.0"
       }
     },
+    "node_modules/i": {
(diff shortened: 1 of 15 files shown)


ArendPeter:
Merging since we're going to use the new results page for a demo

PR #535, opened 2024-03-18
chore: upgrade dependencies and node version

JonBlauvelt:
## Description
chore: upgrade dependencies and node version (including migration from aws sdk v2 to v3)

Screenshot 2024-03-18 at 1 39 57 PM
Screenshot 2024-03-18 at 1 30 13 PM


Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN npm run build -ws
 # Production Stage #
 ####################
 
-FROM node:16.19.1-bullseye-slim
+FROM node:21.7.1-bullseye-slim
```
ArendPeter, line 50: I noticed fetch worked on my local but not when deploying. The old node version might have been the culprit :o 

https://stackoverflow.com/questions/48433783/referenceerror-fetch-is-not-defined

packages/backend/package.json
```diff
@@ -17,6 +17,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.536.0",
```
ArendPeter, line 20: nice, does importing s3 specifically mean our node_modules will be smaller?

packages/backend/src/Controllers/uploadImageController.ts
```diff
@@ -1,9 +1,11 @@
 import { InternalServerError } from "@curveball/http-errors";
 import { randomUUID } from "crypto";
 import { Request, Response, NextFunction } from 'express';
-const S3 = require("aws-sdk/clients/s3");
-const ID = process.env.S3_ID;
-const SECRET = process.env.S3_SECRET;
+import  { Upload, Progress } from "@aws-sdk/lib-storage";
+import { S3 } from "@aws-sdk/client-s3";
+
+const ID = process.env.S3_ID!;
```
ArendPeter, line 7: Never knew about the ! operator, looks handy

ArendPeter (approved):
Looks great, thanks for making this change

PR #538, opened 2024-03-20
chore: add npm org prefix

JonBlauvelt:
## Description
chore: add npm org prefix to npm packages names
update docs and imports accordingly


ArendPeter (approved):
Looks good!

PR #540, opened 2024-03-24
New election creation

ArendPeter:
## Description

Dialog Feature
 - Convert election templates page to a 4 step dialog flow

Text Management
 - Setup i18n library to centralize text
 - Defined text in YAML

Misc
 - Adjusted some wording on the home page
 - Cleaned up an ancient login page
 - Added a reusable ToolTip component

https://github.com/Equal-Vote/star-server/assets/9289903/bccfbae3-7ed7-4435-9231-aa78caa038dc

image


mikefranze:
Looks great! Only comment is on mobile view. I think it would look better on smaller screens to use the full width. Also, would the user have to scroll on mobile to see the cancel button or does it stick to the bottom?

ArendPeter:
> Looks great! Only comment is on mobile view. I think it would look better on smaller screens to use the full width. 

Looks like it was already at fullWidth, but Paper's margin / width logic was limiting it. I pushed a fix and updated the screenshot in the description
 
> Also, would the user have to scroll on mobile to see the cancel button or does it stick to the bottom?
Maybe on the final stepper, but not the other ones (here's a screenshot pre-fullWidth fix)

image

packages/frontend/src/components/ElectionForm/CreateElectionDialog.tsx
```diff
@@ -0,0 +1,351 @@
...
+        onClose()
+    }
+
+    return <Dialog
+        open={createElectionContext.open}
+        onClose={onClose}
+        scroll={'paper'}
+        sx={{
+            maxHeight: '90%',
+            '@media(max-width: 500px)': {
```
ArendPeter, line 236: I know mui has better syntax for this, but I felt like media queries would be cleaner syntax here. I let's me keep all the changes in a single block, and I don't have to be worried about accidentally overriding any parent properties from Dialog or Paper

mikefranze (approved):

Issue #461, opened 2024-02-02
Allow admins to specify voter authentication on public elections (currently it defaults to ip locked)


ArendPeter:
see parent issue: #449

ArendPeter:
![image](https://github.com/Equal-Vote/star-server/assets/9289903/2856ab80-2dcc-4617-be26-33dffd06e780)
Here's an ugly mock up of what I'm visualizing. This will only be shown for elections that don't rely on a voter list (i.e. those that answered no on the 3rd question of the election creation flow). The settings that need to be modified already exist at election.settings.voter_authentication

evanstucker-hates-2fa:
I think we should definitely default this to NOT be IP locked. I was at a conference recently, and I created a poll. Many of the voters were unable to cast their ballots because we were all on shared wifi...

mikefranze:
Going to tackle this this weekend.

ArendPeter:
In addition the updating the admin view, I wonder if we should have a pop-up for ip vs device id when people are making quick polls?

I'm nervous about having either as a default so I think a quick prompt makes sense

mikefranze:
If the user is logged in, should we use the user id instead of device id? I'm thinking no.

mikefranze:
Actually it looks like currently the auth service code doesn't provide the user temp id if they're logged in, so it'd have to be the user id.

PR #598, opened 2024-07-13
Using client temp ID/user ID in open elections

mikefranze:
## Description
Adds the functionality to use the client temp_id or user ID in open elections instead of IP address and set it as default for quick polls. 
This involved adding some logic because open elections didn't look at voter IDs which led to some bugs. Now, when open election access and voter id authentication are both enabled, the server will look for user.sub instead of the voter id cookie. If the user is logged in user.sub is the keycloak user ID, otherwise it is the client temp ID.

PR for the form to switch between IP and ID coming later

Side note, all the logic involving voter access and voter authentication is getting convoluted and bug prone and I'm thinking of the best way to someday refactor it. Suggestions welcome.

https://github.com/user-attachments/assets/a34f437a-03c9-40e0-95a8-874aa5f18d2b

## Related Issues
Starts #461


ArendPeter (approved):
Looks good! 

I agree this could use some refactoring later. I think part of the confusion is because voter_id means different things for closed vs open elections. Maybe voter_id can be the term for the closed elections and then for open elections we'll call it a user_id?

mikefranze:
> Looks good!
> 
> I agree this could use some refactoring later. I think part of the confusion is because voter_id means different things for closed vs open elections. Maybe voter_id can be the term for the closed elections and then for open elections we'll call it a user_id?
> 
> ```
> const user_id = req.user?.sub ?? temp_id
> ```

I was thinking something like that or just not reusing any terms so there's no confusion. So the database would have
voter_id
device_id
user_id/email
ip_hash

PR #604, opened 2024-07-20
Adding race titles above candidates in ballot table

### LINKED ISSUE #562: Update ballot table for multi round elections
Currently admins can view the ballots for multi race elections, but you can't tell which columns belong to which race

Instead we have have the columns follow the format `<race name> - <candidate name>`

## Table
image

## Individual Ballot
image

mikefranze:
## Description
Putting the race titles in the candidate cells made the table too wide. This adds a row above candidates with the race titles, races span multiple columns. The default MUI tables don't have cell borders so I added borders to the races to make it clear, but I don't really like how it looks. Maybe ok for now.

image

image

## Related Issues

closes #562

### DIFF
--- a/packages/frontend/src/components/Election/Admin/ViewBallots.tsx
+++ b/packages/frontend/src/components/Election/Admin/ViewBallots.tsx
@@ -71,7 +71,7 @@ const ViewBallots = () => {
                                 <TableRow >
                                     <TableCell colSpan={3}></TableCell>
                                     {election.races.map(race => (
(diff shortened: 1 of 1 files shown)


ArendPeter (approved):
Awesome, looks good!

Here's an from the census website for wide tables. 

https://data.census.gov/table?g=010XX00US$0400000&y=2010

Perhaps we can add that as an issue for a future milestone, this solution looks great for now

Issue #591, opened 2024-07-10
Move tabulation log to en.yaml 


ArendPeter:
The backend will generate the logs as a series of keys and variables, and then the front end will plug those into en.yaml

PR #605, opened 2024-07-20
Refactor STAR Tabulator (sorting approach)

### LINKED ISSUE #424: Add Tie Support to the STAR Voting Tabulator
This tasks will be very similar to the Approval Tie support here: https://github.com/Equal-Vote/star-server/pull/416/files

Details
* Update frontend to display "tied" or "tied (x won after tie breaker)" depending on if break ties randomly was set in the election settings
* Update tabulator to always break ties randomly, and return the tie breaker type (the random tie breaker setting only affects how it's displayed on the frontend)
* For STAR there's a couple of tie breaker types which are detailed here: starvoting.us/ties . Ties in the runoff round are broken via the score, then the number of 5-stars, and then finally a random tie breaker. Only 5-star and random tie breakers should be displayed as a tie in the frontend

ArendPeter:
## Description
Major rework of star tabulator
 - Apply tie breaking logic using a sort function on the candidates
 - Update the tabulation logging to use keys from en.yaml
 - Update frontend to work with new logging logic

## Related Issues

fixes #424 and #591


ArendPeter:
Mike found an edge case, I added a failing test to ensure that it's covered

Here's our conversation

```
**Mike**: Ok, I've read through it and overall it seems good. My main concern is if the sort logic will result in the same answer as the old code. For example, if you have 4 candidates with tied scores for the second finalist position, it will do the sorting correctly. But then next round you have 3 candidates tied for the two finalist positions. Will the sorting work correctly then? Because you've removed one candidate the pairwise comparison totals could change the answer. I don't have a good grasp of how the sort function works to know the answer. To put another way, with four candidates you might have
A = B
A > C
A > D
B > C
C > D
D > B
The pairwise losses will be
A:0
B:1
C:2
D:2
So sorting A is easy, but if you remove A from the list the pairwise losses become
B:1
C:1
D:1
And sorting B is now more difficult
Second, you're going through the steps a second time to recreate the logs but those steps don't actually impact the end result which makes me nervous. If there's a bug the results and logs would be out of sync. I think it might be better to keep the tiebreaker logic and logs in the same place.

**Arend**: Currently it only computes the pairwise comparisons once, and then continues to use the same values when doing the logging. So values would still be 0, 1, 2, 2 and then after advancing A it would be 1, 2, 2. Does that sounds like the correct behavior or should the pairwise comparisons be recomputed after a candidate is advanced?

**Mike**: Pairwise would have to be recomputed whenever a candidate is removed from the pool. So when you advance a candidate or remove one. This is the original flowchart: https://docs.google.com/drawings/d/1L-tIDY6-8OeKGzW0w1eh2EUjlHbarEcsfvhsegMT8qg/edit?usp=sharing

**Arend**: After reviewing the tie breaker protocol more thoroughly, I'm going to make a smaller PR that avoids the drastic refactor (good work finding that edge case!)
That said, I'd be curious to reopen the tie breaker conversation in the future. In your example, I think there's strong case to be made that B is best candidate when selecting between B, C, D. With our current tie breaker protocol B could lose if it loses the 5-star tie breaker but the fact that it has a better head-to head performance against A should be a stronger indicator of B's majority mandate.
Making that change would simplify the algorithm while also making it more accurate IMO, although granted I wasn't super involved with the tiebreaker conversation, and I'm also biased since I just made that PR :sweat_smile:
```

PR #612, opened 2024-07-23
Center the header

ArendPeter:
## Description
The logo is now in the center of the screen (previously it was centered between the left/right elements)

image


mikefranze:
Does this center the icon in desktop too? Currently its left aligned.

ArendPeter:
ah, yes it does. I'll update the PR

mikefranze (approved):
Looks good, but lets test it on mobile once it goes live

PR #619, opened 2024-07-25
Allow for voting during draft phase

### LINKED ISSUE #608: Allow for voting during draft mode & add warning on the ballots that the votes will be reset

### LINKED ISSUE #618: Choosing "Poll" -  change title from "election" into "poll"
I tried to create a new Poll, and it seems like the Title doesn't rely on my choice:
image

ArendPeter:
## Description
- Voting is allowed during draft phase
- There's a warning indicating that all ballots are test ballots
- Ballots are cleared when races are edited, or if the election is finalized
- Warning message is also applied to email receipt

video walkthrough (I accidentally recorded during my wife's vocal warm ups lol)

https://github.com/user-attachments/assets/c674bf38-1c5d-45f0-bbfd-eee4500eeb9c

## Related Issues

fixes #618 
fixes #608


packages/frontend/src/components/Election/Admin/AdminHome.tsx
```diff
@@ -470,6 +471,7 @@ const AdminHome = () => {
             })
         if (!confirmed) return
         try {
+            await deleteAllBallots()
```
mikefranze, line 474: Deleting ballots during finalization should be done in the back end as a part of the finalize election controller otherwise this would be easy to bypass.

mikefranze (changes requested):

packages/frontend/src/components/Election/Admin/AdminHome.tsx
ArendPeter: Bypassing shouldn't be a problem since the backend has a check to only allow the operation if the election is in draft. 

Moving deleteAllBallots to the backend for finalize would be pretty easy, but the deleteAllBallot calls for updating races would be a little more awkward. Since updating a race applies an update to the whole election, the backend would have to check which part of the structure was updated in order to determine if the ballots need to be deleted.

mikefranze: My concern is that someone could finalize the election without clearing the ballots.

mikefranze: For dealing with updating the race, maybe we could have a parameter in the update election request that indicates what's being changed.

ArendPeter: Updated the PR

During finalize the deleteAllBallots is called from the backend, but for adjusting races I left that in the frontend.

I figure as long as they're guaranteed to reset during finalize, it's not a problem if someone bypasses it for adding races. I'm sure we'll come up with a more general solution later though.

mikefranze (approved):

PR #620, opened 2024-07-26
Dynamically update front page stats based on socket.io

### LINKED ISSUE #593: Convert front page election stats to websocket
- only if polling becomes an issue
 - it might not work with a load balancer
 - socket.io

ArendPeter:
## Related Issues

fixes #593

### DIFF
--- a/packages/frontend/src/components/Election/Voting/VotePage.tsx
+++ b/packages/frontend/src/components/Election/Voting/VotePage.tsx
@@ -18,6 +18,7 @@ import useFeatureFlags from "../../FeatureFlagContextProvider";
 import { Candidate } from "@equal-vote/star-vote-shared/domain_model/Candidate";
 import { Race } from "@equal-vote/star-vote-shared/domain_model/Race";
 import { useSubstitutedTranslation } from "~/components/util";
+import { io } from "socket.io-client";
 
 
 // I'm using the icon codes instead of an import because there was padding I couldn't get rid of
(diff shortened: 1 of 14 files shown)


packages/frontend/src/components/Election/Voting/VotePage.tsx
```diff
@@ -97,6 +102,7 @@ const VotePage = () => {
     setPages(newPages)
   }
   const submit = async () => {
+    socket.emit('new_vote');
```
mikefranze, line 105: I think the triggering of this event should be done from the backend at the end of the post request. If you export io from the socket handler you can emit directly from it anywhere in the server.

For example, I took one of their examples of a [chatroom ](https://codesandbox.io/p/devbox/github/socketio/chat-example/tree/cjs/step5?file=%2Findex.js%3A23%2C1) and appended this at the bottom of the server
```
setInterval(function () {
  io.emit('chat message', 'hello');
}, 1000);
```
And it sends a message to the client every second without the client having to trigger the event.

ArendPeter: Updated!

I wish I could find a way to avoid typing io as "Server | null", but let me know if you have ideas.

I also initially tried having a ``emit('new_vote')`` in the castVoteController and a separate ``on('new_vote')`` in the sockethandler, but that didn't work.

ArendPeter:
I'm working on the build failure...

packages/frontend/src/components/Election/Voting/VotePage.tsx
mikefranze: the null typing is probably due to the order things are initialized. The express app and all the routes/controllers are created first in "app" and then that is used to create io. So at the time the castVoteController imports it io doesn't exist yet.

mikefranze (approved):
Looks good

PR #628, opened 2024-07-28
Adding "Last Updated" to election tables and fixing table date bugs

mikefranze:
## Description
Adds the "update_date" election field and adds it to the election tables and makes it the default sort method. 
Adds local time zones to table header.
Fixes a bug with date formatting. It was trying to pass the entire election object into the formatter as the time zone which was making all dates "Invalid Datetime".
Temporarily fixes bug with date sorting. The table sorting function sorts based on already formatted data, so dates were sorting by their string version which led to many issues. Now the comparator function is passed a isDate parameter that tells the comparator to convert back to dates and then to integers. I hate this solution and myself for implementing it, but it works until I fix the table to sort and then format data. 

image


ArendPeter (approved):
Looks good!

PR #629, opened 2024-07-28
Fixing links in footer, disabled random candidate order as default

### LINKED ISSUE #610: Donations link is not working
https://dev.star.vote/Election/68m34r  - the “donations” link is not working

### LINKED ISSUE #621: Change the default from Random to Non-random - Candidates ordering on the ballot
Candidates ordering on the ballot - Change the default from Random to Non-random - keep the order as entered by Election Administrator.
Typically Election admin is drawing lots to establish the sequence of Candidates.
Ask: change the default from Random to Non-random 

Both Miroslav and I - we had the same initial reaction to the shuffled nature of Candidates on the Ballot.
If this is non-intuitive - we need to pay attention...

Yes, we should keep this shuffling option in software - but it should be default - OFF.

image

mikefranze:
## Description
Fixes links in footer. When using links in en.yaml, the url has to include "http://www.", otherwise it sees it as page on the current site. 

Another note, it doesn't look like its possible to insert a target="_blank" in the yaml file to open the page in a new tab. We should figure out a workaround at some point.

Edit: 
Turning off random candidate order by default

## Related Issues

closes #610
closes #621

### DIFF
--- a/packages/frontend/src/components/ElectionForm/CreateElectionDialog.tsx
+++ b/packages/frontend/src/components/ElectionForm/CreateElectionDialog.tsx
@@ -64,7 +64,7 @@ const defaultElection: NewElection = {
         ballot_updates: false,
         public_results: true,
         time_zone: DateTime.now().zone.name,
-        random_candidate_order: true,
+        random_candidate_order: false,
         require_instruction_confirmation: true,
         invitation: undefined,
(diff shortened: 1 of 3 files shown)


ArendPeter (approved):
Looks good!

PR #633, opened 2024-07-28
Email invitation settings fix

### LINKED ISSUE #631: Missing invite email to participate in the election
I created an election with emails - https://dev.star.vote/bfpr76

I did not receive any email invite to any of the emails:
image

mikefranze:
## Description
The invitation setting was getting set to true instead of 'email'. The reason its a string instead of Boolean is that it leaves open the potential for phone or mail invitations. 

## Related Issues

closes #631


ArendPeter (approved):
Thanks for the fix!

PR #641, opened 2024-08-07
Small Sandbox fixes

mikefranze:
## Description
When we added the election hook in the results page it broke the sandbox page because it doesn't provide an election context. Now we check if the election is defined when used it results. This was done it another place before as well.

Also moves the results in the sandbox to the side so it can be seen on one screen

image

image

## Related Issues
none


ArendPeter (approved):
Looks good!

PR #642, opened 2024-08-13
Move feature flags to sharedConfig

ArendPeter:
## Description

Approval & Ranked Robin weren't being enabled in production because the Dockerfile feature flag variables weren't applying in production. 

That's because react bakes all the "REACT_APP" variables into the code when it's compiled to javascript. The feature flags weren't working since they were referenced as ``process.env[`FF_${flag_name}`]``. 

To fix this I'm moving the feature flags to the sharedConfig


mikefranze (approved):
Looks good

PR #647, opened 2024-08-16
added csamuele as contributer

csamuele:

## Related Issues


ArendPeter (approved):
Looks good!

PR #648, opened 2024-08-16
Forced race title to be unique and fixed bug where races were not being validated for the first two

### LINKED ISSUE #503: Race name not unique (low priority)
race name not unique - are we ok with this?

Low priority ...

We can have two races with the same title:

https://dev.star.vote/Election/6774012d-d161-49b1-9552-bf965600c681/admin
image

csamuele:
image
image

## Related Issues
``fixes #503``

### DIFF
--- a/packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
+++ b/packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
@@ -50,7 +50,6 @@ export const useEditRace = (race, race_index) => {
     const validatePage = () => {
         let isValid = true
         let newErrors: any = {}
-        if (election.races.length > 1) {
             if (!editedRace.title) {
                 newErrors.raceTitle = 'Race title required';
                 isValid = false;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
```diff
@@ -50,7 +50,6 @@ export const useEditRace = (race, race_index) => {
     const validatePage = () => {
         let isValid = true
         let newErrors: any = {}
-        if (election.races.length > 1) {
```
ArendPeter, line 53: Me and Sam discussed removing this line. I believe this was a hold over from earlier UI iterations, and we opted to remove it since it was weakening the validation on the initial races that were created

packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
```diff
@@ -63,7 +62,18 @@ export const useEditRace = (race, race_index) => {
                 newErrors.raceDescription = 'Race title must be less than 1000 characters';
                 isValid = false;
             }
-        }
+            if (election.races.some(race => {
```
ArendPeter, line 65: Neat, I don't think I've used some() before but it looks handy

ArendPeter (approved):
Looks great!

packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
csamuele: I love all the array methods!

Issue #476, opened 2024-02-25
Usability issue - add candidates - missing warning


masiarek:
How to reproduce, add candidate A, B
<img width="321" alt="image" src="https://github.com/Equal-Vote/star-server/assets/857777/2a166b88-882d-4390-ac4b-bf30329227e9">

Then add candidate "C"

Now click on "Save"

As-Is:
System ignores the last candidate “C” - there is warning/error.

Should-be:
either 
- save the candidate “C” 
- or 
- warn that C will be lost (Dear user - click on button “Add”)

Link to doc:
https://docs.google.com/document/d/1R-qvnSC_YqjmlNUB-DOTBa_GFoHVkpJ1faQ397nN_m8/edit

ArendPeter:
Here are the tasks for this issue

- [ ] Remove add button from candidate creation flow (it should work the same as quick poll on the front page)
- [ ] Hide edit button based on the CANDIDATE_DETAILS feature flag
- [ ] Allow candidate names to be edited without requiring the edit flag
- [ ] Add auto scroll to the bottom of the window (you can reuse the scrolling logic from the details section of the results page)

PR #649, opened 2024-08-17
Fixed issue #476

csamuele:
## Description
Removed add button from creation flow to make process similar to quick poll
Hide edit button based on CANDIDATE_DETAILS flag
Allow candidate name to be edited without flag
Auto scroll to bottom of window

https://github.com/user-attachments/assets/1ed7c11f-d95d-429e-bfa0-016c70ba6e4e

https://github.com/user-attachments/assets/146d55dc-4c05-466b-9bae-612fddcf13e6

## Related Issues

``fixes issue #476``


ArendPeter:
For future reference, in your description you should remove the back ticks on \``fixes issue #476`` . That way github will automatically resolve the issue when the PR is merged :)

packages/frontend/src/components/ElectionForm/Races/RaceDialog.tsx
```diff
@@ -5,10 +5,13 @@ import { StyledButton } from '../../styles';
...
 
 
-export default function RaceDialog({ onSaveRace, open, handleClose, children }) {
+export default function RaceDialog({ onSaveRace, open, handleClose, children, editedRace }) {
+
 
     const { election } = useElection()
     const handleSave = () => {
+        //delete any blank candidates
+        editedRace.candidates = editedRace.candidates.filter(candidate => candidate.candidate_name !== '')
```
ArendPeter, line 14: nit

If there's a validation error onSave, will the blank candidate be added back?

csamuele:
Ah ok, I wasn't sure which way I was supposed to do it. Thanks for the clarification

ArendPeter (changes requested):
Great work, this is exactly how I was imagining it!

There's one edge case to double check, but aside from that it looks good

packages/frontend/src/components/ElectionForm/Races/RaceDialog.tsx
csamuele: Do you want the blank candidate to be added back? I'm my testing it wasn't, but I can change it so it will

csamuele:
I have added the requested change, I also memoized the race form so it would run a little faster. I was having some delays while typing.

packages/frontend/src/components/ElectionForm/Races/RaceForm.tsx
```diff
@@ -11,64 +11,98 @@ import { Box, FormHelperText, Radio, RadioGroup, Stack } from "@mui/material"
...
+        [editedRace.candidates]
+    );    // BottomRef creates a reference to the bottom of the window to scroll to when a new candidate is added
+    const bottomRef = useRef(null);
+    useEffect(() => {
+            bottomRef.current.scrollIntoView(false)
+    }, [ephemeralCandidates])
 
-    const onEditCandidate = (candidate: Candidate, index) => {
-        setErrors({ ...errors, candidates: '', raceNumWinners: '' })
+    const onEditCandidate = useCallback((candidate, index) => {
```
ArendPeter, line 37: Neat, I haven't used useCallback much before, this looks like a good use case

ArendPeter (approved):
Amazing work! Very exciting

packages/frontend/src/components/ElectionForm/Races/RaceForm.tsx
csamuele: Yeah it was getting bogged down from having to be recreated on every key stroke, and it's not too much effort to have it useCallback.

Issue #645, opened 2024-08-14
Allow admin to specify rank limit


ArendPeter:
When we launch full public beta, we'll be creating a multi-method presidential poll as one of first promotion efforts

There'll likely be 8 candidates, however we're planning to limit Ranked ballots to 6 ranks. This makes the UI easier, offers parity with STAR, and it's also consistent with real world RCV elections (San Francisco is the only counter example, they allow up to 10 ranks)

Currently the number of ranks on the ballot matches the number of candidates, however it would be great if Admin's had a dropdown in the "Extra settings section", and they could set the maximum rank limit between 3 to 8. The number of ranks shown on the ballot would then be ``min(max_rankings, number_of_candidates)``

![image](https://github.com/user-attachments/assets/490f3ce9-1750-4f89-bdab-369bf6ec4009)

ArendPeter:
You'll need to run the backend in order to test this. Here's similar change that can be used as a reference

https://github.com/Equal-Vote/star-server/pull/385/files

PR #653, opened 2024-08-30
Admin can now set rank limits in ranked ballots

csamuele:
## Description
Admins have the option of limiting the amount of ranks on a ranked ballot to a number between 3 and 8. If no option is selected, the MAX_BALLOT_RANKS environment variable will be used instead

https://github.com/user-attachments/assets/4ce376e9-9821-43fe-96b0-e4e69da03fa5

## Related Issues

to close issue #645

### DIFF
--- a/packages/frontend/src/components/LandingPage/LandingPageHero.tsx
+++ b/packages/frontend/src/components/LandingPage/LandingPageHero.tsx
@@ -20,6 +20,7 @@ import { useSubstitutedTranslation } from '../util'
 import { useTranslation } from 'react-i18next'
 import { CreateElectionContext } from '../ElectionForm/CreateElectionDialog'
 import { Button } from '@mui/material'
+import { max } from 'date-fns'
 
 export default ({}) => {
     const authSession = useAuthSession();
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/LandingPage/LandingPageHero.tsx
```diff
@@ -96,7 +97,8 @@ export default ({}) => {
                 email: ''
             },
             setReceiptEmail: () => {},
-            onUpdate: onUpdate
+            onUpdate: onUpdate,
+            maxRankings: Number(process.env.REACT_APP_MAX_BALLOT_RANKS)
```
ArendPeter, line 101: I think it would be better to reference the environment variable in RankedBallotView

If someone leaves the max ranking undefined, and then we decide to update MAX_BALLOT_RANKS later then I want it to apply to all elections that left it undefined. RankedBallotView can look at the min between MAX_BALLOT_RANKS and ballotContext.maxRankings

I think undefined would be a good maxRankings value. 999999 would work as well, but I think undefined is more clear

ArendPeter (changes requested):
Great work! Just requested a minor change

packages/frontend/src/components/LandingPage/LandingPageHero.tsx
csamuele: Ok, I updated it to reference it in RankedBallotView instead. Good catch

ArendPeter (approved):
Looks great!

Issue #546, opened 2024-04-11
Rework "Election Title / Description" of admin page


ArendPeter:
![image](https://github.com/Equal-Vote/star-server/assets/9289903/1f51c6e2-f40f-41de-9e62-cbee2d49073f)

Here are the updates I'd like to see
* Remove the **key : value** pattern. We can just display the information in a way that it's obvious which is the title/description
* Add Election state (that's communicated in the "elections you manage" table, but not on the admin page). We could do tokens similar to github. For the [brand colors from the CSS](https://github.com/Equal-Vote/star-server/blob/main/packages/frontend/src/index.css#L16)

**Github Example**
![image](https://github.com/Equal-Vote/star-server/assets/9289903/b39cc31b-01e1-4f81-bb75-31c4789c4905)

PR #654, opened 2024-08-31
Updated the look of the Election details form

csamuele:
## Description
Got rid of the key:value pattern for the Election details form and replaced it with typography meant to signal it instead for a cleaner look. Added a chip to display election state. 

image

## Related Issues

fixes issue #546

### DIFF
--- a/packages/frontend/src/components/ElectionForm/Details/ElectionDetailsInlineForm.tsx
+++ b/packages/frontend/src/components/ElectionForm/Details/ElectionDetailsInlineForm.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useMemo, useState } from 'react'
 import Grid from "@mui/material/Grid";
 import { Box, IconButton, Paper, Typography } from "@mui/material"
 import ElectionStateChip from './ElectionStateChip';
@@ -21,6 +21,18 @@ export default function ElectionDetailsInlineForm() {
         const success = await onSave()
         if (success) handleClose()
     }
+    const timeRange = useMemo(() => {
+        if (election.start_time && election.end_time) {
+            return `${formatDate(election.start_time, election.settings.time_zone)} - ${formatDate(election.end_time, election.settings.time_zone)}`
+        } else if (election.start_time) {
+            return `Starts ${formatDate(election.start_time, election.settings.time_zone)}`
+        } else if (election.end_time) {
+            return `Ends ${formatDate(election.end_time, election.settings.time_zone)}`
+        } else {
+            return 'none'
+        }
+    }, [election.start_time, election.end_time, election.settings.time_zone])
+
 
     return (
         <Paper elevation={3}>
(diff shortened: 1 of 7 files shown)


csamuele:
I couldn't really decide if it looked better to display the state above the title, or inline with it. Eventually I went with inline, but I can show you above as well if you are curious. I also added some new colors to the palette that are just more transparent versions as what we have already for cases like these where we need a good background color to match some text. This is one of those things that using Sass would help a lot with.

packages/frontend/src/components/ElectionForm/Details/ElectionDetailsInlineForm.tsx
```diff
@@ -23,34 +24,34 @@ export default function ElectionDetailsInlineForm() {
...
                 >
-                    <Grid xs={12}>
-                        <Typography gutterBottom variant="h4" component="h4">
-                            Election Title: {election.title}
-                        </Typography>
+                    <Grid item container xs={11}>
+                    <Grid item xs={12}>
+                                <Typography variant="h3" component="h4">
+                                    {election.title}
+                                    <ElectionStateChip state={election.state} />
```
ArendPeter, line 39: Nice work setting this up as a component! I have some other places in mind where this can be reused, so this makes things really convenient :D

packages/frontend/src/components/ElectionForm/Details/ElectionDetailsInlineForm.tsx
```diff
@@ -23,34 +24,34 @@ export default function ElectionDetailsInlineForm() {
...
+                            </Grid>
+                        {election.description && <Grid xs={12}>
+                            <Typography gutterBottom variant="h6" component="h5">
+                                {election.description}
+                            </Typography>
+                        </Grid>}
+                        {(election.start_time || election.end_time) && 
+                            <Grid xs={12}>
+                            <Typography sx={{mt: 2}} component="p" variant='subtitle2'>
+                                {election.start_time ? formatDate(election.start_time, election.settings.time_zone) : 'none'} - {election.end_time ? formatDate(election.end_time, election.settings.time_zone) : 'none'}
```
ArendPeter, line 50: I like this format when both are defined, but I think "none - August 13th 2024" could look a bit awkward. 

Maybe it's worth breaking this into separate cases, and you can do "starts at <date>" or "ends at <date>" whenever only one of them is defined?

ArendPeter (changes requested):
Awesome work, I love it!

I just left a nitpick on the date format

ArendPeter:
> I couldn't really decide if it looked better to display the state above the title, or inline with it. Eventually I went with inline, but I can show you above as well if you are curious. I also added some new colors to the palette that are just more transparent versions as what we have already for cases like these where we need a good background color to match some text. This is one of those things that using Sass would help a lot with.

hmm, I wonder if we can prefix it inline? Github can't do this since there can be many tags, but I think could work here since there's always going to be exactly one token

But I also think it looks great as is, so I'm happy with wherever you land on it

packages/frontend/src/components/ElectionForm/Details/ElectionDetailsInlineForm.tsx
csamuele: Ok, I updated it to dynamically display the date base on if one or both is set.

ArendPeter (approved):
Looks great! Nice work

PR #656, opened 2024-09-02
Hitting back when you have a ballot or voter selected takes you back to the corresponding table

### LINKED ISSUE #474: Hitting back when you have a ballot or voter selected should take you back to the corresponding table
https://github.com/Equal-Vote/star-server/assets/9289903/99043d07-2eac-4c01-adb1-af20fe82094f

This issue occurs with the ballots table, and also the voters table (and that's only visible in elections that restrict the voter list)

I think this can be achieved by adding window.pushstate when you click "view voter" or "view ballot"

https://stackoverflow.com/questions/59567134/what-are-the-differences-between-using-the-link-component-in-react-router-and-ca

csamuele:
## Description
fixed bug where hitting back didn't take you to the right table when viewing ballot or voter. 

https://github.com/user-attachments/assets/1999162e-f1ce-47df-a870-1f57fd8acd72

## Related Issues

> fixes #474

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,6 +13,21 @@
         "cwd": "${workspaceRoot}/backend",
         "internalConsoleOptions": "openOnSessionStart",
         "skipFiles": ["<node_internals>/**", "node_modules/**"]
+      },
+      {
+        "name": "Vite Frontend Debug",
+        "type": "chrome",
+        "request": "launch",
+        "url": "http://localhost:3000",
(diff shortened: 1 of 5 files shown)


csamuele:
I tried implementing both solutions I talked about. and ultimately realized that I could do it with just a few lines of code if I didn't use router. I also realized that the onClose function was reloading the entire list each time anyway, so it didn't really matter that I was keep them in memory.

ArendPeter (approved):
Looks great! and thanks for adding the .vscode tooling as well. I'm looking forward to playing with those

Issue #652, opened 2024-08-28
Add ballot error warnings on RCV ballots


ArendPeter:
- Duplicate Ranks
 - Overvotes
 - Skipped Ranks

PR #661, opened 2024-09-04
652 ballot error

csamuele:
## Description
Added warnings for ranked choice errors, we can discuss any differences in design, with this done it should be easy to implement anything that should be different. 

https://github.com/user-attachments/assets/b06a40ab-34a5-426a-b084-a9bb74646e0c

## Related Issues

closes issue #652


ArendPeter:
This looks amazing! and thanks for the refactor as well!

I'll do a full review soon, but here are my initial thoughts
* No need to include the best / worst. We may change our mind but for now these are based on the official reference ballots. https://www.equal.vote/ranked_robin https://www.equal.vote/star
* The red on the warning message is a little hard to read. Maybe you can use a softer red, along with an icon ([example](https://as2.ftcdn.net/v2/jpg/01/30/79/89/1000_F_130798901_PIcXrTi0TFtdAE6oVEPm1Ts2VRGPn4R3.jpg))? The MUI Paper component could also look good here, you can use the DraftWarning component as a reference
* I haven't checked the code yet, but can you confirm that Ranked Robin is unaffected by these changes? We might add the skipped column highlights to RR in the future, but for now it shouldn't be affected

csamuele:
Ah ok, I wasn't sure about the best/worst. My console kept throwing warnings that it couldn't find them. I can update that.

csamuele:
Ranked Robin slipped my mind, that can be fixed with just a couple lines of code

packages/shared/src/domain_model/Candidate.ts
```diff
@@ -10,4 +10,5 @@ export interface Candidate {
     candidate_url?: string; // link to info about candidate
     partyUrl?:      string; // link to info about party
     photo_filename?:string; // link to info about party
+    score?:         number; // score for the candidate
```
mikefranze, line 13: Rather than add score here, maybe we should create a new type that extends from the candidate that is only used in the vote pages.

ArendPeter:
Looks good, great work! Thanks for applying the feedback, and fixing the Tip component

PR #662, opened 2024-09-07
fixed maxRankings error on RankedBallotView

csamuele:
## Description
I noticed that the ranked ballot view was broken again. It looks like the environment variable somehow got changed in RankedBallotView around line 40.


ArendPeter:
We chatted about this PR in slack. default ranks and max ranks are different. At the moment the ranks default to 6, but an admin can override to be 8. 

We can either close this PR, or update it to add the default_ranks variable to sample.env and provide default behavior when the environment variables aren't specified

PR #668, opened 2024-09-15
Adding Basic STAR PR widgets

mikefranze:
## Description
Adds widgets for STAR PR using existing components.
First widget is a round by round bar chart with a pagination button to navigate rounds.
Second widget is a basic table showing all the round scores. 

This doesn't do any special highlighting because the existing components don't work for multi winner highlighting. 

image


ArendPeter:
Looks great! I like the pagination component. The tab approach didn't scale as when there were too many rounds

Also the build is failing because I pushed some broken changes, but Samuel fixed them so they should be resolved once the PR is merged

ArendPeter (approved):

PR #670, opened 2024-09-15
Adding block STV and IRV

mikefranze:
## Description
Updates the IRV a bit to allow block IRV and STV

- Adds a ballot weight fraction to each vote
- Groups ballots by their currently supported candidate to make it easier to reallocate them. Before we were just going through all ballots to determine vote count totals
- Now uses some tie breaker logic and applies tiebreak order if that fails
- Uses fixed droop quota if using STV, otherwise uses quota based on current active votes
- If using STV, apply fractional weight on ballots based on surplus
- If not using STV, reset candidate pool without winners, reallocate votes, and starts IRV process over until enough candidates have been elected
- Adds round results to keep track of who has been elected or eliminated

Follow up tasks are to add a results page for STV and update IRV results to make it clear when candidates are elected or eliminated. 

none at the moment. Can't get my local server running so can't test outside of unit tests.


packages/backend/src/Tabulators/IRV.ts
```diff
@@ -23,94 +42,190 @@ export function IRV(candidates: string[], votes: ballot[], nWinners = 1, randomT
...
+
+
+function distributeVotes(remainingCandidates: candidate[], candidateVotes: weightedVote[][], exhaustedVotes: weightedVote[], votesToDistribute: weightedVote[]) {
+    for (let i = votesToDistribute.length - 1; i >= 0; i--) {
+        // loop backwards over the ballots so exhausted votes can be removed
+        let ballot = votesToDistribute[i]
+        let topRemainingRank: number = 0
+        let topRemainingRankIndex: number = 0
+        let isOverVote = false
+        remainingCandidates.forEach(candidate => {
```
ArendPeter, line 188: nit: I think Array.reduce could be a good fit here

packages/backend/src/Tabulators/IRV.ts
```diff
@@ -23,94 +42,190 @@ export function IRV(candidates: string[], votes: ballot[], nWinners = 1, randomT
...
+        candidateVotes[i] = [];
+    }
+
+    // Initial vote distribution, moves weighted votes into the appropriate candidate pools 
+    distributeVotes(remainingCandidates, candidateVotes, exhaustedVotes, weightedVotes)
+
+    // Set quota based on number of winners and if its proportional
+    let quota = 0
+    if (proportional) {
+        quota = Math.floor(activeVotes.length/(nWinners + 1) + 1)
```
ArendPeter, line 75: How do quota rules work when there's polarized factions? If there's 4 polarized factions fighting over 2 seats, then is it possible that no one reaches the quota?

Edit: I see that the non-proportional case recalculates the quota to account for this, but I feel like this is necessary for STV as well

Edit2: now I see we have a section where we elect the remaining candidates if remainingCandidates <= remainingSeats

packages/backend/src/Tabulators/IRV.ts
```diff
@@ -23,94 +42,190 @@ export function IRV(candidates: string[], votes: ballot[], nWinners = 1, randomT
...
-            remainingCandidates = remainingCandidates.filter(c => !results.elected.includes(c))
-            // TODO: Some STV stuff
+            roundResults.winners.push(summaryData.candidates[winningCandidateIndex])
+            if (proportional) {
+                remainingCandidates = remainingCandidates.filter(c => !results.elected.includes(c))
+                let fractionalSurplus = new Fraction(maxVotes - quota).div(maxVotes)
+                let winningCandidateVotes = candidateVotes[winningCandidateIndex]
+                candidateVotes[winningCandidateIndex] = []
+                winningCandidateVotes.forEach(vote => {
+                    vote.weight = vote.weight.mul(fractionalSurplus).floor(5)
```
ArendPeter, line 133: The weight is already a fraction type, why do we need to floor it?

packages/backend/src/Tabulators/IRV.ts
```diff
@@ -23,94 +42,190 @@ export function IRV(candidates: string[], votes: ballot[], nWinners = 1, randomT
...
+        let sortedVoteCounts = [...roundVoteCounts].sort((a, b) => {
+            if (a.voteCount !== b.voteCount) {
+                return (b.voteCount - a.voteCount)
             }
-            else {
-                // give vote to top rank candidate
-                roundVoteCounts[topRemainingRankIndex] += 1
+
+            // break tie in favor of candidate with higher score in previous rounds
+            for (let i = results.voteCounts.length - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
mikefranze: Yes this is the Scottish STV rules for tiebreaking. Seems to apply to IRV as well except for San Francisco and Oakland rules. Looking at it now only thing that might break here is it could look at previous rounds of block IRV, but maybe that's ok.

packages/backend/src/Tabulators/IRV.ts
```diff
@@ -1,16 +1,35 @@
...
 
-export function IRV(candidates: string[], votes: ballot[], nWinners = 1, randomTiebreakOrder:number[] = [], breakTiesRandomly = true) {
-    // Determines Instant Runoff winners for given election
+
+const Fraction = require('fraction.js');
+
+type weightedVote = {
+    weight: typeof Fraction,
+    vote: ballot,
+    overvote: boolean
```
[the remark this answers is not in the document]
mikefranze: I'd prefer to not rely on the front end for that. We could someday connect to different frontends that don't have those protections and I'd like the tabulators to be able to validate votes independently. Same reason we check that STAR votes aren't outside of 0-5.

packages/backend/src/Tabulators/IRV.ts
mikefranze: Yup, we will eliminate candidates until one gets above the quota or we get enough candidates to fill all seats.

packages/backend/src/Tabulators/IRV.ts
mikefranze: This is also used in Scottish STV. We don't really need it but wanted to be consistent. Even with floor we still need fractions because of floating point errors (0.2 + 0.1 < 0.3). What the fraction floor method does is set the denominator to 10000 so it keeps the fraction sizes manageable and consistent, rather than 12341245/12342343547 you have 999/10000.

packages/backend/src/Tabulators/IRV.ts
mikefranze: Not really familiar with reduce outside of computing sums. I got something to work but it's not that clean given that it has to output both the rank and index. I'm going to leave it alone for now.

PR #675, opened 2024-09-19
665 add candidate UX

### LINKED ISSUE #665: "Add Candidate" UX
I like the new Add Race flow, but there's a few things we can add to further improve it

- [ ] **jarring candidate addition**: When I start typing a new candidate name, the new candidate automatically gets added and this causes the scroll to suddenly change. I'm not sure how to fix this. Maybe we can smooth out the scroll animation? (for example you can reference the scroll animation when expanding a details panel on the results page). I tested this on Trello, they don't do a smooth scroll, but they scroll when the new element is focused instead of when the typing starts. That feels a lot better
- [ ] **candidate delete edge cases**: I like how clearing a candidate deletes it, but it's not consistent. If I create candidates A, B, C, then clear B there's a gap. Then if I hit the delete button on B, and try to do the same on C I (incorrectly) get a validation error. That said clearing C does work and it gets deleted. 
- [ ] **dragable candidates**: As a stretch goal, it could be cool to replace the arrows with a "drag handle" / 6 dots icon and then support dragging to reorder. We'd need to test if this works on mobile. Trello could be a good reference point again

image

csamuele:
## Description
Updated added candidate UX to make it scroll when the new element is focused instead of when it is added. 
I updated the delete behavior to make it so a candidate is deleted when they are cleared wherever it is in the list, unless there is only one left (not including the grayed out one). I made it so you have to press spacebar again on the cleared candidate before it is deleted though, because it can be frustrating if it happens automatically while you are trying to edit a candidate. 

I also replaced the up and down arrows with a drag and drop list. 

## Screenshots

For some reason my screen recorder isn't working. 

## Related Issues

 closes #665

### DIFF
--- a/packages/frontend/src/components/Election/Voting/RankedBallotView.tsx
+++ b/packages/frontend/src/components/Election/Voting/RankedBallotView.tsx
@@ -104,7 +104,6 @@ export default function RankedBallotView({ onlyGrid = false }) {
     return columnValues.map(v => t('number.rank', { count: v, ordinal: true }));
   }, [columnValues, t]);
 
-  console.log(ballotContext.race.voting_method);
   return (
     <GenericBallotView
       key="rankedBallot"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1737,6 +1737,55 @@
         "date-fns": "^2.0.0"
       }
     },
+    "node_modules/@dnd-kit/accessibility": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/@dnd-kit/accessibility/-/accessibility-3.1.0.tgz",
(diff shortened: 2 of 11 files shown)


packages/frontend/src/components/Election/Voting/RankedBallotView.tsx
```diff
@@ -104,7 +104,6 @@ export default function RankedBallotView({ onlyGrid = false }) {
     return columnValues.map(v => t('number.rank', { count: v, ordinal: true }));
   }, [columnValues, t]);
 
-  console.log(ballotContext.race.voting_method);
```
ArendPeter, line 107: my bad 😅, thanks for removing that!

ArendPeter (approved):
Looks great! I'm excited to play with this

PR #680, opened 2024-09-19
664 election state chip for election table

### LINKED ISSUE #664: Apply ElectionStateChip to election tables
The ElectionStateChip component has been a great visual for the Admin Home page. We should also add it to https://dev.star.vote/ElectionsYouManage to keep things consistent.

I think that's the only other spot that displays election state, but feel free to double check for others

csamuele:
## Description
updated the election state column to use the colored chips in the elections table

image

## Related Issues

closes #664

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -2,6 +2,11 @@
     "version": "2.0.0",
     "tasks": [
       {
+        "label": "Build",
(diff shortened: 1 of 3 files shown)


ArendPeter (approved):
Looks great, nice work!

PR #683, opened 2024-09-21
Fixing STAR PR bug

mikefranze:
## Description

Arend found a bug in the STAR PR tabulator when the number of ballots is less than or equal to the number of winners.

When this happens the quota to win a seat becomes 1 or less, meaning for the first round that one voter will always fill or exceeds the quota. One function tries to find the score of the voter right before the quota is filled to find the split point for adjusting ballot weights. But if 1 voter fills the quota than the slit point is undefined. 

This change just says to count the first voter as under the quota so there will always be a split point.

Update: Noticed the results weren't rounding which was leading to filling the bar chart with long repeating numbers. Updated it to round to first decimal. 
image


ArendPeter (approved):
Looks good, great work!

PR #686, opened 2024-09-25
added json download option for ballot data

### LINKED ISSUE #682: Add option to download ballot data as JSON
Perhaps I could convert DownloadCSV to a dropdown so users can select CSV or JSON

csamuele:
image

## Related Issues
closes #682

### DIFF
--- a/packages/frontend/src/components/Election/Results/BallotDataExport.tsx
+++ b/packages/frontend/src/components/Election/Results/BallotDataExport.tsx
@@ -1,5 +1,5 @@
-import { CSVLink } from 'react-csv';
-import { useState, useEffect } from 'react';
+import { CSVLink, CSVDownload } from 'react-csv';
+import React, { useState, useEffect } from 'react';
 import { Election } from '@equal-vote/star-vote-shared/domain_model/Election';
 import { AnonymizedBallot, Ballot } from '@equal-vote/star-vote-shared/domain_model/Ballot';
 import { useGetAnonymizedBallots } from '~/hooks/useAPI';
@@ -7,6 +7,9 @@ import MenuItem from "@mui/material/MenuItem";
 import BorderAll from '@mui/icons-material/BorderAll';
 import DataObject from '@mui/icons-material/DataObject';
 import { MenuButton } from '~/components/MenuButton';
+import ReactDOM from 'react-dom';
+import { cs } from 'date-fns/locale';
+import { set } from 'date-fns';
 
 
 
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/Election/Results/BallotDataExport.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+
+interface Props {
+    election: Election;
+}
+
+export const BallotDataExport = ({ election }: Props) => {
+    const [csvData, setCsvData] = useState([]);
+    const [csvHeaders, setCsvHeaders] = useState([]);
+	const { data, error, isPending, makeRequest } = useGetAnonymizedBallots(election.election_id);
```
ArendPeter, line 23: I'm realizing that this could be a expensive call for larger elections, can you have it wait to do the call until the download is confirmed?

packages/frontend/src/components/Election/Results/BallotDataExport.tsx
```diff
@@ -0,0 +1,97 @@
...
+	useEffect(() => { makeRequest() }, [])
+	const ballots: AnonymizedBallot[] = data?.ballots || [];
+    const buildCsvData = () => {
+
+        let header = [
+            { label: 'ballot_id', key: 'ballot_id' },
```
ArendPeter, line 29: Could you also include precincts here too?

packages/frontend/src/components/Election/Results/BallotDataExport.tsx
```diff
@@ -0,0 +1,97 @@
...
+        a.download = `Ballot Data - ${limit(election.title, 50)}-${election.election_id}.json`;
+        a.click();
+    }
+
+    return (
+        <>
+            {!isPending && (
+                <MenuButton 
+                    label={"Download"}
+                    menuItems={[
```
ArendPeter, line 71: nitpick

I think it would be cleaner to add MenuItems as children to the MenuButton rather than creating a separate variable for input

ArendPeter (changes requested):
Left some minor comments

packages/frontend/src/components/Election/Results/BallotDataExport.tsx
csamuele: Yeah, I was having a super hard time getting it to work with the hook, so I ended up just doing it outside. But I'll keep working on it

packages/frontend/src/components/Election/Results/BallotDataExport.tsx
```diff
@@ -64,34 +74,42 @@ export const BallotDataExport = ({ election }: Props) => {
...
                             <DataObject sx={{ marginRight: 1 }} />
                             Download JSON
                         </MenuItem>
-                    ]}
+                </MenuButton>
+                    
+                  
+                
+                {csvData.length > 0 && (
+                    //For some reason, CSVDownload doesn't allow you to set the filename, so we use CSVLink instead,
```
ArendPeter, line 94: Good context, sounds like it was a tricky fix

ArendPeter (approved):
Look great!

PR #688, opened 2024-09-27
fix ballot validation

csamuele:
## Description
fixed errors with posting ballots where it was invalidating proper ballots.


ArendPeter (approved):
Looks great!

PR #706, opened 2024-10-15
made hyperlinks green

### LINKED ISSUE #695: Web urls should be highlignted green
I initially tried this by adding a selector for a tags in the css, but that selection was too broad and also captured a lot of the buttons

Be sure to use var(--brand-green) from the index.css

csamuele:
image

## Related Issues

closes #695


ArendPeter (approved):
Great work!

I'm confused why :visited had that side effect but we can look into that more later. Thanks for finding a fix!

PR #707, opened 2024-10-15
fixed frontend to comply with max rankings validation

### LINKED ISSUE #690: Fix  Rank Limit validation
The backend validation of the rank limit would give an error when trying to change rank limit in the front end

It's been disabled temporarily but this needs a proper fix
https://github.com/Equal-Vote/star-server/commit/89061cf8b30f3bed684a74029de18cc6825d7547

csamuele:
## Description
max rankings was being sent as a string rather than a number, I fixed the frontend to send it as a number

## Related Issues

closes #690


ArendPeter (approved):
Great work!

PR #709, opened 2024-10-21
Change to Nunito Sans

fsargent:
## Description
Updated typography to use a Sans Serif font.
Modified theme so that buttons could use lower case.

OLD:
image

NEW:
image

OLD:
image

NEW
image

### DIFF
--- a/packages/frontend/src/theme.tsx
+++ b/packages/frontend/src/theme.tsx
@@ -157,41 +157,19 @@ const brandPalette: PaletteOptions = {
 }
 
 const brandTypography: TypographyOptions = {
-  // fontFamily: 'Montserrat',
-  fontFamily: 'Verdana, sans-serif',
-  button:{
-    fontFamily: 'Montserrat, Verdana, sans-serif',
+  fontFamily: '"Nunito Sans", sans-serif',
+  button: {
+    fontFamily: '"Nunito Sans", sans-serif',
+    textTransform: 'none'
   },
-  h1: {
-    fontFamily: 'Montserrat, Verdana, sans-serif',
-    marginTop: '1rem',
-    marginBottom: '1rem',
-  },
-  h2: {
-    fontFamily: 'Montserrat, Verdana, sans-serif',
-    marginTop: '1rem',
-    marginBottom: '1rem',
-  },
-  h3: {
-    fontFamily: 'Montserrat, Verdana, sans-serif',
-    marginTop: '1rem',
-    marginBottom: '1rem',
-  },
-  h4: {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/theme.tsx
```diff
@@ -157,41 +157,19 @@ const brandPalette: PaletteOptions = {
...
-    fontFamily: 'Montserrat, Verdana, sans-serif',
-    marginTop: '1rem',
-    marginBottom: '1rem',
-  },
-  h6: {
-    fontFamily: 'Montserrat, Verdana, sans-serif',
-    marginTop: '1rem',
-    marginBottom: '1rem',
-  }
+  ...Object.fromEntries(['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].map(level => [
```
ArendPeter, line 165: Nice! Much cleaner

packages/frontend/src/theme.tsx
```diff
@@ -157,41 +157,19 @@ const brandPalette: PaletteOptions = {
...
+    level,
+    {
+      fontFamily: '"Nunito Sans", sans-serif',
```
ArendPeter, line 168: We chose Montserrat with Verdana as a fallback in order to be consistent with Equal Vote branding, those should also be I think those are also sans serif?

packages/frontend/src/theme.tsx
```diff
@@ -157,41 +157,19 @@ const brandPalette: PaletteOptions = {
...
 
 const brandTypography: TypographyOptions = {
-  // fontFamily: 'Montserrat',
-  fontFamily: 'Verdana, sans-serif',
-  button:{
-    fontFamily: 'Montserrat, Verdana, sans-serif',
+  fontFamily: '"Nunito Sans", sans-serif',
+  button: {
+    fontFamily: '"Nunito Sans", sans-serif',
+    textTransform: 'none'
```
ArendPeter, line 163: I don't have a strong opinion about the all caps, I'm fine with removing the textTransform to see how it looks

ArendPeter (commented):
Looks good!

Could you clarify the choice of Nunito sans? Monserrat and Verdana should also be sans-serif, and they have the benefit of mirroring the Equal Vote branding 

I'm happy to reconsider the font choice, but I wanted to have that context included

fsargent:
Most of it is just that I think it looks better. I can remove the font choices and just keep the code improvements. The best thing is allowing the buttons to be lowercase, I think. I will say it doesn’t look like it matches equal.vote right now because of the color choices. Is that a design consideration?
Nunito sans may be a bit arbitrary. Suggestions welcome.

PR #710, opened 2024-10-21
10 21 decrease gaps between elements in the landing pages

fsargent:
## Description

- Removed *huge* gap value.

Old:
bettervoting com_ (1)
New:
CleanShot 2024-12-22 at 11 30 58@2x
CleanShot 2024-12-22 at 11 31 35@2x
CleanShot 2024-12-22 at 11 31 55@2x

## Related Issues

### DIFF
--- a/packages/frontend/index.html
+++ b/packages/frontend/index.html
@@ -22,7 +22,7 @@
     <link rel="manifest" href="/manifest.json" />
     <!-- Importing for ballot bubble font-->
     <link href='https://fonts.googleapis.com/css?family=Archivo Black' rel='stylesheet'>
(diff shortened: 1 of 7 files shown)


fsargent:
Ok, so is there anything I can do to get this merged in then? Let me know!

ArendPeter:
> Ok, so is there anything I can do to get this merged in then? Let me know!

Sorry for the delays! I think remove the font change, and making gap slightly larger (it went from 20 to 1, but I think something in the 3-5 range could still add some helpful separation). Then I'll merge add we'll see how we feel with it

For the STAR Ballot I noticed the new font caused the text to wrap so I think we should hold off on that
image

fsargent:
@ArendPeter Finally fixed this up. I think it looks a lot better. Reverted the font changes and tried to be a bit more consistent about gaps at 2rem.

PR #714, opened 2024-11-16
Precincts Update

mikefranze:
## Description
Small updates to get precincts to work.

A precinct is a subset of voters that share a unique ballot. If a race is given a list of precincts than only voters in those precincts can vote in that race. If a race has no precincts then all voters can vote in that race. Each voter can only have one precinct and can only vote in races that contain that precinct or no precint.

For example, an organization has an election to vote for its new president and leaders of each of its chapters A-D. There would be 4 precincts:
A
B
C
D

All voters can vote in presidential election, so it has no precincts. 
image

Chapter A would have precinct A, Chapter B would have precinct B, etc
image

When adding voters, check the precinct box and add a second column with the precinct for each voter. If a voter has no precincts than leave the precinct column blank but include a comma. In this example chapter A has voters voter_1a-voter_4a, chapter B has voters voter_1b-voter_4b, and there are four voters voter_1-voter_4, who don't belong to a chapter but can still vote for president.

```
voter_1a,A
voter_2a,A
voter_3a,A
voter_4a,A
voter_1b,B
voter_2b,B
voter_3b,B
voter_4b,B
voter_1,
voter_2,
voter_3,
voter_4,
```

image

voter_1a then enters their ID (or receives an email link if using emails) 
image

voter_1a is then able to vote in both the Presidential race and Chapter A race but not the Chapter B race
image
image

The processed results then show one vote for president, one for chapter A, and none for Chapter B
image

Next, voter_1 votes and can only vote in the presidential race
image


ArendPeter (approved):
Looks great! Thanks for the thorough example

I also fixed the build errors on the main branch so I'll go ahead and merge

PR #716, opened 2024-11-27
WIP: Email campaign API

mikefranze:
## Description
Adds new API endpoint for sending email campaigns. Similar to the send invites but more generic with custom subjects, body text, and ability to send invites or just regular email. See expected request data format in email_request_data type. 

Still need to test it and add swagger info to route page but want to get visibility on it.


ArendPeter (approved):
Looks great!

PR #717, opened 2024-12-16
Fix edge case for single fractional surplus

ArendPeter:
This PR addresses a split point issue and abstentions

## Split Point Off-by-One Error

This addresses an of by one error when selecting the split point.

Currently after Marge is elected in [the Simpsons election](https://bettervoting.com/frmcv6/results) the log outputs

```
The 5 voters who gave Marge Simpson 5 stars are partially represented and will be reweighted by 1.12 for future rounds.
```

This is incorrect since a split point at 5 stars is not enough to fill the quota. (There are 5 voters who gave her 5 stars, but the quota is 5.6)

The split point should be set based on the first ballot to exceed the quota, however it current sets the split point based on the last ballot before exceeding the quota. 

Here's the updated log after the fix

```
The 5 voters who gave Marge Simpson more than 4 stars are fully represented and will be weighted to 0 for future rounds.

The 1 voters who gave Marge Simpson 4 stars are partially represented. 60% of their remaining vote will go toward Marge Simpson and 40% will be preserved for future rounds.
```

TODO: I still need to fix the plural logic in the wording 

## Filtering Abstentions

I noticed there was a mismatch between the vote count in the tabulator log, and at the top of the results. That because one was including abstentions and the other wasn't. I updated the tabulators to filter out abstentions

### DIFF
--- a/packages/backend/src/Controllers/Election/getElectionResultsController.ts
+++ b/packages/backend/src/Controllers/Election/getElectionResultsController.ts
@@ -36,7 +36,8 @@ const getElectionResults = async (req: IElectionRequest, res: Response, next: Ne
         const cvr: number[][] = []
         ballots.forEach((ballot: Ballot) => {
             const vote = ballot.votes.find((vote) => vote.race_id === race_id)
-            if (vote) {
+            let maxScoreGiven = vote?.scores.reduce((prev, s) => Math.max(prev, s.score ?? 0), 0) ?? 0;
+            if (vote && maxScoreGiven > 0) {
                 cvr.push(vote.scores.map((score: Score) => (
                     score.score
                 )))
--- a/packages/backend/src/Controllers/Election/getElectionResultsController.ts
+++ b/packages/backend/src/Controllers/Election/getElectionResultsController.ts
@@ -36,8 +36,7 @@ const getElectionResults = async (req: IElectionRequest, res: Response, next: Ne
         const cvr: number[][] = []
         ballots.forEach((ballot: Ballot) => {
             const vote = ballot.votes.find((vote) => vote.race_id === race_id)
-            let maxScoreGiven = vote?.scores.reduce((prev, s) => Math.max(prev, s.score ?? 0), 0) ?? 0;
-            if (vote && maxScoreGiven > 0) {
+            if (vote) {
                 cvr.push(vote.scores.map((score: Score) => (
                     score.score
                 )))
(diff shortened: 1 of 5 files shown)


packages/backend/src/Controllers/Election/getElectionResultsController.ts
```diff
@@ -36,7 +36,8 @@ const getElectionResults = async (req: IElectionRequest, res: Response, next: Ne
         const cvr: number[][] = []
         ballots.forEach((ballot: Ballot) => {
             const vote = ballot.votes.find((vote) => vote.race_id === race_id)
-            if (vote) {
+            let maxScoreGiven = vote?.scores.reduce((prev, s) => Math.max(prev, s.score ?? 0), 0) ?? 0;
```
mikefranze, line 39: This logic should be done in the tabulation parser. 
Currently it looks like we're marking undervotes as not valid but still pushing them to the scores that are used.
```
else if (ballotValidity.isUnderVote) {
            underVotes += 1
            scores.push(row)  <-----
            // under votes should not count as a valid vote
            //validVotes.push(voter);
        }
```

packages/backend/src/Tabulators/AllocatedScore.ts
```diff
@@ -374,17 +373,17 @@ function normalizeArray(scores: ballot[], maxScore: number) {
...
 
 function findSplitPoint(cand_df_sorted: winner_scores[], quota: typeof Fraction) {
-    var under_quota : any[] = [];
-    var under_quota_scores: typeof Fraction[] = [];
-    var cumsum = new Fraction(0);
-    cand_df_sorted.forEach((c, i) => {
+    let cumsum : typeof Fraction = new Fraction(0);
+    for(const c of cand_df_sorted ){
         cumsum = cumsum.add(c.ballot_weight);
-        if (cumsum < quota || i == 0) {
```
mikefranze, line 382: We should add a test case for this i == 0 scenario. I think it was added when the number of votes is less that number of seats so the first voter will always exceed the threshold.

mikefranze (approved):
Looks right, good catch!

packages/backend/src/Tabulators/AllocatedScore.ts
ArendPeter: Added!

packages/backend/src/Controllers/Election/getElectionResultsController.ts
ArendPeter: Yeah that's better. Fixed!

PR #718, opened 2024-12-22
Remove fade-in effect from landing pages.

fsargent:
The effect is bad for usability and also reset if you scroll up and down.

https://share.cleanshot.com/zLLKPRp9
Ignore width issues, unrelated to this PR.


ArendPeter (approved):
Looks good!

Yes, I agree. especially with the front page elements closer together I don't like the fade effect anymore

PR #722, opened 2024-12-23
Update pt-BR.yaml

mickaelfo:
Commit 1: another small correction for the correct gender of the verb.

Commit 2: I changed the word "desempate" to "segundo turno", because "desempate" in Portuguese can be the same as "tiebreaker". "Segundo Turno" means runoff (second round in the literal meaning).
These changes were accepted by consensus with 5 other Brazilian speakers.


ArendPeter (approved):
Awesome! Thank you for getting all the feedback

Issue #723, opened 2025-01-13
Uploading Elections / Ballots


ArendPeter:
(TODO)

ArendPeter:
Uploaded Elections are live at bettervoting.com/PublicArchive

I'll resolve this as the core item is complete, but I'm sure there will be more issues created later as we iron out the usage scenarios

PR #724, opened 2025-01-13
Implement bulk uploading ballots from the UI

ArendPeter:
## Description
I've already pushed changes adding a placeholder "upload elections" UI and also the backend upload ballots API. This PR connects the two and implements the following
 - Parse CSV
 - Create New Election & Race
 - Upload Ballots in batches
 
 This will create closed elections, since people aren't intended to vote in these through the UI, and then upload the ballots
 
 There's more work that needs to be done, checkout #723 for next steps

(no sound, only video)

https://github.com/user-attachments/assets/e6b5f7f0-50c5-4fd5-b536-d0aa5565f110


mikefranze:
Trying to follow along but with previous pushed changes its hard to tell what's new. Is there any way to tell what ballots have been uploaded by the admin and what were uploaded by the voters? Is there any way to know that ballots were uploaded?

ArendPeter:
> Trying to follow along but with previous pushed changes its hard to tell what's new. 

My bad, yeah this is building on some things I already pushed. Here are some other relevant commits prior to the PR
 - Initial Placeholder UI: https://github.com/Equal-Vote/star-server/commit/3f81c9cf932005ae2e17ee27dfb576c427adc403
 - Refactoring shared function in castVoteController: https://github.com/Equal-Vote/star-server/commit/3a8b627f666abc85a8505af46cdc727c432ddcb0
 - Implementing initial UploadBallots API: https://github.com/Equal-Vote/star-server/commit/a8b0665e769ae7e99d644c810513abb6bcef4107
 
 > Is there any way to tell what ballots have been uploaded by the admin and what were uploaded by the voters? Is there any way to know that ballots were uploaded?
 
 Not yet, what do you think the best approach would be? Perhaps we when updating the history we can set the action_type to "submitted_by_admin" instead of "submit"? OR we could add a new field to the Ballot for "submitted_by_voter"? I guess it depends on how we plan to use the information

mikefranze:
> Not yet, what do you think the best approach would be? Perhaps we when updating the history we can set the action_type to "submitted_by_admin" instead of "submit"? OR we could add a new field to the Ballot for "submitted_by_voter"? I guess it depends on how we plan to use the information

Yea not sure how exactly how the information would be used. If we find that we need a specific query for it than we would add a field. But either way adding it to the history is a good idea. The history has a field for who did the action right? Could add the admin info there and just put voter if it was the voter.

If that change is made I'd be ok approving.

ArendPeter:
> > Not yet, what do you think the best approach would be? Perhaps we when updating the history we can set the action_type to "submitted_by_admin" instead of "submit"? OR we could add a new field to the Ballot for "submitted_by_voter"? I guess it depends on how we plan to use the information
> 
> Yea not sure how exactly how the information would be used. If we find that we need a specific query for it than we would add a field. But either way adding it to the history is a good idea. The history has a field for who did the action right? Could add the admin info there and just put voter if it was the voter.
> 
> If that change is made I'd be ok approving.

Done! I updated the PR

It's probably easiest to review my most recent 4 commits individually rather than the full diff

packages/backend/src/Migrations/2025_01_29_admin_upload.ts
```diff
@@ -0,0 +1,31 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<any>): Promise<void> {
+    await db.schema.alterTable('electionDB')
+        /* ballot_source types
+        live_election: Ballots submitted by voters during election
+        prior_election: Election admin uploaded ballots from a previous election
+        */
+        .addColumn('ballot_source', 'varchar' )
```
mikefranze, line 9: It looks like in the domain model this is a non-optional field, so you should set this to be not null. publick_archive_id is optional and doesn't need this.
Example from other migration:
`.addColumn('election_id', 'varchar', (col) => col.notNull())`

mikefranze (changes requested):

packages/backend/src/Migrations/2025_01_29_admin_upload.ts
ArendPeter: Good catch, updated!

mikefranze (approved):
Looks good, thanks!

PR #734, opened 2025-01-16
fix: break any long word across lines.

### LINKED ISSUE #485: Election Title Overflow - Incorrect Display - Election Title and Description - test maximum length
Both fields Overflow:
- Title - max 130 characters
- Description (999 characters)
image

waugh:
## Description

If a word is too long to fit in a visible block that contains it, with this change, the browser will break the word across lines. The problem of this not happening was noticed particularly with election names; however, the present proposed solution applies very broadly, probably to all messages the app displays. Technically, it applies to everything in CSS class "MuiTypography-root", which is probably everything.

I'm not sure whether the location of the new declaration is appropriate. Prior, there were no "Mui"- declarations there, and I may be abusing the intended pattern for administering a project's use of the Material User Interface (MUI) library.

#### Before
before_mobile

#### After
after_mobile

#### Before
before

#### After
after

## Related Issues

Intended to resolve #485.

### DIFF
--- a/packages/frontend/src/index.css
+++ b/packages/frontend/src/index.css
@@ -49,6 +49,10 @@
   --dark-bkg-gray: #1F1F1F;
 }
 
+.MuiTypography-root {
+  overflow-wrap: break-word;
+}
+
 .racePageStepper .MuiStepConnector-line {
   opacity: 0;
 }


ArendPeter (approved):
Thank you, this look great!

We can deploy it with the broad css selector for now, and we can always tweak it afterwards if there's any edge cases.

Issue #729, opened 2025-01-14
Clear race dialog when closed


ArendPeter:
Currently when adding multiple races to an election the state of the dialog isn't fully reset

https://github.com/user-attachments/assets/715b956d-7ecb-4e03-8f9c-8aeff2505119

Instead all fields should start at empty, and the voting method stepper should also reset to the first step

ArendPeter:
@FarzeelAhmad99 Fixed the first half of this issue in #736 ! 

Still need one more piece though. In the voting method selector, the stepper is preserved when reopening the window, but this should reset to the first step

Farzeel, I'll go ahead an assign this to you, in case you're still interested

PR #736, opened 2025-01-17
Clear race dialogue on saving

FarzeelAhmad99:


ArendPeter (approved):
Nice find, great work!

ArendPeter:
I'll go ahead and merge this one, but we need one more change before #729 can be closed

In the voting method selector, the stepper is preserved when reopening the window, but this should reset to the first step

ArendPeter:
oh, and I'm currently deploying a change to add you as a contributor on https://bettervoting.com/about :D

PR #771, opened 2025-02-01
Create shared summaryData function for tabulators

ArendPeter:
## Description

New things in this PR
 - Moves ballot validation to a shared function
 - Move summaryData to a shared function
 - Renames error stats to be more specific (outOfBoundsVotes, tallyVotes, and abstentionVotes) instead of (invalidVotes, validVotes and underVotes)

It also removes some features from the backend. In general I think the tabulators should only focus on what's necessary to generate the results. Anything related to deeper analysis can be left to the frontend widgets. Examples include counting bullet votes, and the equal preferences. 

This also hard codes random tiebreakers for some of the tabulators. The backend should always do the full tiebreaker flow, and then the frontend can obscure the final random steps if the user prefers for the race to be reported as a true tie. That said there will be more tie breaker updates in a future PR


mikefranze (approved):
Looks good! I do think however that simple stats should stay in the backend. We're assuming here that bettervoting will always be the UI but there might be other clients (discord bots for example) that might want those stats without having to rewrite the same frontend code to calculate it. We can debate that later though.

mikefranze:
Also would be a good idea to check the before and after of a few elections (like DSA) to verify that the new tabulator code is producing the same results.

ArendPeter:
> I do think however that simple stats should stay in the backend. We're assuming here that bettervoting will always be the UI but there might be other clients (discord bots for example) that might want those stats without having to rewrite the same frontend code to calculate it. We can debate that later though.

Sounds good, I'm happy to talk about that further

> Also would be a good idea to check the before and after of a few elections (like DSA) to verify that the new tabulator code is producing the same results.

Good idea, I verified all the pres24 races and one of the dsa races before merging. Everything matched

PR #773, opened 2025-02-04
Create es.yaml

LukasHeaton:
Testing out translating into Spanish- have only done the first 2 lines so far


ArendPeter (approved):
Looks good!

PR #776, opened 2025-02-06
Optimize uploading ballots for recreating prior elections

ArendPeter:
## Description
This significantly reduces the time it takes to upload ballots in bulk. All benchmarks are done with the Burling 2009 mayor election which has 8980 votes. And unless otherwise stated each test was using a batch size of 100. Increasing the batch size above 100 would exceed the request size limit

**Prior optimizations**
I already pushed a changing bypassing the ballot queue system. That design is important when processing real time votes and ensuring each vote is counted exactly once, but it's not as important when analyzing prior elections. This effectively reduce the time from several minutes to 77 seconds

**Optimizations in Pull Request**
Baseline - 77s
Backend no-op - 6.7s  - This gives an estimate for how much of 
w/ reduced logging - 69.791s - Previous logging was very verbose and it slowed down the response time.
w/ bulk database update - 34.935s - Executes the bulk upload in a single database query, rather than looping with one row per query
w/ batch size of 50 - 58.495s - Decreasing the batch size make it significantly slower (and justified optimizing batch size in the other direction)
w/ batch size of 700 - 25 s - Refactor the request format to allow for larger batch sizes

the video was even faster than my benchmarks. I probably should have averaged over more iterations to get a better sense

https://github.com/user-attachments/assets/99598b67-b8dc-45d7-96e8-6ca9b58bef60


ArendPeter:
This PR got messier since I also included changes fixing issues with the rcv-errors PR

Issue #741, opened 2025-01-19
results - display issue - scoring round


masiarek:
We have a "display" issue in scoring round results.
I started testing Block-STAR - so maybe this is related.
Another difference - the first election is using "**abstention**" ballots.

Compare these two elections:
https://youtu.be/mKO2tOOh5x8 

1) First election - incorrect
https://bettervoting.com/cqr6qm/results

<img width="997" alt="Image" src="https://github.com/user-attachments/assets/31626876-dd11-4603-8be1-db9ac8a767a0" />

2) Second election - correct
https://bettervoting.com/wyycth/results

<img width="1063" alt="Image" src="https://github.com/user-attachments/assets/1e58f8ae-81fe-4e63-ac2f-b5c49674f453" />

ArendPeter:
This looks correct to me. I think you had different tabs selected when you were comparing the elections

That said we can open a new issue for fixing graphs when the have zeros. Currently the zeros overlap with the category titles and it's messy.

But I think we can close this issue

masiarek:
Closing note: per @ArendPeter's investigation above this was a tab-selection mix-up rather than a display bug, and he suggested closing at the time. The follow-on observation — zero-height bars overlapping their labels — was flagged there as deserving its own issue, and is not tracked by this one.

PR #793, opened 2025-02-15
Fix: results bar chart equal preference all zeroes

### LINKED ISSUE #756: Update ResultsBarChart to display 0 without overlapping with category names
> That said we can open a new issue for fixing graphs when the have zeros. Currently the zeros overlap with the category titles and it's messy.

 _Originally posted by @ArendPeter in [#741](https://github.com/Equal-Vote/star-server/issues/741#issuecomment-2611297004)_

recursivesquircle:
## Description
Adds a tick margin to the equal preferences results bar chart when equal preference votes are all zeroes to fix overlapping text.

image
image
image
image

Closes #756

### DIFF
--- a/packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
+++ b/packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
@@ -2,22 +2,46 @@ import { useState } from "react";
 import { Bar, Cell, ComposedChart, LabelList, Legend, Line, ResponsiveContainer, XAxis, YAxis } from "recharts";
 import { CHART_COLORS, formatPercent, truncName } from "~/components/util";
 
-export default ({
+
+interface ResultsBarChartData {
+  name: string;
+  votes?: number | undefined; 
+  count?: number | undefined; 
+  score?: number | undefined; 
+  [key: string]: any;
+}
+
+interface ResultsBarChartProps {
+  data: ResultsBarChartData[];
+  runoff?: boolean;
+  colorOffset?: number;
+  sortFunc?: ((a: ResultsBarChartData, b: ResultsBarChartData) => number) | false;
+  xKey?: 'votes' | 'count' | 'score';
+  percentage?: boolean;
+  percentDenominator?: number;
+  star?: boolean;
+  majorityLegend?: string;
+  majorityOffset?: boolean;
+  height?: number;
+  maxBarSize?: number;
+}
+
+export default function ResultsBarChart({
   data,
   runoff = false,
   colorOffset = 0,
   sortFunc = undefined,
-  xKey = "votes",
+  xKey = 'votes',
   percentage = false,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
```diff
@@ -1,6 +1,10 @@
 import { Bar, Cell, ComposedChart, LabelList, Legend, Line, ResponsiveContainer, XAxis, YAxis } from "recharts";
 import { CHART_COLORS, truncName } from "~/components/util";
 
+export enum ResultsBarChartType {
```
ArendPeter, line 4: nitpick: we've usually use string literals instead of enums, and then using typescript to force the list of strings. More details at https://devsparks.goooseman.dev/hacks/20240715-use-string-literals-not-enums/

That said, I don't think you need the resultsBarChartType at all here. This change will be useful in a variety of contexts in addition to equal preferences. For example, we saw the same issue occur on the scoring chart in 
https://github.com/Equal-Vote/star-server/issues/741

ArendPeter (commented):
Looks great, thank you for the fix!

I think this fix will be useful more generally than equal preferences so it can opened up more

packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
recursivesquircle: Okay I will remove the type.  I take it that assumes that any other bar chart that takes all zeroes is going to look right with that margin added?  And I will take note of the string literal over enum preference that you have

packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
```diff
@@ -2,22 +2,46 @@ import { useState } from "react";
 import { Bar, Cell, ComposedChart, LabelList, Legend, Line, ResponsiveContainer, XAxis, YAxis } from "recharts";
 import { CHART_COLORS, formatPercent, truncName } from "~/components/util";
 
-export default ({
+
+interface ResultsBarChartData {
```
ArendPeter, line 6: Nice work adding the types here!

ArendPeter (approved):
Looks great, approved!

packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
ArendPeter: Yep, I think they should all look good with the margin logic

ArendPeter:
The build error is because VoterErrorStats doesn't use the format of the new typescript interface.

I pushed a change so that the convention complies, so one more rebase should get the build passing on this PR

https://github.com/Equal-Vote/star-server/commit/80545265c9a6c43290d2cc00771708af39a8740c

PR #794, opened 2025-02-17
Update pl.yaml - 32 lines

masiarek:
the first 32 lines

## Related Issues

### DIFF
--- a/packages/frontend/src/i18n/pl.yaml
+++ b/packages/frontend/src/i18n/pl.yaml
@@ -3,32 +3,33 @@ election_home:
   # start_time example: https://bettervoting.com/tcpwth
   start_time: '{{capital_election}} Wybory zaczynają się {{datetime, datetime}}'
   end_time: '{{capital_election}} Wybory kończą się {{datetime, datetime}}'
-  ended_time: '{{capital_election}} ended on {{datetime, datetime}}'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/i18n/pl.yaml
```diff
@@ -3,32 +3,33 @@ election_home:
...
 # Ballot (ex. https://bettervoting.com/j87vqp/vote)
 ballot:
   this_election_uses: This {{election}} will use {{voting_method}} to elect {{spelled_count}} $t(winner.winner).
-  instructions_checkbox: I have read the instructions
-  learn_more: Learn more about {{voting_method}}
-  previous: Previous
-  next: Next
-  submit_ballot: Submit
-  submitting: Submitting...
+  Te {{election}} wykorzystują metodę głosowania {{voting_method}}, aby wybrać {{spelled_count}} $t(winner.winner)
```
ArendPeter, line 17: I think this link needs to be applied above after the "this_election_uses"

ArendPeter: You should be able to fix this by going into the "Files Changed" tab and fixing the entry

image

ArendPeter (approved):
Looks great, thank you for making the fix!

PR #803, opened 2025-02-20
Result of npm audit.

waugh:
`npm audit fix` made some changes but did not claim to have resolved all warnings. It said that for some packages, no fixes were available yet.


ArendPeter (approved):
Thank you!

Issue #801, opened 2025-02-19
BetterVoting - mission statement


masiarek:
## BetterVoting - mission statement

Fix two typos on the main page - https://bettervoting.com/
As-Is:
BetterVoting is here to help make your dream election **acessible**

Should-Be:
BetterVoting is here to help make your dream election **accessible**

## BetterVoting Features
Features - As-Is:
Setup multi-race elections for when there are **mutiple** issues or positions to vote on.

Features - Should-Be:
Setup multi-race elections for when there are **multiple** issues or positions to vote on.

Discussion document:
- https://docs.google.com/document/d/1n_oNrRzFrg7uPRtoYygPzUclt0IvrBu-hvJBRFgziFc/edit?tab=t.0

ArendPeter:
PR is merged so I'll close this one

PR #804, opened 2025-02-20
Typos 'acessible' and 'mutiple'

waugh:
Correct typos 'acessible' and 'mutiple' that had been showing up in the page footer. Corrects #801.


ArendPeter (approved):
Good catch!

PR #816, opened 2025-02-24
Reset stepper in racial dialog every time it is closed

### LINKED ISSUE #729: Clear race dialog when closed

waugh:
## Before

Let's say you go to add a race to an election you are working on, and let's say you go through the stepper to set the election type. Then you commit that change to add the race for real. Then say you go to add another race. The dialog would open with the stepper still on the second step.

## After

When you open the racial dialog for the second time, the stepper is reset to the first step.

I don't want to shoot my screen.

## Related Issues

Fixes #729 .

### DIFF
--- a/packages/frontend/src/components/ElectionForm/Races/AddRace.tsx
+++ b/packages/frontend/src/components/ElectionForm/Races/AddRace.tsx
@@ -13,6 +13,9 @@ export default function AddRace() {
     const handleOpen = () => setOpen(true);
     const handleClose = () => setOpen(false);
 
+    const [activeStep, setActiveStep] = useState(0);
+    const resetStep = () => setActiveStep(0);
+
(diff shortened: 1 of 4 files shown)


ArendPeter (approved):
Looks great! Thank you for the fix.

ArendPeter:
fyi, I edited your description from "Completed #729" to "Fixes #729" that way github will detect it and automatically close the issue after the merge

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Issue #802, opened 2025-02-19
Distribution of Equal Preferences - incorrect display


masiarek:
View / report - **Distribution of Equal Preferences** - incorrect display

- https://bettervoting.com/qdxyf4/results
- video: https://youtu.be/acu3PIQ4GfE

Maybe we should be dumping the data as a text file - without worrying about formatting in the browser? Potential solution - create a link to Results (CSV, tabulation, **Distribution of Equal Preferences** - text files).
- https://bettervoting.com/qdxyf4/results/text_files

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/d2a29865-a6fb-46f8-a6ab-e0cab2dff990" />

More info:
- https://docs.google.com/document/d/1VMLteQZnWTjDILj-796nEGfOV27Q_ErpJ8_RvkpWQ-8/edit?tab=t.0

ArendPeter:
Can you check this again? It should have been fixed recently
https://github.com/Equal-Vote/star-server/pull/793

mickaelfo:
Fixed.

(https://bettervoting.com/rbw9ck/results)
![Image](https://github.com/user-attachments/assets/4ab93969-ae1e-4397-9d95-e64d284aeb10)

masiarek:
Yes, the display issue has been resolved- but now we have another issue (do you want me to open a new ticket?) - we have incorrect data.
We have two "equal votes" - yet distribution shows zeros everywhere.

<img width="1076" alt="Image" src="https://github.com/user-attachments/assets/3eef0e85-d6c2-4d1a-a141-f8ac89f3a1fa" />

mickaelfo:
How strange, this election is completely bugged. In the Automatic Runoff Round tab, the contest is between Wang and Gus Grumpy, but in the Runoff Table, the contest is between Wang and Don Defiant. In the Tabulation Steps tab, the contest goes back to being between Wang and Gus. @ArendPeter @masiarek 

Wang vs Gus Grumpy
![Image](https://github.com/user-attachments/assets/b6d8d8d9-f693-4b6b-9bf2-a487d573d20f)

Wang vs Don Defiant
![Image](https://github.com/user-attachments/assets/8d7acddd-ad8e-4042-a388-f8eb72b921e7)

Wang vs Gus Grumpy
![Image](https://github.com/user-attachments/assets/8cae1f62-ff9b-4ca5-a7a2-0bab16757b2f)

mikefranze:
[...]

PR #829, opened 2025-03-01
Fixing Five STAR Tiebreaker Bug

### LINKED ISSUE #828: Multiple errors in tabulation steps
https://bettervoting.com/ffrp4f/results

Image

mikefranze:
## Description
Bug was introduced where once a tiebreaker got to the five star section, it would use all candidates and not just those remaining in the tiebreaker. 

This is demonstrated in #828 where two candidates are in a score tiebreaker for second place and a candidate who was already a finalist is selected again because they have the most 5 stars. So, both candidates in the runoff were the same candidate which again caused the tiebreaker to result in picking the candidates (not from the finalist pool) with the two highest 5 stars. This resulted in the runner up being not only not a finalist but one having the lowest score as well but happened to have the second most 5 stars.

Fix was to add back a function that would sort the candidates by 5 stars but filter out candidates not in the tiebreaker. I added a new test that would fail but now passes. 

I also changed the log language when entering the score tiebreaker. Before it would include all candidates and then narrow it down by those tied. This just starts with the tied candidates and lists the tied score.

Before:
image
image

After:
image
Logs aren't displayed in sandbox mode but here's the console logged results
image

Another election example 
Before:
image
image

## Related Issues

Fixes #828 
Relevant to #802 but doesn't fully resolve


ArendPeter (approved):
Thank you for fixing this! I'm guessing I broke this during the refactor, so thank you for addressing it, and adding a test case

Issue #731, opened 2025-01-14
Update bloc voting method races to use numbered pages


ArendPeter:
Currently our bloc voting results use tabs to represent the multiple rounds of counting.

This can be achieved by refactoring the WinnerResultTabs component to use the Pagination component (see PRResultsViewer as an example)

Currently WinnerResultTabs is only used for STAR, but the new pagination component should also be applied for bloc-RCV elections.

waugh:
In the context of our app, "bloc" and "basic multiwinner" are used to mean the same thing.

masiarek:
Bloc STAR Voting or Basic Multi-Winner STAR Voting: https://docs.google.com/document/d/1XsfRVA6zdHFRAkRCMyxBTK18WIDq03vSRKpJLnrHxC4/edit?tab=t.0

masiarek:
Test case: https://docs.google.com/document/d/13zcXHjJDaemkR2iA89X8brgsoC7wKOrdMfdcpLTjUf4/edit?tab=t.0

waugh:
@ArendPeter, I'm not sure what has to be shown for the rounds of non-PR multiwinner IRV, and in what form it should be shown.

masiarek:
Test case (Key Challenges Facing USA - 2025) with 10 seats: https://bettervoting.com/xmvfcc/

waugh:
> Test case (Key Challenges Facing USA - 2025) with 10 seats: https://bettervoting.com/xmvfcc/

@masiarek, toward what purpose do you cite that?

ArendPeter:
@waugh It would be similar to bloc-STAR. The 2 top-level STAR widgets (score & automatic runoff) would be replaced with the 2 top-level widgets from the IRV results. 

For now, any widgets in the race details that are specific to a certain bloc-tabulation round can be removed for the bloc results. For example, the "Ranked Choice Tabulation Rounds" table can be removed for bloc-IRV results. The bloc-STAR Voting results also removed most of the detailed widgets. we can have a future discussion item for figuring out how these widgets would work with bloc elections.

![Image](https://github.com/user-attachments/assets/7cfad330-b8f7-438a-9d10-0ee95fac56e3)

ArendPeter:
> Bloc STAR Voting or Basic Multi-Winner STAR Voting: https://docs.google.com/document/d/1XsfRVA6zdHFRAkRCMyxBTK18WIDq03vSRKpJLnrHxC4/edit?tab=t.0

[...]

PR #840, opened 2025-03-08
Pages 0

waugh:
## Description

Shows results from basic (non-PR) multiwinner STAR with page numbers on the bottom instead of tabs at the top.

## Before

image

## NAFTA

image

## Related Issues

Partial solution to #731, only addressing the basic multiwinner STAR aspect.


packages/frontend/src/components/Election/Results/Results.tsx
```diff
@@ -507,6 +509,8 @@ export default function Results({ race, results }: {race: Race, results: Electio
     .map((s,i) => ([<React.Fragment key={i*2}>{s}</React.Fragment>, <React.Fragment key={i*2+1}>&nbsp;</React.Fragment>]))
     .flat()
 
+  /* console.log('votingMethod', results.votingMethod); */
```
ArendPeter, line 512: not a blocker, but for future PRs can you remove the commented console log messages?

packages/frontend/src/components/Election/Results/Pages.tsx
```diff
@@ -0,0 +1,24 @@
+import { Pagination } from "@mui/material";
+import Widget from "./components/Widget";
+import WidgetContainer from "./components/WidgetContainer";
+
+const Pages = ({
+  children, pageCount, title, page, setPage /* one-based */
```
ArendPeter, line 6: Good comment! This has tripped me up before when working on this

I also like that you've Pages abstracted for reuse in other areas

packages/frontend/src/components/Election/Results/WinnerResultPages.tsx
```diff
@@ -0,0 +1,34 @@
...
+    console.log("`children` wasn't an array."); */
+  
+  const childArray = React.Children.toArray(children); // Ensure it's an array
+
+  return (
+    < Pages
+      pageCount={numWinners} page={page} setPage={setPage} title={false}
+    >
+      {childArray.map((child, i) => (
+        <div
```
ArendPeter, line 21: also not a blocker,  but we usually use Box instead of div in order to follow the MUI convention. This allows us to use other MUI features like the sx propery, but for this case I think the result will be the same.

ArendPeter (approved):
Looks great! I left a few nitpicky comments, but nothings a blocker so I'll go ahead an merge.

Thank you for the fix!

packages/frontend/src/components/Election/Results/WinnerResultPages.tsx
waugh: Planning to correct this in my next PR.

Issue #733, opened 2025-01-14
Replace all UUIDs with shorter IDs


ArendPeter:
We should stop using so many UUIDs. They increase request sizes, clutter UI, and make it more difficult when users need to manually enter IDs. 

I recommend reusing the [makeID function](https://github.com/Equal-Vote/star-server/blob/3bec66341447ec6416b170725a6d140ee881745f/packages/backend/src/Util.ts#L31), and add a prefix. So ballots can be prefixed with b- (ex. b-12345678), candidates with c-, and voters with v-.

This might require moving the function to the shared package so that it can be consumed by both frontend and backend, but it's so short that I would also be fine if we copied it in both places. 

The election ids have already been shortened, so here's the ones that can be shortened next

 - Ballots IDs : b-12345678
 - Voter IDs : v-12345678
 - Candidate IDs: c-123
 - Race IDs: r-123

Ballot and voter ids should be longer for security. Voter ids are used to access ballots so they shouldn't be brute force guessable. 

Race and candidate ids can be very short since they only need to be unique relative to that election.

NOTE: This change does not need to be retroactive. It's simpler and safer just have this apply to all new IDs moving forward

ArendPeter:
(copying from email)

From Adam
> I always hated these long IDs myself- but I thing they served some security purpose. 
In any case. It will be great to shorten this stuff in some places for sure!

From Arend
> For race and candidate IDs, I don't think there's security concerns as these IDs are publicly accessible via the API.
>
> Voter IDs and ballot IDs are more sensitive since they can be used to authenticate a voter to vote. I think the most important security factor is that these IDs are unpredictable.
> 
> If the IDs were generated in a predictable way (voter 1, voter 2, voter 3, etc), then it would be easy for someone to guess the voter ID and cast fraudulent votes.  
> 
> Since the IDs will still be random, I think it's secure. And we can also always add more digits if we're worried about guessing.

PR #841, opened 2025-03-09
WIP: Write in candidates

mikefranze:
## Description
Work in progress, do not merge! Just wanted to get some eyes on it.

This adds write in fields to the domain models, adds routes for getting the write in names and setting the processed write in names, and updates to the results processing to check for write ins in the ballots and assign the scores to the correct candidates.

In the race object, enable_write_in has been added which will be a setting in the race menu. This enables the write in field on the ballot and allows the backend to include them for tabulation.

On the score object in the ballot, a write in name has been added.

A new route getWriteIns provides the admin with a list of all write in names and their number of occurrences for each race. The admin then processes the names, groups any similar or misspelled names together, and approves or rejects them the candidates.

Admin then uses a new route setWriteInResults to save the processed write in data to the election. 

While processing the results, if it comes across a vote with an invalid candidate ID but with a write in name, it checks if that write in name is an alias of an approved write in candidate and maps the scores. This required updating the code to check all of the candidate IDs of the votes and make sure they map to the candidate IDs in the race.


packages/shared/src/domain_model/WriteIn.ts
```diff
@@ -0,0 +1,12 @@
+import { Uid } from "./Uid";
```
ArendPeter, line 1: Note that @recursivesquircle is working on shorter IDs at the moment. 
https://github.com/Equal-Vote/star-server/issues/733

Depending on when this is merged you may need to rebase so that the race id uses the new format

packages/backend/src/Controllers/Election/getElectionResultsController.ts
```diff
@@ -31,17 +31,37 @@ const getElectionResults = async (req: IElectionRequest, res: Response, next: Ne
     const election = req.election
     let results: ElectionResults[] = []
     for (let race_index = 0; race_index < election.races.length; race_index++) {
-        const candidateNames = election.races[race_index].candidates.map((Candidate: any) => (Candidate.candidate_name))
-        const race_id = election.races[race_index].race_id
+        const race = election.races[race_index]
+        const candidateNames = race.candidates.map((Candidate: any) => (Candidate.candidate_name))
+        const candidateIDs = race.candidates.map((Candidate: any) => (Candidate.candidate_id))
+        const useWriteIns = race.enable_write_in && race.write_in_candidates && race.write_in_candidates.length>0
+        const writeInCandidates = useWriteIns && race.write_in_candidates ? race.write_in_candidates : []
```
ArendPeter, line 38: nit, I think you can avoid the terinary expression by doing ```race.write_in_candidates ?? []``

packages/backend/src/Controllers/Election/getElectionResultsController.ts
```diff
@@ -31,17 +31,37 @@ const getElectionResults = async (req: IElectionRequest, res: Response, next: Ne
...
+                    const candidateIndex = candidateIDs.indexOf(score.candidate_id)
+                    if (candidateIndex >= 0) {
+                        row[candidateIndex] = score.score
+                    }
+                    else if (useWriteIns && score.write_in_name) {
+                        const write_in_name = score.write_in_name // typescript sees score.write_in_name as possibly undefined unless I extract it to another variable for some reason
+                        const writeInCandidateIndex = writeInCandidates.findIndex(candidate => candidate.aliases.includes(write_in_name) && candidate.approved)
+                        if (writeInCandidateIndex >= 0) {
+                            row[writeInCandidateIndex + candidateNames.length] = score.score
+                        }
```
ArendPeter, line 61: Should we add an else clause to throw (or log) a warning?

Feels like the writeInCandidateIndex is expected to always resolve

packages/backend/src/Controllers/Election/setWriteInResultsController.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    const race_index = election.races.map(race => race.race_id).indexOf(write_in_results.race_d);
+    if (race_index != -1) {
+        throw new BadRequest('Invalid Race ID')
+    }
+    if (!election.races[race_index].enable_write_in) {
+        throw new BadRequest('Write-In not enabled for this race')
+    }
+    election.races[race_index].write_in_candidates = write_in_results.write_in_candidates
```
ArendPeter, line 29: Should we also update write_in_candidates during the cast vote flow to ensure that any new write ins are added as new candidates or alias?

ArendPeter (approved):
Mostly nitpicks, but overall it looks good!

packages/shared/src/domain_model/WriteIn.ts
mikefranze: Good to know and I'll keep an eye out, but this is just the Uid type which is just a string which I doubt will change.

packages/backend/src/Controllers/Election/getElectionResultsController.ts
mikefranze: I'll give that a shot but I may have tried that. Typescript wasn't very cooperative in this section.

packages/backend/src/Controllers/Election/getElectionResultsController.ts
mikefranze: It wouldn't in the case that the write in candidate wasn't approved by the admin. But maybe that should be a separate check? First check if they are an alias to a candidate, if not throw error and admin can revisit write in section. Then check if approved, then add score.

packages/backend/src/Controllers/Election/setWriteInResultsController.ts
mikefranze: I'd rather not update the election every time a vote is cast. I was thinking the write in processing would occur after all votes are entered. Or the admin can do processing as the election is running.

packages/backend/src/Controllers/Election/getElectionResultsController.ts
mikefranze: Actually, maybe it would be better to have a counter for unprocessed write ins. That way the results can still be retrieved live and display with the results that X write in names are still unprocessed.

mikefranze: Updated results to have an unprocessed write in candidate count

ArendPeter: Nice! Yeah I was just thinking that we should have some default write-in behavior in addition to allowing the admin to configure the aliases

Counting each new variation as it's own candidate is a good starting point, but I could also see us performing some minimal typo checks by default.

PR #846, opened 2025-03-11
Free height of part of race form

### LINKED ISSUE #842: Chrom description text missing (Plurality and RCV)
Peter from Florida reported this issue:

On [bettervoting.com](http://bettervoting.com/) when creating an race for an election in Google Chrome, some of the description text under Choose One Plurality is missing. Also Ranked Choice Voting option is missing.
However Opera browser did not have those issues. Differences shown in both images.

Image

Image

waugh:
## Description

Removes fixed height from part of race form, but at the cost of a height transition. The opacity transition remains.

## Before

image

## After

image

## Related Issues

fixes #842


ArendPeter (approved):
Thanks for the fix!

We'll see it how it looks without the height transition, and then we can iterate further if needed.

PR #854, opened 2025-03-16
Fixed Punctuation on About Page

Luluanaki:
## Description
Added a "-" and a "." on about page that were missing.

image

## Related Issues


ArendPeter (approved):
Thank you, nice work!

Issue #694, opened 2024-10-13
Featured election panel is sending continuous requests


ArendPeter:
The featured elections panel on the homepage is disabled since it was sending continuous requests to the backend. This needs to be investigated and then reenabled

ArendPeter:
@mikefranze I'm assigning this one to you since I think you already did some initial investigation

jwidan:
For context, here is what the issue looks like:

![Image](https://github.com/user-attachments/assets/5839d80f-3e37-45bc-9586-b4a1f316f5ae)

ArendPeter:
Thank you for the fix! I'll go ahead and close this issue

PR #855, opened 2025-03-16
fix: stop sending repeated or empty electionId requests

jwidan:
## Description
Guard against empty or malformed electionId values in REACT_APP_FEATURED_ELECTIONS that led to repeated or unnecessary backend requests.

No more errors seen.

Without featured elections:
image

With featured elections:
image

## Related Issues
fixes issue #694


ArendPeter (approved):
Looks great! Thank you for the fix

Issue #754, opened 2025-01-24
Ballots with mix of zeros and blanks should not count as abstention


ArendPeter:
> That said, there is a bug where all zeros currently count as an abstention. A vote should only count as an abstention if all rows were left blank 

 _Originally posted by @ArendPeter in [#740](https://github.com/Equal-Vote/star-server/issues/740#issuecomment-2611282633)_

ArendPeter:
To recreate this I created a poll with a single race, and then filled in a couple of zero bubbles

![Image](https://github.com/user-attachments/assets/6bf2e1c8-c1ca-4c14-a014-8ea215b7f8d1)

This receipt shows abstained, but instead it should show the contents of the ballot that's being submitted.

This can be resolved by fixing pageIsUnderVote function

https://github.com/Equal-Vote/star-server/blob/cf99b29d59fbc5262812ef5747df28d6bf566469/packages/frontend/src/components/Election/Voting/VotePage.tsx#L191

Luluanaki:
When I change it to only say Abstained when all inputs are Null, and there are some inputs that are 0 and others that are Null, it will return 0, 0, 0 anyways. Should I make Null inputs be returned as 'N/A'?

PR #857, opened 2025-03-17
Issue #754: Abstention Issue

Luluanaki:
Ballots with mix of zeros and blanks no longer count as abstention

## Description
Ballots with mix of zeros and blanks should not count as abstention. This fixes the immediate issue but also raises another issue of votes left blank should not be counted as '0'.

image
image

## Related Issues

### DIFF
--- a/packages/frontend/src/components/Election/Voting/VotePage.tsx
+++ b/packages/frontend/src/components/Election/Voting/VotePage.tsx
@@ -189,7 +189,7 @@ const VotePage = () => {
   }
 
   let pageIsUnderVote = (page) => {
-    return page.candidates.reduce((prev, c) => prev && (c.score == 0 || c.score == null), true)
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Election/Voting/VotePage.tsx
```diff
@@ -189,7 +189,7 @@ const VotePage = () => {
   }
 
   let pageIsUnderVote = (page) => {
-    return page.candidates.reduce((prev, c) => prev && (c.score == 0 || c.score == null), true)
+    return page.candidates.every(c => c.score == null);
```
ArendPeter, line 192: oh interesting

I don't use every() very often but this is a good use case!

ArendPeter (approved):
Nice work!

PR #877, opened 2025-03-28
Paginate IRV rounds.

### LINKED ISSUE #731: Update bloc voting method races to use numbered pages

waugh:
## Description

Paginate IRV rounds.

## Before

image

## After

image

## Related Issues

Fixes #731.


ArendPeter:
I was visualizing this differently. Here's a mockup

image

I think we're doing the same thing except your paginator has each of the elimination rounds also included. I can see how yours works too. 

The runBlocTabulator utility function that's used by Approval & STAR is another option for updating the backend, but that could be tricky considering that the IRV and STV tabulators are combined.

It also looks like this impacts how the single winner IRV results looks? But I would like that to be unimpacted by this change.

So all that said, I see 2 paths forward for this PR.

Option 1. Update Results.tsx to use the new approach when num_winners > 1, otherwise use the old approach.

OR

Option 2. Update the PR to work more similarly to my mock up. Using runBlocTabulator is optional.

So sorry for the confusion! Feel free to schedule a call via starvoting.org/meeting if that would be helpful.

mikefranze:
I would agree with Arend's mock up. Putting all the elimination rounds for each seat round in one pagination would be confusing. Just showing first and final counts for each seat round should be enough and we can show elimination round scores in the race details tables.

ArendPeter:
Let me know if this is ready for another review

And if so, can you post updated screenshots in the description?

waugh:
Yeah, it's not ready yet.

waugh:
@ArendPeter, I am trying to understand the correctness and rationale of the current code as it applies to a single-winner case. Consider for example I have three candidates, called Jack, Gilbert and Ada. There are only two votes:

- 1: Jack > Ada > Gilbert
- 1: Jack > Gilbert > Ada

This is a landslide election in that the tally only generates a single round, in which Jack is found to win based on the quota. Here is how the results are shown:

image

Is this what we want to see in this case?

```
  const firstRoundData = results.voteCounts[0].map((c,i) => ({
    name: results.summaryData.candidates[i].name, votes: c
  }));

  const runoffData = results.voteCounts.slice(-1)[0]
    .map((c,i) => ({name: results.summaryData.candidates[i].name, votes: c}))
    .sort((a, b) => b.votes - a.votes)
    .filter(a => a.votes != 0) // filter out eliminated candidates
    .concat([{
      name: t('results.rcv.exhausted'),
      votes: results.exhaustedVoteCounts.slice(-1)[0]
    }])
```

We can see in this code that the `runoffData` calculation is assuming that Ada and Gilbert have been eliminated (they haven't), on the grounds that they have zero Hare "votes".

ArendPeter:
> @ArendPeter, I am trying to understand the correctness and rationale of the current code as it applies to a single-winner case. Consider for example I have three candidates, called Jack, Gilbert and Ada. There are only two votes:
> 
> * 1: Jack > Ada > Gilbert
> * 1: Jack > Gilbert > Ada
> 
> This is a landslide election in that the tally only generates a single round, in which Jack is found to win based on the quota. Here is how the results are shown:
> 
> image
> 
> Is this what we want to see in this case?
> 
> ```
>   const firstRoundData = results.voteCounts[0].map((c,i) => ({
>     name: results.summaryData.candidates[i].name, votes: c
>   }));
> 
>   const runoffData = results.voteCounts.slice(-1)[0]
>     .map((c,i) => ({name: results.summaryData.candidates[i].name, votes: c}))
>     .sort((a, b) => b.votes - a.votes)
>     .filter(a => a.votes != 0) // filter out eliminated candidates
>     .concat([{
>       name: t('results.rcv.exhausted'),
>       votes: results.exhaustedVoteCounts.slice(-1)[0]
>     }])
> ```
> 
> We can see in this code that the `runoffData` calculation is assuming that Ada and Gilbert have been eliminated (they haven't), on the grounds that they have zero-count of Hare "votes".

Good point! The current behavior is fine as a quick solution, I agree it would be better to generate the runoffData based on which candidates were actually eliminated during the tabulation rounds.

waugh:
In the above example, there is only one round. There isn't really a runoff. Should we continue to show two charts?

waugh:
Here is an example showing a single-winner election looking similar to how it looked before. Even though this case finds the winner on the first round, I still display two charts. However, the candidates who got zero votes are no longer suppressed from the bottom chart. This is appropriate since they were not eliminated in prior rounds.
image

waugh:
Here is a multiwinner non-PR example showing pagination. I have not achieved yet, the side-by-side presentation that you call for in your mock-up. I'm asking you to review the content before I try that. Also, I am showing a human-language translation as TODO. Please review the concept before I ask Adam to add the phrase to the book of translations.

image
image
image

Next page:

image

ArendPeter:
> Here is an example showing a single-winner election looking similar to how it looked before. Even though this case finds the winner on the first round, I still display two charts. However, the candidates who got zero votes are no longer suppressed from the bottom chart. This is appropriate since they were not eliminated in prior rounds, since there haven't been any prior rounds. image

Looks good, nice work!

packages/frontend/src/components/Election/Results/IRV/ifc.ts
```diff
@@ -0,0 +1,15 @@
...
+import {
+  irvRoundResults
+} from "@equal-vote/star-vote-shared/domain_model/ITabulators";
+
+export interface irvContext {
+  candidatesByIndex: {name: string}[],
+  t: Function /* international translator */
+}
+
+export interface irvWinnerSearch {
```
ArendPeter, line 12: Nice use of ts interfaces!

packages/frontend/src/components/Election/Results/IRV/winner.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return round.standings.map(
+    ({ candidateIndex, hareScore }) =>
+    ({ name: candidatesByIndex[candidateIndex].name, votes: hareScore })
+  )
+};
+
+export function IRVWinnerView ( {win, context}:{
+  win: irvWinnerSearch, context: irvContext
+}) {
+  const {t} = context;
```
ArendPeter, line 31: nit

Instead of passing ``t`` through the functions using irvContext. You can also use ``const {t} = useRace();`` everywhere. The useRace context provider was added the the project in order to minimize how many variables we need to pass between functions.

https://react.dev/learn/passing-data-deeply-with-context

ArendPeter (approved):
Looks good!

ArendPeter:
Merging for now to avoid conflicts, I'll make a separate task for the remaining action items.

PR #899, opened 2025-04-08
fixed environment variables for web service in docker compose and removed keycloak configuration step

csamuele:

## Related Issues


ArendPeter (approved):
Thanks! Looks good

PR #923, opened 2025-04-18
End to end testing

csamuele:
## Description
I updated unlabeled components to use aria label and used more descriptive labels. I'm not sure why so many components have the name attribute, it seemed a bit redundant, but I tried to change just the ones I've been working on


ArendPeter (approved):
Looks great!

PR #924, opened 2025-04-18
add eslint and use clean up code with recommended settings

csamuele:
## Description
Added linting to clean up some of the errors in the console and have a bit more type safety


ArendPeter (approved):
Thank you!

PR #934, opened 2025-05-17
Update es.yaml

LukasHeaton:
Translated more lines


ArendPeter (approved):
Thanks for translating these!

PR #936, opened 2025-05-22
Update Spanish translations for ballot submission and voting instructions.

pablohernandezb:
## Description

I updated the es.yaml file to have a full Spanish translation. I didn't translate "Ranked Robin" because there is not a direct translation other than "Todos contra todos con orden de preferencia" or "Votación por orden de preferencia en formato todos contra todos" which is basically explaning that all candidates are paired in one-on-one match-ups. I left it as "Ranked Robin" for the time being. In baseball, and sports, we use in "Round Robin" in English to signal that they are 1 on 1 match-ups so I believe most of Spanish speakers will get the idea.

### DIFF
--- a/packages/frontend/src/i18n/es.yaml
+++ b/packages/frontend/src/i18n/es.yaml
@@ -10,7 +10,7 @@ election_home:
   ended_time: '{{capital_election}} finalizó el {{datetime, datetime}}'
   vote: Votar
   or_view_results: o ver los resultados
-  ballot_submitted: '{{capital_ballot}} Entregada'
+  ballot_submitted: '{{capital_ballot}} Enviada'
   view_results: Ver los resultados
   drafted: Esta elección está en fase de elaboración
   archived: Esta elección se ha archivado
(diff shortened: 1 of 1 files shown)


LukasHeaton:
Wow, thank you so much for all you did here! 👏 Amazing work.

Here are my suggestions for edits:

**Line 34** I believe the meaning of "Candidates left blank are ranked last" may have been misinterpreted. Rather than "last" in the sense of timing, here "last" is referring to least favorite, so maybe something like "Los candidatos que se dejen en blanco se clasificarán en último lugar"

**Lines 49, 50, 63 and 64** Change "Dale" to "Dele" and "tu(s)/tu" to "su(s)/su" to keep with formal "you" used everywhere else

**Line 56** Change "de" to "del"

**Line 68** Change "Tu" to "Su" and "Prefieras" to "Prefiera"

**Lines 68 and 69** Looks like extra text may have gotten copied- the second sentence in line 68 should be on line 69, and line 69 doesn't belong there

**Lines 78, 79, 89, 90 and 97** I'm not 100% sure about this, but does there need to be an "a" before "candidatos", i.e. "Ordene a los candidatos"?

**Line 177** Remove "to support"

**Line 208** should be "ganan" instead of "ganó" because it's referring to multiple candidates winning and is in present tense

**Line 252** I would recommend "Sumar" instead of "Suma" because it seems to be explaining how the voting method works. Or, if it's a literal command, change it to "Sume" to keep with the formal "you"

**Line 280** I think it should be "más preferidos" instead of just "preferidos", but could be wrong on this

pablohernandezb:
Awesome, thank you so much, @LukasHeaton! I made all the changes based on your comments. I put the more informal language because that's the route we took when crafting the materials for Votación Estrella but the formal it's ok too. Looking forward to keep contributing to the project.

ArendPeter:
Thank you both for your help. Looks good to me!

LukasHeaton:
@pablohernandezb Totally! I think either formal or informal "you" is fine, just wanted to make sure it was consistent across all our sites and materials, which mostly use formal. But we could consider changing everything to tú if you think that makes more sense

PR #947, opened 2025-06-06
replace uuids with shorter ids

### LINKED ISSUE #733: Replace all UUIDs with shorter IDs

recursivesquircle:
## Description
This PR introduces a new ID generation utility that replaces the previous UUID-based system with shorter, more user-friendly IDs. 

## Related Issues

``fixes #733 `` to close issue #733

### DIFF
--- a/packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
+++ b/packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
@@ -6,7 +6,7 @@ import useElection from '../../ElectionContextProvider';
 import { Race as iRace } from '@equal-vote/star-vote-shared/domain_model/Race';
 import structuredClone from '@ungap/structured-clone';
 import useConfirm from '../../ConfirmationDialogProvider';
-import { v4 as uuidv4 } from 'uuid';
+import { makeID, ID_PREFIXES, ID_LENGTHS } from '@equal-vote/star-vote-shared/utils/makeID';
 import { Candidate } from '@equal-vote/star-vote-shared/domain_model/Candidate';
 import { useDeleteAllBallots } from '~/hooks/useAPI';
 import useSnackbar from '~/components/SnackbarContext';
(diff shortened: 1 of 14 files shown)


packages/frontend/src/components/ElectionForm/Races/useEditRace.tsx
```diff
@@ -30,7 +30,7 @@ export const useEditRace = (race: iRace | null, race_index: number) => {
         voting_method: 'STAR',
         candidates: [
             { 
-                candidate_id: uuidv4(),
+                candidate_id: makeID(ID_PREFIXES.CANDIDATE, ID_LENGTHS.CANDIDATE),
```
ArendPeter, line 33: Seems like the frontend uses makeID instead of makeUniqueID, but I think these would also need to be unique?

ArendPeter (commented):
Thanks for adding this, looks great! Just added a question for clarification.

PR #948, opened 2025-06-07
Fix: Open external link in new tab (Issue #937)

### LINKED ISSUE #937: Set all links to open in a new tab
Image

When visitors click on a link from our page, they lose where they are because it replaces the current page rather than opening in a new tab

Priyanshu7798:
## Description
This PR fixes the issue where the Link in footer and About page open in same tab.

Video of footer

https://www.loom.com/share/b6eaffdedb6646a8a9ce1365681d4a02?sid=3dc07305-01a1-4e8d-82e6-bd7f586b732b Be sure to include both desktop and mobile views (mobile could be as low as 320 px wide) 

🔗 Related Issue
Closes #937


ArendPeter (approved):

Issue #814, opened 2025-02-22
Election title and race title - improve consistency


masiarek:
Low priority consistency fix - labels are not consistent :

Discussion document: 
- https://docs.google.com/document/d/1NMmWDf0TppA83AxEVlTQDXm9SI5j77bffZEeslS8zCY/edit?tab=t.0

Compare Race and Election:

<img width="882" alt="Image" src="https://github.com/user-attachments/assets/c2ab0db4-ba03-43b1-9982-d975aa96c9b3" />

## Election
Election - As-Is:
- Election Title *
- Description

Election - Should-Be:
- Election Title
- Election Description

## Race
Race - As-Is
- Title
- Description

Race - Should-Be

- Race (Contest) Title - Position or Question Type
- Race (Contest) Description

<img width="743" alt="Image" src="https://github.com/user-attachments/assets/d0b9d948-c792-4aca-9ef5-9ef50cfffcfa" />

<img width="972" alt="Image" src="https://github.com/user-attachments/assets/8e2a66c1-579d-42ab-a7fa-1cb574a0dc5b" />

Test cases:
- BV340 - Change Election Title and Description
- BV345 - Change Race Title and Description (one race)
- BV350 - Change Race Title and Description (multiple races)

ArendPeter:
Good point. I like "Race Title" and "Race Description"

PR #958, opened 2025-07-01
fix:consistency rank and election #814

Priyanshu7798:
## Description

I changed the heading as per required in the in the issue

## Related Issues

> fixes issue #814


ArendPeter (approved):
Thanks Priyanshu!

PR #964, opened 2025-07-22
Update wording for single-candidate supporters

numbersguy104:
## Description
Make the wording clearer in the results screen, showing the percentage of voters that only supported one candidate.

Before: "25% of Kamala Harris supporters only voted for one candidate"
After: "25% of Kamala Harris supporters did not vote for any other candidate"


ArendPeter (approved):
Looks great, thanks!

Issue #745, opened 2025-01-21
“Continue” button  and “Election title"


masiarek:
We start by clicking a radio button picking between “election” and “poll”. 

Example - we click on “election”, and then we click “Continue” button to activate the next box: “Election title”. 

<img width="573" alt="Image" src="https://github.com/user-attachments/assets/6217acc2-0302-40cd-b01e-743dd38d1441" />

My proposal - remove the Continue button - once user clicks on one of the radio buttons - jump to the "Election title" automatically.

User should be able to change the type (poll or election) in the window "Admin Page" for Election.

Open questions:
- what if user wants to change the type from poll to election (or vice versa)?
- do we have functionality to change the type (election vs poll)
- should the change be allowable? and until what stage in the process? I guess once election is Finalized (in progress / running) - the change should not be possible

ArendPeter:
I agree, this will improve the experience. And people can still use the back button on the next steps if they change their mind.

I also agree that this should be added to the extra settings (I thought it was already there but I guess not)

- [ ] Add poll vs election option to extra settings panel

PR #971, opened 2025-07-28
Fixing #745 I eliminated the "continue" button

pablohernandezb:
## Description

Issue #745 asked to eliminate the "continue" button on the first step when creating a New Election. @ArendPeter also suggested that I added like a quarter of a second to make the transition to next step smoother.

Screenshot 2025-07-28 180851
Screenshot 2025-07-28 180804


ArendPeter:
Can you update the PR to apply the same pattern to step 3?

pablohernandezb:
> Can you update the PR to apply the same pattern to step 3?

I don't know if this is possible because if you click "Yes" then you'll have to type the email, not go straight to the last step, right? I think it will be odd to add the same pattern as step 1.

PR #974, opened 2025-08-08
Add close/open election functionality

### LINKED ISSUE #812: Add button to admin view for manually closing an election
Currently elections will close if there's an end time set and the election passes the end time, but there's no way to close the election if the end time isn't set.

The admin should have a toggle button for closing the election. And then untoggling it will move the election back to open.

For now this feature should only be available for elections where the start and end time isn't set. That way we don't have to worry about the logic of the two systems conflicting. 

In the future there should be more protections for how often the admin can do this (see break glass auditing flow https://github.com/Equal-Vote/star-server/issues/317 ), but for now I think it's fine to let the admin freely toggle it on and off

numbersguy104:
(Note: This is my first backend change and could probably use a second set of eyes, especially for the API additions.)

## Description
Allows election admins to close and open an election using a new button in the admin home page.
The button will not show up if the election has a start/end time set, to avoid issues with conflicting systems.

## Related Issues
Should close #812

### DIFF
--- a/packages/backend/src/Controllers/Election/index.ts
+++ b/packages/backend/src/Controllers/Election/index.ts
@@ -9,5 +9,6 @@ export * from './getElectionResultsController';
 export * from './getElectionsController';
 export * from './sandboxController';
 export * from './sendInvitesController';
+export * from './setOpenStateController';
 export * from './setPublicResultsController';
 export * from './sendEmailController';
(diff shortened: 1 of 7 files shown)


ArendPeter (commented):
(ugh, old comment I forgot to publish)

Issue #730, opened 2025-01-14
Add support for write-in candidates


ArendPeter:

masiarek:
Write-in discussion document:
- https://docs.google.com/document/d/1zzHHTovOnpnLxzjMSqHHzLPsZzvqraSQk1IJorRTiA4/edit?tab=t.0

What kind of test cases do we need?

jacksonloper:
As mentioned in the goole doc above, grouping write-in candidates (e.g. if one person writes JFK and someone else writes John Kennedy) is a tricky part of any write-in support.

In my point of view at this moment, it would be nice to provide election admins a service that helps them with this grouping process.  In fact, I think (though could easily be persuaded) that it would be worth it to help them with auto-grouping even if the admins start to group write-ins before polls close, and then want to auto-group unprocessed candidates later when more votes come in.  

If we want to provide that "full-service" to the election admins, we would need to decide on an algorithm for taking
* a pre-existing set of approved write-in candidates (e.g., "John Fitzgerald Kennedy, Jr.") and their aliases (e.g., "JFK") and
* a set of unprocessed write-in names (e.g. "John Kennedy")
and yielding a new best-effort set of write-in candidates with aliases.   #983 offers a proposal for the algorithm and a testable UI that you could try it to look for corner cases (building on existing name_match_testing from @mikefranze).  

Is this the direction we want to go?

PR #983, opened 2025-10-07
Feat/name match aggregation

jacksonloper:
## Description

Currently

https://bettervoting.com/name_match_testing/

exists at least in part in my understanding in order to help us test how write-ins could be grouped (e.g., if a voter writes JFK and another writes John F. Kennedy).  

This PR increases the scope of the grouping problem that the name_match_testing endpoint tests.

In particular, consider the following set of happenings.

1. Election starts
2. Votes come in
3. Admins auto-group
4. Admins twiddle a bit with the auto-grouping to get what they're happy with 
5. More votes come in with new write-ins
6. Admins need to auto-group some more

In this case, we have the "true" groupings that admins have settled upon (from step 4) and want to incorporate new writeins.  Its... complicated.  

We have two options (1) make auto-grouping completely delete anything that was already there (plausible if we enforce that admins can't start grouping write-ins until after polls close) or (2) actually figure out how to do this fancy auto-grouping-integrating-with-extant-aliases business.   In case we want to go ahead with (2), I have prepared a version of name_match_testing that implements this tricky business.

## Related Issues

Figuring out which way we want to go is a prereq to #730

### DIFF
--- a/packages/frontend/src/components/NameMatchingTester.tsx
+++ b/packages/frontend/src/components/NameMatchingTester.tsx
@@ -140,14 +140,6 @@ const computeHybridSimilarity = (name1: string, name2: string): number => {
     const initialScore = bothInitialized ? 0 : compareInitials(tokens1.initials, tokens2.initials);
 
     const bonus = wordScore > 0 && initialScore > 0 ? 0.1 : 0; // 10% bonus if both match
-    console.log(name1, name2)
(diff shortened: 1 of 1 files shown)


mikefranze (approved):
Looks good at first glance. Since this is just a testing page we can approve it so I can experiment with it more live.

Issue #811, opened 2025-02-22
Omit Date Submitted From the Admin Ballots Tab


ArendPeter:
![Image](https://github.com/user-attachments/assets/fe87ce1c-0cdc-4957-bd5d-42673514b9eb)

Date submitted can also be seen in the Voters tab so this information can be used to link the ballot to the voter

Tasks

- [x] At minimum this should be removed from the UI
- [ ] Ideally we should also remove it from the API. The implementer should check if it's possible to switch to using the getAnonymizedBallot API

BckupMuthu:
@ArendPeter :can you  pls close this

ArendPeter:
Thanks for the fix! Your item resolved the first half of the task list (and that was the most important), but I'll leave the issue open for the second half

masiarek:
- retest remove "submitted date" - passed: https://docs.google.com/document/d/1lEW_uR7EvzBQg2z-DI7iVvrlebTvZjA918T44sw2rmk/edit?tab=t.0
- test unauthorized view (passed): https://docs.google.com/document/d/1DOESlpKtGLiPjYvi-oxVQDt94JmI45rwjt0AWI6h5YU/edit?tab=t.0

ArendPeter:
Triage Updates
 - Tagged `Complexity: Medium`, `Role: Back End`, because fixing the back end api will likely require multiple files. 
 - Created an ER to #1109 for tracking Jackson's updates
 - Created https://github.com/Equal-Vote/bettervoting/issues/1111 to give newer member extra context for the item
 - Added dependency label
 - Set to milestone to tech debt. Technically this is a security issue, but it feels more like tech debt

PR #985, opened 2025-10-08
fix: scrub some potentially identifying information from an endpoint

jacksonloper:
## Description

The `/API/Election/[electionid]/ballots` endpoint currently reveals some information that could perhaps be used to let election admins figure out the ballot_id <-> voter_id link.  This scrubs the info before sending it out on the  endpoint.

I made some informal tests and it didn't immediately break things, but I wouldn't want to push this to main without some way of more clearly guaranteeing that I didn't break anything :)!  What is protocol for that?

## Related Issues

In service of #811 among other things.

### DIFF
--- a/packages/backend/src/Controllers/Ballot/getBallotsByElectionIDController.ts
+++ b/packages/backend/src/Controllers/Ballot/getBallotsByElectionIDController.ts
@@ -20,8 +20,20 @@ const getBallotsByElectionID = async (req: IElectionRequest, res: Response, next
         Logger.info(req, msg);
         throw new BadRequest(msg)
     }
+
+    // Scrub identifying information from ballots to preserve voter anonymity
+    const scrubbedBallots = ballots.map(ballot => ({
+        ...ballot,
+        history: undefined,
+        date_submitted: undefined,
+        create_date: undefined,
+        update_date: undefined,
+        user_id: undefined,
+        ip_hash: undefined
+    }));
+
     Logger.debug(req, "ballots = ", ballots);
-    res.json({ election: req.election, ballots: ballots })
+    res.json({ election: req.election, ballots: scrubbedBallots })
 }
 
 export {


packages/backend/src/Controllers/Ballot/getBallotsByElectionIDController.ts
```diff
@@ -20,8 +20,20 @@ const getBallotsByElectionID = async (req: IElectionRequest, res: Response, next
         Logger.info(req, msg);
         throw new BadRequest(msg)
     }
-    Logger.debug(req, "ballots = ", ballots);
-    res.json({ election: req.election, ballots: ballots })
+
+    // Scrub identifying information from ballots to preserve voter anonymity
+    const scrubbedBallots = ballots.map(ballot => ({
```
ArendPeter: This looks reasonable to me. This API is only used for the view ballot page and that page only cares about the ballot anyway

packages/backend/src/Controllers/Roll/getElectionRollController.ts
```diff
@@ -40,7 +47,15 @@ const getByVoterID = async (req: IElectionRequest, res: Response, next: NextFunc
         Logger.info(req, msg);
         throw new BadRequest(msg)
     }
-    res.json({ electionRollEntry: electionRollEntry })
+
+    // Scrub ballot_id to prevent linking voters to ballots
+    const scrubbedEntry = {
```
ArendPeter: This looks  safe too. I don't think the code path is actually used at the moment, since the viewing an individual roll is currently passed an entry from the full list

https://github.com/Equal-Vote/bettervoting/blob/409273b8f44e29e755d8c753cd292ce8d1c918ca/packages/frontend/src/components/Election/Admin/EditElectionRoll.tsx#L22

packages/backend/src/Controllers/Ballot/getBallotsByElectionIDController.ts
ArendPeter: Looks good, this one is only used on the ballot table and it doesn't need any of those fields https://github.com/Equal-Vote/bettervoting/blob/409273b8f44e29e755d8c753cd292ce8d1c918ca/packages/frontend/src/components/Election/Admin/ViewBallots.tsx#L89

ArendPeter (approved):
Looks good to me!

I'm pretty confident this won't break anything, so I think we can agree on a time to merge, and then we (you, me, Adam) can do a sanity check to make sure the following pages are working

 - Ballots Page
 - Individual Ballot Page
 - Voter Rolls Page
 - Individual Voter Roll Page
 
 Perhaps I'll merge Sunday morning 9 am PST | noon EST?

ArendPeter:
I've deployed it and did a sanity check on those 4 pages. Looks good! Adam will do further testing from here.

Issue #907, opened 2025-04-12
Preliminary Results and Response Submitted pages do not state what voting method was used


flynnwastaken:
<img width="1177" alt="Image" src="https://github.com/user-attachments/assets/fafdd2c1-9707-4f98-9f70-0776b7324ba5" />

<img width="1169" alt="Image" src="https://github.com/user-attachments/assets/d1b0248b-d476-4da6-8edb-b8084fa40ae6" />

This election was 2-winner STAR-PR I believe, but nowhere does it say that.

On the Preliminary Results page, there is duplication of the poll name, so instead of the first instance of that, it could say:
"Election run with **_voting_method_**" or something like that

On the Response Submitted page, it could say right underneath the poll name:
"A **_voting_method_** election" or something like that

masiarek:
Add Voting Method name below Race Name in the Results View: https://github.com/Equal-Vote/star-server/issues/879

ArendPeter:
We have it displayed in small font in the corner, but I agree it could be made more clear. I'll add it to polish milestone for now

Daifaizonaldridge:
I'm a new to this project but would like to give this a shot?

ArendPeter:
Susan and I were just discussing this in a meeting, and she'll update the voting method to be listed under the voter count

@Daifaizonaldridge I saw you were also interested in #874 so feel free to start with that one 🙂

crayne:
@ArendPeter  If there is more than one race, as I read the code, the method and the learning link are currently repeated in the lower right-hand corner, like this (if there are 3 races):


```
Voting Method: STAR Voting
How STAR Voting works (link)

Voting Method: Ranked Choice Voting
How Ranked Choice Voting works (link)

Voting Method: Ranked Robin Voting
How Ranked Robin Voting works (link)
```


Would we want all three of these to be moved to the place that we discussed?  Or do I have this wrong?

PR #1005, opened 2025-10-19
Fix/move voting method info

crayne:
## Description
Move voting method and learning link from lower right to just above results

## Related Issues

Fixes issue #907

### DIFF
--- a/packages/frontend/src/components/Election/Results/Results.tsx
+++ b/packages/frontend/src/components/Election/Results/Results.tsx
@@ -256,13 +256,6 @@ function ResultsViewer({ methodKey, children }:{methodKey: string, children:Reac
   return (
     <Box className="resultViewer">
       {children}
-       {/*
-      <Typography component="p" sx={{textAlign: 'right', color: '#808080', fontSize: '.8rem', marginTop: '20px'}}>
-        {t('results.method_context', {voting_method: votingMethod})}
-        {i18n.exists(learnLinkKey) && <><br/><a href={t(learnLinkKey)} style={{color: 'inherit'}}>{t('results.learn_link_text', {voting_method: votingMethod})}</a></>}
-      </Typography>
-      */}
-
     </Box>
   );
 }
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Election/Results/Results.tsx
```diff
@@ -258,10 +258,13 @@ function ResultsViewer({ methodKey, children }:{methodKey: string, children:Reac
   return (
     <Box className="resultViewer">
       {children}
+       {/*
       <Typography component="p" sx={{textAlign: 'right', color: '#808080', fontSize: '.8rem', marginTop: '20px'}}>
         {t('results.method_context', {voting_method: votingMethod})}
         {i18n.exists(learnLinkKey) && <><br/><a href={t(learnLinkKey)} style={{color: 'inherit'}}>{t('results.learn_link_text', {voting_method: votingMethod})}</a></>}
       </Typography>
+      */}
+
```
JonBlauvelt: can we just delete this?

packages/frontend/src/components/Election/Results/ViewElectionResults.tsx
```diff
@@ -43,7 +61,26 @@ const ViewElectionResults = () => {
             <Typography variant="h4" component="h4">
               {t("results.election_title", { title: election.title })}
             </Typography>
-            {isPending && <div> {t("results.loading_election")} </div>}
+
+
+            {/* Voting method and learning link */}
+              {/* Test of array of voting methods and learning links */}
+              {Array.from({ length: raceCount }).map((_, i) => (
+                  <Typography key={i} variant="h5" component="h5">
```
ArendPeter: Thanks for adding this! I would recommend moving the voting method text to the Results file instead https://github.com/Equal-Vote/bettervoting/blob/main/packages/frontend/src/components/Election/Results/Results.tsx#L488

I think that'll also be a little easier since you'll already have the specific race handy in that context

Also when it's updated, can you post a screenshot in the PR description?

ArendPeter (changes requested):

packages/frontend/src/components/Election/Results/ViewElectionResults.tsx
crayne: @ArendPeter I tried doing that, but I could not get the voting method and learning link to appear in the correct position, before the race results.  The data is retrieved asynchronously, and that does not finish until the race results are rendered.

ArendPeter: Ah ok, perhaps we can co-work on that then. Feel free to book a meeting with me at starvoting.org/meeting 🙂

Issue #1001, opened 2025-10-18
Editable Ballots - Disable adding voter IDs if user selects email list


JonBlauvelt:
Currently, a user can add a combination of email addresses and voterIDs to an election where they select either voterIDs or email addresses.  Going forward, email list elections should not allow voterIDs to be added.

Issue #998, opened 2025-10-15
Editable Ballots - Update “Copy unique voting url” to be similar to a delete confirmation flow


JonBlauvelt:
Currently, the voter detail view on `https://bettervoting.com/<electionId>/admin/voters has a "copy unique voterId" button with text underneath "(action will be captured in audit log)".  We can remove this text in favor of a confirmation modal which advises that this will reveal sensitive voter info which will be captured in an audit log.

masiarek:
draft doc: Button/link “Copy unique voting url - "break-glass" procedure (bypassing communication via email): https://docs.google.com/document/d/1sc5tTVOAGGQLtJDrEAt1kvpBL2OJ0x2f99AVAsxkv8c/edit?tab=t.0

Issue #997, opened 2025-10-15
Editable Ballot - Verify that voter ID is not accessible through he UI


JonBlauvelt:
This issue is to identify any touch points where the election admin may have access to the voterId in an email list based election.

PR #1017, opened 2025-10-21
Add redact_voter_ids setting to hide voter IDs from admins

jacksonloper:
Implements voter ID redaction for elections that require voter privacy.  Basically after going around in circles a few times it just seemed simplest to make it an election setting (that was forced to be true with editable ballots).  This addresses #997  and also #998 and also a little bit #1001 (though not the frontend side of #1001).  Maybe I tried to do too much here and just created a mess 😂.

Not 100% confident in this yet.  Happy to rewrite.


packages/backend/src/Controllers/Ballot/castVoteController.ts
```diff
@@ -39,7 +39,7 @@ type BallotSubmitType = 'submitted_via_browser' | 'submitted_via_admin' | 'submi
 
 const castVoteEventQueue = "castVoteEvent";
 
-async function makeBallotEvent(req: IElectionRequest, targetElection: Election, inputBallot: NewBallot, submitType: BallotSubmitType, voter_id?: string){
+async function makeBallotEvent(req: IElectionRequest, targetElection: Election, inputBallot: NewBallot, submitType: BallotSubmitType, voter_id?: string, adminEmail?: string){
```
jacksonloper: When an admin bulk uploads ballots on behalf of voters, the actor field records who performed the action. Without the fix, it would store the voter's voter_id as the actor, which is misleading because when redact_voter_ids is enabled - the voter_id gets redacted to "Voter", making the audit log say "Voter submitted ballot for Voter" which is not actually true.  I'm not sure this is the right way but it could work.

packages/frontend/src/components/Election/Admin/EditElectionRoll.tsx
```diff
@@ -59,20 +65,86 @@ const EditElectionRoll = ({ roll, onClose, fetchRolls }:Props) => {
     }) => {
         setDialogOpen(false);
 
-        if (!await sendEmails.makeRequest({
-            target: target,
+        const payload: {
```
jacksonloper: Frontend doesn't generally *know* the voter id anymore.  So sometimes we have to tell the backend to send email on behalf of some [email_address] instead.  This is the frontend part of this issue.

packages/backend/src/Controllers/Election/sendEmailController.ts
```diff
@@ -76,14 +77,29 @@ const sendEmailsController = async (req: IElectionRequest, res: Response, next:
     let message_id = ''
 
     if (email_request.target == 'single') {
-        const electionRollResponse = await ElectionRollModel.getByVoterID(electionId, email_request.voter_id ?? '', req)
```
jacksonloper: Frontend doesn't generally know the voter id anymore. So sometimes we have to tell the backend to send email on behalf of some [email_address] instead.  This is the backend part of this issue.

packages/shared/src/domain_model/ElectionSettings.ts
```diff
@@ -39,6 +39,7 @@ export interface ElectionSettings {
     contact_email?: string; // Public contact email for voters to reach out to
     exhaust_on_N_repeated_skipped_marks?: number; // number of skipped ranks before exhausting
     draggable_ballot?: boolean; // Use draggable interface for IRV ballots
+    redact_voter_ids?: boolean; // Redact voter IDs from all views except emergency break glass endpoint. Required for ballot_updates.
```
JonBlauvelt: Rather than having this as a separate setting, would it make sense to just have this be the behavior for all email-list based elections?

packages/backend/src/Controllers/Ballot/castVoteController.ts
```diff
@@ -163,15 +164,16 @@ async function uploadBallotsController(req: IElectionRequest, res: Response, nex
...
             makeBallotEvent(
                 req,
                 targetElection,
                 structuredClone(
                     mapOrderedNewBallot(ballot, req.body.race_order as RaceCandidateOrder[])
                 ),
                 'submitted_via_admin',
-                voter_id
+                voter_id,
+                req.user?.email // Pass admin email for audit trail
```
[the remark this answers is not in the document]
ArendPeter: The audit log story is still a bit fuzzy to me. We want there to a public version at some point, so names instead of email makes more sense to me. Maybe store user ID for now, and then we can figure out how to retrieve the corresponding name information later?

packages/backend/src/Controllers/Roll/getElectionRollController.ts
```diff
@@ -5,11 +5,102 @@ import { expectPermission } from "../controllerUtils";
...
+    voterId: string | undefined,
+    redact: boolean
+): ElectionRoll['history'] => {
+    if (!history) return history;
+    return history.map((entry: any) => {
+        if (Array.isArray(entry)) {
+            return sanitizeHistory(entry as ElectionRollAction[], voterId, redact);
+        }
+        if (entry && typeof entry === 'object') {
+            const { email_data: _omitEmailData, ...rest } = entry;
```
ArendPeter: Doesn't modifying rest ensure that the original entry isn't modified? I'm pretty sure that's true, but I wanted to double check

packages/frontend/src/components/Election/Admin/EditElectionRoll.tsx
```diff
@@ -59,20 +65,86 @@ const EditElectionRoll = ({ roll, onClose, fetchRolls }:Props) => {
...
             email: { subject, body },
-            voter_id: roll.voter_id,
-        })) return; 
+        };
+
+        if (roll.voter_id) {
+            payload.voter_id = roll.voter_id;
+        } else if (requiresRevealFlow && roll.email) {
+            payload.recipient_email = roll.email;
+        } else if (requiresRevealFlow) {
```
ArendPeter: Could we safely just make this an else instead of else if?

We want to ensure that either voter_id or receipient_email is set before calling sendEmails right?

packages/backend/src/Controllers/Roll/getElectionRollController.ts
jacksonloper: If I'm understanding your question right, then yes.  I believe what you're asking is answered most directly by a snippet like

```
A={'a':1,'b':2};
const {a: _locala, ...rest} = A;
rest.b=3;
(A.b===2) // <-- true
```

Usual caveats about deep modifying of `...rest` apply, but we don't deep-modify it.

packages/backend/src/Controllers/Election/sendEmailController.ts
```diff
@@ -76,14 +77,32 @@ const sendEmailsController = async (req: IElectionRequest, res: Response, next:
...
+        }
+
         if (!electionRollResponse) {
             const msg = `Voter not found`;
             Logger.info(req, msg);
             throw new BadRequest(msg)
         }
         electionRoll = [electionRollResponse]
-        message_id = `dm_${email_request.voter_id}_${0}` //TODO: retreive count of previous dms
+        message_id = `dm_${email_request.voter_id ?? email_request.recipient_email}_${0}` //TODO: retreive count of previous dms
```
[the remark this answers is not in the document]
jacksonloper: Yes I love the idea of voter_id being completely public with one-time-passes for auth :).  It is a bit goofy trying to help frontend render voter information when the obvious voter ID identifier for each voter that we could use in all elections (including non-email-ones) is *sometimes* a secret!

packages/frontend/src/components/Election/Admin/EditElectionRoll.tsx
jacksonloper: Oh yeah good point.  Stuck in my revealflow world not wanting to change anything else.  But in doing so just making it more confusing 😂 .  Fixed.

packages/frontend/src/components/Election/Admin/EditElectionRoll.tsx
```diff
@@ -196,8 +301,41 @@ const EditElectionRoll = ({ roll, onClose, fetchRolls }:Props) => {
...
+
+            {requiresRevealFlow &&
+                <Dialog
+                    open={confirmDialogOpen}
+                    onClose={() => setConfirmDialogOpen(false)}
+                >
+                    <DialogTitle>Obtain Unique Voting URL</DialogTitle>
+                    <DialogContent>
+                        <DialogContentText>
+                            This action reveals the voter ID for this voter so you can share their voting link if the email on file is not reaching them.
```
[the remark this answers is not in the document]
jacksonloper: Fixed

packages/frontend/src/components/Election/Admin/EditElectionRoll.tsx
```diff
@@ -196,8 +301,41 @@ const EditElectionRoll = ({ roll, onClose, fetchRolls }:Props) => {
...
+                        </DialogContentText>
+                        <DialogContentText sx={{ mt: 2 }}>
+                            This action will:
+                        </DialogContentText>
+                        <DialogContentText sx={{ mt: 1 }}>
+                            • Create a permanent audit log entry<br />
+                            • Record who performed this action<br />
+                            • Should only be used when necessary
```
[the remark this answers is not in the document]
jacksonloper: Fixed

packages/backend/src/Controllers/Election/sendEmailController.ts
```diff
@@ -76,14 +77,32 @@ const sendEmailsController = async (req: IElectionRequest, res: Response, next:
     let message_id = ''
 
     if (email_request.target == 'single') {
-        const electionRollResponse = await ElectionRollModel.getByVoterID(electionId, email_request.voter_id ?? '', req)
+        let electionRollResponse: ElectionRoll | null = null;
+
+        // When redact_voter_ids is enabled, voter_id is not available, so use recipient_email instead
```
ArendPeter: nit, I think this comment is outdated

ArendPeter (approved):
Looks good!

packages/backend/src/Controllers/Roll/addElectionRollController.ts
```diff
@@ -23,6 +23,15 @@ const className = "VoterRolls.Controllers";
...
     expectPermission(req.user_auth.roles, permissions.canAddToElectionRoll)
     Logger.info(req, `${className}.addElectionRoll ${req.election.election_id}`);
+
+    // Prevent creating voters by voter_id when using email invitations
+    if (req.election.settings.invitation === 'email') {
+        const hasVoterId = req.body.electionRoll.some((rollInput: ElectionRollInput) => rollInput.voter_id);
+        if (hasVoterId) {
+            throw new BadRequest('Cannot create voters with voter_id when using email invitations');
+        }
+    }
```
JonBlauvelt: Heads-up: there is going to be merge conflict here bc I also made this change and pushed it

PR #1048, opened 2025-10-30
fix: updated test for some of our ui changes

jacksonloper:
## Description

A few changes to the end-to-end tests to make them compatible with our current UI.  This is still in draft state.  Making the PR now in part to see if my npx playwright calls are matching github CI.

### DIFF
--- a/packages/frontend/src/components/ElectionForm/Details/ElectionDetailsForm.tsx
+++ b/packages/frontend/src/components/ElectionForm/Details/ElectionDetailsForm.tsx
@@ -189,7 +189,7 @@ export default function ElectionDetailsForm({editedElection, applyUpdate, errors
                             />
                             <FormHelperText error={!!errors.startTime} sx={{ pl: 0, mt: 0 }}>
                                 {errors.startTime || (editedElection.start_time && timeZone !== DateTime.now().zone.name &&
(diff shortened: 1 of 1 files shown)


jacksonloper:
I think its looking good.  Almost done.  It is currently failing only on tests that actually should fail.  It suggests the following changes:
- we should refuse to continue creating election if " " is the title of the election
- we should refuse to continue creating election if title is too long
- we should set start time of election to current time in current timezone (instead of current time in PDT timezone)
- once the above is fixed, there's a `await page.getByRole('link', { name: 'Results' }).click();` that maybe actually needs fixing in this test but I can't be 100% sure because I'd have to fix the timezone issue before we can even get there :)

tl;dr -- I think I'm going to fix the timezone issue, get that merged into main, send that back over here, then I can finish making sure that the tests are only failing where they are supposed to be failing.

ArendPeter:
Thanks for working on this!

> * we should refuse to continue creating election if title is too long

I think you can remove this test. We used to have a 50 character limit on the election title, but we don't really enforce a limit at the moment (I set it to 1000 arbitrarily 😅)

The other suggestions look good!

ArendPeter:
Looks good! I've checked the test report, and there are many more tests passing. I'll go ahead and merge this and we can fix the remaining tests in a future pass.

PR #1054, opened 2025-10-31
Implement election claim flow

### LINKED ISSUE #429: Implement flow for claiming temp polls (similar to straw poll)

ArendPeter:
## Description

With the new flow, an election can be drafted in the full editor without the user needing to make an account. 

Instead the full editor will show prompts for the user to sign up and "claim ownership" of the election.

Pushing account creation to later in the flow, will hopefully get more users to engage with more of our features, and increase the odds that they sign up.

## Security Flow

All users have a temp_id in their cookies regardless of whether they have an account. Those with an account will have an additional user_id from keycloak. Election's can be owned by a temp user or by a keycloak user.

1. When election is first created, it's owner_id is set to the user's temp_id.
2. The election will have a claim_key. The full claim_key is stored locally in session storage (as <election_id>_claim_key), and a hash is sent to the backend to be saved with the election.
3. When the user is ready to make an account, the election id is saved in session storage as "election_to_claim" for future reference. Then the standard sign in flow starts.
4. Keycloak will redirect to the manage page (so the user can immediately see their election list)
5. When login is complete, the claim api can be called with the "election_to_claim" and "<election_id>_claim_key" from session storage.
6. The backend will verify the following
    1. The user is logged in.
    2. The user's temp_id matches the owner_id of the election
    3. The hash of the claim_key matches the stored claim_key_hash for the election
 7. If it all checks out, the owner_id will be set to the user_id from the keycloak 
 8. The manage page will refresh the election list, as soon as the login and claim flows are complete. 

https://github.com/user-attachments/assets/e9f8a570-1e6a-4eed-b5e4-cf99a580d3c7

## Related Issues

fixes #429

### DIFF
--- a/packages/frontend/src/i18n/en.yaml
+++ b/packages/frontend/src/i18n/en.yaml
@@ -725,8 +725,8 @@ tips:
     description: Randomizes the order of candidates on the ballots.
 
   ballot_updates:
-    title: Allow Voters To Edit Vote
-    description: Allow voters to update their ballots while election is still open
+    title: Allow Voters To Update Ballot
+    description: Allow voters to change their vote while election is still open
 
   public_results:
     title: Show Preliminary Results
--- a/packages/shared/src/domain_model/ElectionSettings.ts
+++ b/packages/shared/src/domain_model/ElectionSettings.ts
@@ -74,13 +74,13 @@ function settingsCompatiblityValidation(settings: ElectionSettings, electionStat
     let errorMsg = ''
     if (settings.ballot_updates) {
         if (settings.public_results && electionState != 'closed') {
-            errorMsg += 'Preliminary results are not permitted when ballot editing is enabled.  ';
+            errorMsg += 'Preliminary results are not permitted when ballot updating is enabled.  ';
         }
         if (settings.voter_access == 'open') {
-            errorMsg += 'Vote editing is not permitted on open elections.  ';
+            errorMsg += 'Ballot updating is not permitted on open access elections.  ';
         }
         if (settings.invitation != 'email') {
-            errorMsg += 'Vote editing is only permitted on email list elections.  ';
+            errorMsg += 'Ballot updating is only permitted on email list elections.  ';
         }
     }
     return errorMsg;


packages/backend/src/Controllers/Election/getElectionsController.ts
```diff
@@ -20,7 +20,7 @@ const getElections = async (req: IElectionRequest, res: Response, next: NextFunc
 
     /////////// ELECTIONS WE OWN ////////////////
     var elections_as_official = null;
-    if(email !== '' || id !== ''){
```
ArendPeter: Previously the /manage page would show all the elections you've made, even if you don't have an account.

This created an awkward window in the login flow where you see your old elections before seeing your new ones. Since the /manage people is only accessible outside of login by direct link, it seemed fine to disable the election list when you're not logged in.

packages/frontend/src/hooks/useSessionStorage.js
```diff
@@ -6,14 +6,10 @@ export const useSessionStorage = (key, defaultValue, updateRate = null) => {
     // The optional input updateRate allows for periodic checking to see if the value in session storage has changed
     // to allow multiple components using the same key to be updated
     const getStoredValue = (key, defaultValue) => {
-        // getting stored value
```
ArendPeter: useSessionStorage used to assume a json structure, but it's not used anymore so I made it more generic

packages/backend/src/Controllers/Election/claimElectionController.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    var failMsg = "Failed to update Election";
+    req.election.owner_id = req.user.sub;
+    const updatedElection = await ElectionsModel.updateElection(req.election, req, `Transferring Ownership`);
+    if (!updatedElection) {
+        Logger.info(req, failMsg);
+        throw new BadRequest(failMsg)
+    }
+
+    res.json({ success: true })
```
JonBlauvelt: Can't we just return an empty 200 or the election object?

packages/backend/src/Controllers/Election/claimElectionController.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    // verify claim key
+    if(hashString(req.body.claim_key) != req.election.claim_key_hash){
+        throw new Unauthorized("User does not have permissions: claim_key mismatch");
+    }
+
+    var failMsg = "Failed to update Election";
+    req.election.owner_id = req.user.sub;
+    const updatedElection = await ElectionsModel.updateElection(req.election, req, `Transferring Ownership`);
+    if (!updatedElection) {
```
JonBlauvelt: I don't think this case is actually possible since in the model we are using `.takeFirstOrThrow()` which will throw an exception if there are no results not return null. Also, an aside, I think we should be separating 4xx from 5xx better.  Bad request should fail input validation or could be thrown in some cases where there is no expected object, but for other failures between server and db we should be throwing 5xx.

packages/backend/src/OpenApi/swagger.json
```diff
@@ -3489,6 +3490,60 @@
...
+        "responses": {
+          "200": {
+            "description": "If claim was successful",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "success": {
+                      "type": "boolean"
```
JonBlauvelt: This still seems weird to me since a 2xx response already indicates success.

packages/frontend/src/components/Election/Admin/AdminHome.tsx
```diff
@@ -223,6 +224,9 @@ const AdminHome = () => {
         />
     }
 
+    const claimElection = () => {
+    }
+
```
JonBlauvelt: why?

packages/frontend/src/components/Election/ElectionStateWarning.tsx
```diff
@@ -3,18 +3,21 @@ import useElection from "../ElectionContextProvider";
...
 
-    return <Paper sx={{display: 'flex', flexDirection: 'row', maxWidth: 600, gap: 2, padding: 2, m: 'auto', mb:4}}>
-        <Typography component="h3">⚠️</Typography>
-        <Box>
-            <Typography component="p"><b>{t(title)}</b></Typography>
-            <hr/>
-            <Typography component="p">{t(description)}</Typography>
+    return <Paper sx={{display: 'flex', flexDirection: 'column', maxWidth: 600, gap: 2, padding: 2, m: 'auto', mb:4}}>
+        <Box display='flex' flexDirection='row' gap={2} sx={{p: 2, m: 'auto'}}>
+            {!hideIcon && <Typography component="h3">⚠️</Typography>}
```
JonBlauvelt: It's outside the scope of this feature, but it seems like we could use an MUI icon for this

packages/frontend/src/hooks/useAPI.ts
```diff
@@ -99,13 +99,17 @@ export const useSetPublicResults = (election_id: string) => {
 }
 
 export const useDeleteAllBallots = (election_id: string) => {
-    return useFetch<{ public_results: boolean }, { election: Election }>(`/API/Election/${election_id}/ballots`, 'delete')
+    return useFetch<undefined, { election: Election }>(`/API/Election/${election_id}/ballots`, 'delete')
```
JonBlauvelt: This API terrifies me

packages/backend/src/Controllers/Election/claimElectionController.ts
ArendPeter: Ah, that makes sense.

Sounds like I can just remove this clause then. If updateElection returns a NoResultError and I don't catch it, then I believe it will translate to a 5xx for the user

packages/frontend/src/components/Election/Admin/AdminHome.tsx
ArendPeter: woops!

packages/frontend/src/components/Election/ElectionStateWarning.tsx
ArendPeter: Sure, that's quick. Here's a comparison. I also did a search and we don't use that emoji anywhere beside the email template (which I think doesn't support mui icons)

Image

Image

packages/backend/src/Controllers/Election/claimElectionController.ts
JonBlauvelt: That seems right - it means we're assuming something went wrong on our side if insert fails which seems reasonable since we should have already validated request and thrown 4xx upstream of this if there was an issue.

PR #1061, opened 2025-11-01
fix: trim empty rows when adding voters to roll

### LINKED ISSUE #1026: Trim empty rows when adding new voters to the roll
Currently including empty lines when adding new voters will result in blank voters being added to the roll

MagsMagnoli:
## Description

Fixes #1026

Currently, including empty lines when adding new voters results in blank voters being added to the roll. This PR filters out empty/whitespace-only rows to prevent this issue.

### Frontend (`AddElectionRoll.tsx`)
- **Manual text entry**: Filter empty/whitespace-only rows before processing
- **CSV upload**: Filter empty rows and roll objects where all fields are empty

### Backend (`addElectionRollController.ts`)
- Add validation to filter out empty roll entries (defense-in-depth)
- Prevents empty voters from reaching the database even if frontend filters are bypassed

## Testing

- ✅ No linter errors
- Existing Playwright tests should continue to pass
- Manual testing: Adding voters with empty lines will now skip those lines

## Implementation Details

Empty row detection:
- Row is empty if `trim().length === 0`
- Roll object is empty if voter_id, email, and precinct are all falsy after trimming

### DIFF
--- a/packages/frontend/src/components/Election/Admin/AddElectionRoll.tsx
+++ b/packages/frontend/src/components/Election/Admin/AddElectionRoll.tsx
@@ -38,7 +38,7 @@ const AddElectionRoll = ({ onClose }: { onClose: () => void }) => {
     const onSubmit = async (e) => {
         e.preventDefault()
         try {
-            const rows = voterIDList.split('\n').filter(row => row.trim().length > 0)
+            const rows = voterIDList.split('\n').filter(row => row.trim())
             const rolls = []
             const expectedCounts = Number(enableVoterID) + Number(enableEmail) + Number(enablePrecinct)
             rows.forEach((row) => {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/Election/Admin/AddElectionRoll.tsx
```diff
@@ -38,7 +38,7 @@ const AddElectionRoll = ({ onClose }: { onClose: () => void }) => {
     const onSubmit = async (e) => {
         e.preventDefault()
         try {
-            const rows = voterIDList.split('\n')
+            const rows = voterIDList.split('\n').filter(row => row.trim().length > 0)
```
JonBlauvelt: This filter expression could also be written equivalently as `row => row.trim()` because trim will return the empty string which is falsey in ts/js

packages/frontend/src/components/Election/Admin/AddElectionRoll.tsx
```diff
@@ -94,14 +94,19 @@ const AddElectionRoll = ({ onClose }: { onClose: () => void }) => {
                 alert('Invalid headers')
                 return
             }
-            const csvRows = text.slice(text.indexOf("\n") + 1).split("\n");
+            const csvRows = text.slice(text.indexOf("\n") + 1).split("\n").filter(row => row.trim().length > 0);
```
JonBlauvelt: Same here

JonBlauvelt (approved):
Hey - thanks for working on this! I left some comments on things but nothing show stopping.  If you're into any of the suggestions, cool, but if not this also works just fine.

PR #1074, opened 2025-11-07
fix: block ballot updates for test ballots

JonBlauvelt:
## Description
Fixes breaking assumption around the presence of voterIds


ArendPeter (approved):
Looks good!

PR #1081, opened 2025-11-10
Feature/validate for duplicate entries

crayne:
## Description
After adding voters and clicking on Submit, the user is asked if they want to eliminate voters wirh duplicate emails.  If they do  want to eliminate duplicates, the duplicates are eliminated.

image

## Related Issues


ArendPeter:
Hi Susan, sorry for the delay on this one. The remove duplication logic looks perfect, but I have some feedback on the confirm flow.

Good work making the confirm window, but we actually have a reusable confirm window in ConfirmationDialogProvider.tsx that you can use. You can see an example of it being used for confirming candidate deletion in RaceForm.tsx. You can also look at "finalize_email_confirm" in the en.yaml for an example that sets custom confirm and cancel messages.

In the near future, I'd like to write up a document for common UI components and utilities so that this will be more clear up front. 
 
For the confirmation options, we shouldn't let the user proceed with duplicate emails. I think it's safe to assume that it was a mistake on the users end. So we're really asking them if they'd like us to resolve it automatically, or let them resolve it manually. How about the following:

Q: You entered duplicate emails, which is not supported. Would you like us to remove duplicates?
Options: Yes, Cancel

PR #1093, opened 2025-11-13
Bug/wrong message invalid owner

crayne:
## Description
When a user was not signed in and tried to create an election or a poll, an exception was thrown.  I created a custom error dialog and called it with a corrected error message, bypassing the exception. The custom error dialog is in components/ErrorDialog.tsx, and can be used to correct other instances where exceptions are thrown because of user error.

image

## Related Issues
In my opinion, this error is being caught too late.  If a user is not signed in, they should not even be allowed to start creating an election or dialog.  It would be best if a user who is not signed could be prevented from seeing "Create Election" on the top menu bar.


crayne:
Yes, I agree.

PR #1169, opened 2025-12-13
Feat/disable admin page for non logged in users

### LINKED ISSUE #1118: Disable Admin Page for non logged in users
### Overview

If a users opens the admin panel for an election they don't own, then they should have an empty page with an error snack. The current experience is confusing and also difficult to maintain.

### Action Items

- [ ] In the AdminHome.tsx, use authsession to verify if the user is logged in
- [ ] If so, then return a Typography element indicating the user doesn't have access (doesn't need to be fancy, this page shouldn't be discoverable anyway except by direct link)

### Resources/Instructions

(note to self, I should make this field optional)

- [x] Role
- [x] Complexity

crayne:
## Description
If a user visits an admin page while not logged on, they will see an error message.  

image
## Related Issues

Fixes #1118


ArendPeter (approved):
Looks good! Thanks for adding this.

PR #1248, opened 2026-02-24
feat: developer election set (+local backend dev fixes)

### LINKED ISSUE #1157: Define a baseline of test elections for setting up your environment
### Overview

The main server has a plethora of fun test examples to choose from. However, in dev work, we can't "dig in" to those specific elections.  It would be very helpful for testing and internal documentation to have a set of elections that are "natural test cases."

### Action Items

- [ ] Create a small curated collection of test elections, including lots of votes.
- [ ] Create a way to store these test elections in the repo
- [ ] Create a mechanism for dev folks to use the elections for actual testing.  I can imagine two ways of doing this.
  - Create a mechanism for dev folks to automatically reset these elections to a particular state in the db (from CLI).  Create another mechanism so dev can (locally!) assign or unassign themselves to be owners of these test elections (from CLI).
  - OR a mechanism that automatically *creates* all of the test elections (from CLI) with a fixed owner and automatically deletes previous versions of the elections.  It would be nice if the elections always even always had the same URLs 😅 .
 
Note: we need to think carefully about how to ensure there's no chance this "convenience" introduces security issues in production.  

### Resources/Instructions

This issue necessarily involves a few different kinds of folks in the project, but I think anyone could take a stab at any of the  action items above and assign themselves if they want.  

- [x] Role
- [x] Complexity

jacksonloper:
This is possibly a terrible PR, because it

* Creates a system for local dev users to create a suite of fixed known elections that can be used for testing things out.  Right now it has only one election in it.  If developed slightly more with adding more actual curated elections, this would close #1157 
* Fixes dev-realm.json to not need ssl (otherwise in my hands it fails if you use docker-compose to launch keycloak)
* Fixes the backend to not fail if azure isn't there, but instead to use the mock blob service that someone already setup for that purpose in testing

which is too many things.  But anyway the main thing is the developer elections set.  Some design decisions here...
* Run via `npm run makedevelections` (creates the suite of development elections) 
* Run via `npm run makedevelections -- -force` (deletes existing elections from the suite and all ballots for those elections, and recreates the suite).  
* We add the elections to whatever `DATABASE_URL` is set in `packages/backend/.env`.  
* The set of elections in the suite is the set of files in the `packages/backend/src/DevElections/elections/` directory
* Elections have ids like devtest[name], for example devtestwizardstar.  So they won't conflict with the 7-character election ids that we normally create.  If the elections don't have name like this we'll refuse to add to database.  
* Ballots have names like b-[electionid]-ballot[number], like `b-devtestwizardstar-ballot0`.  So they won't conflict with 8-character ballot ids that we normally create.  If ballots dont have a name like this we'll refuse to add to database.
* Election owner is always `7bdcad1b-55cd-4cfd-842f-6be3fa89f1c3`, which is the PlayWrightTest user already spawned when keycloak starts in dev mode (password is "test"), so devs can play being that user or not being that user.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -13,7 +13,6 @@
     "migrate:up": "node ./build/src/Migrators/migrate-up.js",
     "migrate:down": "node ./build/src/Migrators/migrate-down.js",
     "db_sandbox": "npm run-script build && node ./build/src/test/database_sandbox.js",
-    "makedevelections": "tsx ./src/DevElections/makeDevElections.ts",
     "generate:openapi": "ts-node ./src/OpenApi/generateSwaggerJSON.ts"
   },
   "keywords": [],
--- a/packages/shared/src/utils/makeID.ts
+++ b/packages/shared/src/utils/makeID.ts
@@ -55,6 +55,34 @@ export async function makeUniqueID(
   return currentId;
 }
 
+// Dev election constants and helpers
+export const DEV_ELECTION_PREFIX = 'devtest';
+
+export function devElectionId(name: string): string {
+    return `${DEV_ELECTION_PREFIX}${name}`;
+}
+
+export function devBallotId(electionId: string, index: number): string {
+    return `${ID_PREFIXES.BALLOT}-${electionId}-ballot${index}`;
+}
+
+export function devBallotIdPrefix(electionId: string): string {
+    return `${ID_PREFIXES.BALLOT}-${electionId}-ballot`;
+}
+
+export function validateDevElectionId(electionId: string): void {
+    if (!electionId.startsWith(DEV_ELECTION_PREFIX)) {
+        throw new Error(`Election ID "${electionId}" must start with "${DEV_ELECTION_PREFIX}"`);
+    }
+}
+
+export function validateDevBallotId(ballotId: string, electionId: string): void {
+    const expectedPrefix = devBallotIdPrefix(electionId);
+    if (!ballotId.startsWith(expectedPrefix)) {
+        throw new Error(`Ballot ID "${ballotId}" must start with "${expectedPrefix}" — use devBallotId() to generate ballot IDs`);
+    }
+}
+
 // Synchronous version for when collision checking is needed but async isn't required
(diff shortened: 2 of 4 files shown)


packages/backend/src/ServiceLocator.ts
```diff
@@ -142,7 +142,13 @@ function emailService(): EmailService {
 
 function blobService(): BlobService {
     if (_blobService == null) {
-        _blobService = new BlobService();
```
ArendPeter: I also ran into this and added a code change at  https://github.com/Equal-Vote/bettervoting/pull/1252/changes#diff-b01c0eac972dace00be57c8e4a6647bf0f7f6bb6ef575df8b9d66b0c549c55eaL9 

That said, I think I like your mock approach better since we it lets us avoid handling edge cases in the production code. 

Can you remove the blobService changes when merging #1252 ?

packages/backend/package.json
```diff
@@ -13,6 +13,7 @@
     "migrate:up": "node ./build/src/Migrators/migrate-up.js",
     "migrate:down": "node ./build/src/Migrators/migrate-down.js",
     "db_sandbox": "npm run-script build && node ./build/src/test/database_sandbox.js",
+    "makedevelections": "tsx ./src/DevElections/makeDevElections.ts",
```
ArendPeter: nit: I prefer to have keep the command be in one place. Maybe it's slightly more intuitive for the user to have it defined in the root package? But I also see the case for keeping it in backend since that's where the code lives. I could go either way.

ArendPeter (changes requested):
Looks great, and it's not too long at all! Just requesting a few minor changes.

packages/shared/src/utils/makeID.ts
```diff
@@ -55,6 +55,34 @@ export async function makeUniqueID(
   return currentId;
 }
 
+// Dev election constants and helpers
```
ArendPeter: Great, thanks for adding these here!

ArendPeter (approved):
Looks great!

PR #1262, opened 2026-03-05
Write in candidates

jacksonloper:
A very minimal write-in candidate frontend implementation based on Mike Franze's backend work.  If write-ins are checked as allowed for a race....

### Voter experience

Voter can add write in candidates to that race, which then get treated as if they were there the whole time by the frontend (these should be kept to the bottom since they are "special" like none of the above).  

Screenshot 2026-03-05 at 12 22 09 PM

Frontend treats all trim-lowercase as the same "name" and so refuses to add JOHN if John has already been added.

Screenshot 2026-03-05 at 1 41 45 PM

### Admin experience

Admin will see for those races a fun little new icon

Screenshot 2026-03-05 at 12 20 14 PM

If they click on it they will be taken to an approval page where they can approve or disapprove and set the official name of a write-in candidate

Screenshot 2026-03-05 at 12 46 55 PM

### Results page

Tabulation then treats approved candidates as normal and others as ignored.  User sees something like this in the results

Screenshot 2026-03-05 at 12 25 40 PM

### Closing thoughts

This approach essentially (1) ignores capitalization (2) lets admin pick "real name" and (3) doesn't let admin merge "JFK" and "John F Kennedy".  It is conceptually simple, weak, but i think viable for "ticking the box" in elections where bylaws require it.  Basically election admins won't even know that our backend supports an aliasing system that can support admin-specified write-in merging.  When we're ready (or if anyone asks for it) we can implement the rest.

Here's my rationale for not implementing frontend for the candidate merging in this PR: write-in-merging has insane edge cases.  Here's one example, but I fear there may be lots like this.  What if a voter makes writein "JFK" and gives them 3 stars and also "John F Kennedy" and gives them 5 stars in a multi-winner election, and admin merges "JFK" and "John F Kennedy"?  How should ballot be counted?  It would seem to necessitate either A Choice or else A Complicated User Interface for admins to handle the mess.  I don't want to make either of those things if I can avoid it 😄. 

## Summary by CodeRabbit

* **New Features**
  * Voters can add up to 5 write-ins; write-ins persist on ballots, appear in UI, results, and admin review/approval flows.
  * Race editor: "Allow write-ins" toggle and admin page to approve/edit/publish write-in candidates.
  * New admin API endpoints to fetch write-in name counts and submit approved write-in results.

* **Bug Fixes / Validation**
  * Client/server validation for write-in names, duplicates, limits, and optimistic concurrency on publishing.

* **Tests**
  * End-to-end tests covering write-in submission, validation, aggregation, approval, and results.

### DIFF
--- a/packages/frontend/src/components/Election/Results/Results.tsx
+++ b/packages/frontend/src/components/Election/Results/Results.tsx
@@ -462,7 +462,8 @@ export default function Results({ race, results }: {race: Race, results: Electio
           <Typography variant='h5'>⭐{results.summaryData.candidates[0].name} wins uncontested⭐</Typography>
           {results.writeInDiagnostics?.numScoresDisregarded > 0 &&
             <Typography component="p" sx={{color: '#808080', fontSize: '0.9rem', mt: 1}}>
-              {results.writeInDiagnostics.numScoresDisregarded} write-in vote{results.writeInDiagnostics.numScoresDisregarded === 1 ? '' : 's'} not included in results
+              {results.writeInDiagnostics.numScoresDisregarded} write-in score{results.writeInDiagnostics.numScoresDisregarded === 1 ? '' : 's'} not counted.{' '}
+              <a href="https://docs.bettervoting.com/help/faq.html#write-in-scores-not-counted" target="_blank" rel="noopener noreferrer" style={{color: '#808080'}}>Learn more</a>
             </Typography>
           }
         </>}
--- a/docs/help/faq.md
+++ b/docs/help/faq.md
@@ -8,9 +8,22 @@ parent: BetterVoting Documentation
 # Frequently Asked Questions
 
 - [How do you run races with "None of the Above"?](#how-do-you-run-races-with-none-of-the-above)
+- [What does "write-in scores not counted" mean?](#write-in-scores-not-counted)
 
 ## How do you run races with "None of the Above"?
 
 Some elections require including “None of the Above” (NOTA) as an option. On BetterVoting.com, you can include NOTA by simply adding “None of the Above” as a candidate. If NOTA wins, the election may need to be repeated. If NOTA receives a majority but does not win, the election would proceed according to your organization’s bylaws.
 
 Most voting methods on BetterVoting allow voters to support multiple candidates, which can include NOTA. For example, in an Approval voting election, a voter could select both a candidate and “None of the Above.” This aligns with Equal Vote’s principle that voters should be able to express their full and honest opinions, and that their vote should always be able to make a difference—even if their favorite candidate can’t win.
+
+## What does “write-in scores not counted” mean?
+{: #write-in-scores-not-counted}
+
+When write-in candidates are enabled for a race, voters can add candidates that aren’t on the official ballot. Before these write-in candidates appear in results, an election administrator must review and approve them. This prevents misspellings, duplicates, and joke entries from cluttering the results.
+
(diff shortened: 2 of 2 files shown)


jacksonloper:
The nefariousness-avoiding commit made tests fail!

Screenshot 2026-03-05 at 2 53 43 PM

When our ballots have "null" my nefariousness checker doesn't like it.  In general submitting strings for ballots could be bad (think 0+"5" = "05" = cascading failure of tabulation 😂 ).  I'll contemplate and make a plan tomorrow or next week!

packages/frontend/src/components/Election/Admin/WriteInApproval.tsx
```diff
@@ -0,0 +1,202 @@
...
+    TableContainer, TableHead, TableRow, Paper, Checkbox, TextField
+} from '@mui/material';
+import ArrowBackIcon from '@mui/icons-material/ArrowBack';
+import IconButton from '@mui/material/IconButton';
+import { PrimaryButton } from '../../styles';
+import useElection from '../../ElectionContextProvider';
+import { useGetWriteIns, useSetWriteInResults } from '../../../hooks/useAPI';
+import { WriteInCandidate } from '@equal-vote/star-vote-shared/domain_model/WriteIn';
+
+const trimLower = (s: string) => s.trim().toLowerCase().normalize('NFC');
```
ArendPeter: ah here's another trimLower. Sounds like this needs to be added to a util script within shared

packages/frontend/src/components/Election/Results/Results.tsx
```diff
@@ -478,6 +479,11 @@ export default function Results({ race, results }: {race: Race, results: Electio
             </Typography>
           }
           <Typography variant="h6">{t('results.vote_count', {n: results.summaryData.nTallyVotes})}</Typography>
+            {results.numExcludedWriteIns > 0 &&
+              <Typography component="p" sx={{color: '#808080', fontSize: '0.9rem', mt: 1}}>
+                {results.numExcludedWriteIns} write-in vote{results.numExcludedWriteIns === 1 ? '' : 's'} not included in results
```
ArendPeter: This doesn't look right to me. results.numExcludedWriteIns refers to excluded candidates, not excluded votes right?

I'm not sure where we want to display the number of excluded write in candidates. If the information is noisy, then  it could be buried in the detailed results somewhere.

packages/frontend/src/components/Election/Voting/VotePage.tsx
```diff
@@ -233,9 +277,11 @@ const VotePage = () => {
         setWarnings: setWarnings,
         alertBubbles: pages[currentPage].alertBubbles,
         setAlertBubbles: setAlertBubbles,
-
+        addWriteIn: addWriteIn,
+        removeWriteIn: removeWriteIn,
       }}>
         <BallotPageSelector votingMethod={pages[currentPage].voting_method} />
+        <WriteInSection />
```
ArendPeter: Having the write in selector so far away from the ballot feels like odd UX, but we can save that till later. I'm happy to help with that UI/UX pass.

ArendPeter (changes requested):
Great work @jacksonloper ! Thank you so much for picking up where Mike left off, and bringing the feature to a shippable state :D 

I've requested several changes, but many of those could probably be saved for a future PR.

I've also realized we have the LPO election coming up this Sunday, so I'll hold off on approving this until next week.

PR #1264, opened 2026-03-08
fix(account): prevent custom auth_key spoofing and algorithm confusion

jwidan:
## Description
This PR stops an authentication bypass flaw that let anyone forge their own JWT verification keys. Before this PR, the backend insecurely took the auth_key straight from the incoming JSON body. This let anyone sign their own tokens and attach the corresponding public key to their request body.

To fix this, the auth_key override was removed. The desired custom key feature should still work with electionSpecificAuth and accountService.extractUserFromRequest.

Main changes:

* Removed the vulnerable inputElection payload extraction block from AccountServiceUtils.ts
* Specified allowed algorithms and created a quick heuristic to detect if the provided key is symmetric or asymmetric
* Cleaned up unused dependencies left behind

My previously successful attempts to get the server to accept my spoofed JWT no longer work with this PR.

### DIFF
--- a/packages/backend/src/Services/Account/AccountServiceUtils.ts
+++ b/packages/backend/src/Services/Account/AccountServiceUtils.ts
@@ -1,4 +1,3 @@
-import { Election } from "@equal-vote/star-vote-shared/domain_model/Election";
 import { IRequest } from "../../IRequest";
 import Logger from "../Logging/Logger";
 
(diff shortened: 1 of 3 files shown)


packages/backend/src/Services/Account/AccountServiceUtils.ts
```diff
@@ -11,17 +10,12 @@ export default class AccountServiceUtils {
         token: string,
         key: string
     ) => {
-        const inputElection: Election | undefined = req.body.Election;
```
ArendPeter: Auth_key is important for 3rd party integrations (like discord), that don't use keycloak for authentication. So we need to find a way to keep it without having the auth_key spoofing loophole for elections created on the bettervoting website.

The request goes through multiple layers of middleware when it hits the background. So there's certain parts of it that were safely inserted by the backend, rather than coming directly from thee user.

I had thought that req.body.Election that trustworth election, but after reviewing [the getElectionByID middleware](https://github.com/Equal-Vote/bettervoting/blob/036ada08bdb15599b8027dc49a9fbdce3f95b2b7/packages/backend/src/Routes/elections.routes.ts#L729) I believe that might need to be req.Election ?

Can you confirm if this would be sufficient to resolve the issue?

Issue #1225, opened 2026-01-30
Clarify Strange Owners


jacksonloper:
Elections can have weird owners.  Some examples.

- I can make an election with owner id 0 (backend doesn't validate, and in fact it will default to 0 if temp_id cookie gets disrupted)
- I can make live elections owned by my temp id
- Whether or not I am shown the `electionToClaim` flow component gets messed up by the wizard somehow, so I can actually create a *finalized* election whose owner is my temp id, without logging in, and learn only 10 hours later that actually I can no longer visit the admin home.  
- Finalized elections can become "orphaned" after 10 hours, so they have *no* owner anymore.

There's a bit of cruft, in short.  Figuring out the intended flow (as Jon has already made inroads towards, but I think this anonymous-owner piece was missing from his picture) would be a good thing.  

My immediate knee jerk instinct is to just remove the "temporary owners" feature as we try to prepare to leave beta, as it introduces a lot of confusing edge cases.  But happy to contemplate alternatives!

PR #1266, opened 2026-03-08
fix(security): prevent temp_id cookie spoofing and impersonation

jwidan:
## Description

This PR stops an account impersonation vulnerability that let unauthenticated users grant themselves the 'owner' role on any election.

Before this PR, `AccountService` parsed missing JWT tokens by packaging unverified temp_id cookie strings into a mock `{ typ: 'TEMP_ID' }` user object. If a user set their `temp_id` cookie to match an election's UUID, the `electionPostAuthMiddleware` blindly trusted the equality match `(req.election.owner_id == req.user.sub)` and instantly granted them full owner privileges.

To fix this, we also require that `req.user.typ !== 'TEMP_ID'`. Because authenticated users receive a `typ: 'ID'` from Keycloak, this blocks an unverified cookie payload from passing the ownership check.

My previously successful attack to takeover elections as an unauthenticated user no longer works with this PR.

### Additional Fixes
While testing, some blockers in the anonymous-to-authenticated claim progression were also discovered and resolved:

* Returning anonymous users who had historically broken session state were defaulting to `temp_id: '0'`. Introduced a migration hook in `<Header>` and a strict intercept in `<Wizard>` to guarantee all unauthenticated sessions initialize with `v-` before election creation.
* Legitimate claims were succeeding on the backend but not triggering the success snack. The claim correctly returned an empty body, resulting in `undefined`, which evaluated to falsy on the frontend. `<ElectionsYouManage>` was updated to explicitly check `if (res !== false)` to correctly trigger the success snack and clean the claimKey cookie.
* An anonymous user who immediately published their election was previously blocked from claiming it because the claim banner was hardcoded exclusively to `state='draft'`. `<ElectionStateWarning>` was updated to allow anonymous owners to claim their elections regardless of state.

## Related Issues
#1225

### DIFF
--- a/packages/frontend/src/components/Election/ElectionStateWarning.tsx
+++ b/packages/frontend/src/components/Election/ElectionStateWarning.tsx
@@ -8,7 +8,7 @@ export default function ElectionStateWarning
     
     const { t, election } = useElection();
     
-    if(election.state !== state) return <></>
+    if(state && election.state !== state) return <></>
 
     return <Paper sx={{display: 'flex', flexDirection: 'column', maxWidth: 600, gap: 2, padding: 2, m: 'auto', mb:4}}>
         <Box display='flex' flexDirection='row' gap={2} sx={{p: 2, m: 'auto'}}>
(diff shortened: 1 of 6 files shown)


ArendPeter (approved):
Looks good, nice catch!

Just confirm that the claim flow showed on https://github.com/Equal-Vote/bettervoting/pull/1095 still works, and then we should be good to merge

PR #1284, opened 2026-03-19
Set up table and frontend for storing email webhook information

jacksonloper:
In theory, this PR finishes the full MVP for election admins to see the status of their emails.  For now it looks like this.

Screenshot 2026-03-19 at 10 13 06 AM

Key design decisions in this PR.  

1. Separate table, not embedded on the roll. Email events live in their own emailEventsDB table rather than being stuffed into the roll's email_data JSON. Events are joined at response time via a new ElectionRollResponse type. This keeps the roll document clean and lets events arrive asynchronously from webhooks.
1. Two insertion points. A "sent" event is recorded when the app sends an email (keyed by SendGrid's x-message-id). Subsequent lifecycle events (delivered, bounced, etc.) come in via SendGrid webhook and are matched back to the original send by stripping the .filter suffix from sg_message_id.
1. Fail-safe everywhere. Every email event DB operation is wrapped in try/catch. A broken events table never prevents emails from sending or rolls from loading — it just logs and moves on.
1. No pagination (intentional deferral). The rolls endpoint was already unpaginated. Email events add to the payload but the decision was to leave that alone for now rather than take on the full pagination lift. Can revisit if elections with thousands of email voters become a problem.
1. Migration runs automatically. The app's startup script runs migrate-to-latest.js before the server starts, so the new table will be created on the next deploy with no manual steps.                                                                                                                       

### `emailEventsDB` table schema

| Column | Type | Description |
|--------|------|-------------|
| `id` | serial PK | Auto-increment |
| `message_id` | varchar | SendGrid's `x-message-id` (base, `.filter` suffix stripped) |
| `election_id` | varchar | Which election |
| `voter_id` | varchar | Which voter |
| `event_type` | varchar | `sent`, `processed`, `delivered`, `bounce`, `deferred`, `dropped`, `open`, `spamreport`, etc. |
| `event_timestamp` | bigint | Epoch ms for "sent" events, SendGrid's timestamp for webhook events |
| `details` | json, nullable | Extra context (varies by event type, see below) |

**Indexes:** `(election_id, voter_id)` and `(message_id)`

### What goes in `details`

The webhook handler destructures out `email`, `unique_args`, `sg_message_id`, `event`, and `timestamp` before storing — everything else from the SendGrid payload lands in `details`:

| Event type | Typical `details` |
|------------|-------------------|
| sent | `{ status_code: 202 }` |
| delivered | `{ response: "250 OK" }` |
| bounce | `{ reason: "550 No such user", status: "5.1.1", bounce_classification: "Invalid" }` |
| deferred | `{ reason: "421 Try again later", attempt: "1" }` |
| dropped | `{ reason: "Bounced Address", status: "5.0.0" }` |
| open / spamreport | typically empty |

## Summary by CodeRabbit

* **New Features**
  * Email delivery tracking: sent messages and subsequent delivery events are recorded and shown in the admin UI per voter, with timestamps, event types, and status/details.
* **Tests**
  * Webhook tests added to validate signature/timestamp checks, event normalization, multi-event handling, and DB inserts while protecting PII.
* **Chores**
  * Database migration, dev fixtures, mocks, and service wiring added to support persistent email events.

### DIFF
--- a/packages/frontend/src/App.tsx
+++ b/packages/frontend/src/App.tsx
@@ -61,9 +61,6 @@ const App = () => {
         <Box display='flex' flexDirection='column' minHeight={'100vh'} sx={{backgroundColor:'white'}} onDragOver={(e) => e.preventDefault()} onDrop={(e) => e.preventDefault()} >
           <ReturnToClassicLayer/>
           <Header />
-          <Typography sx={{textAlign:'center', padding: 2, opacity: 0.5}}>
-            {t('nav.beta_warning')}
-          </Typography>
           <Box
             sx={{
               width: '100%',
--- a/packages/frontend/src/components/LandingPage.tsx
+++ b/packages/frontend/src/components/LandingPage.tsx
@@ -56,13 +56,16 @@ const LandingPage = () => {
                 alignItems: 'center',
                 textAlign: 'center',
             }}>
+                <Typography sx={{textAlign:'center', padding: 2, opacity: 0.5}}>
+                    {t('nav.beta_warning')}
+                </Typography>
                 <Typography variant="h4" color={'lightShade.contrastText'}> {t('landing_page.hero.title')} </Typography>
                 <LandingPageCarousel />
                 <Typography component="p" sx={{margin: 'auto', width: '80%', textAlign: 'center', mt: 4}}>
                     <i>"BetterVoting is your one-stop, open-source tool for handling all your election needs. Whether it's informal polls or highly secure elections, electronic or paper, single-seat or multi-seat, we've got you covered!" <span className="nobr">- The BetterVoting Team</span></i>
                 </Typography>
             </Box>
-            <LandingPageStats/>
+            {/*<LandingPageStats/> temporarily disabling, https://github.com/Equal-Vote/bettervoting/issues/1294*/}
             {/* putting the anchor on the Wizard component scrolls too far */}
             <div id='wizard'></div>
             <Wizard/>
--- a/packages/frontend/src/i18n/en.yaml
+++ b/packages/frontend/src/i18n/en.yaml
@@ -832,7 +832,7 @@ tips:
 draft_warning:
   title: Test Mode
   description: >
-    DO NOT SHARE PUBLICLY! This {{election}} is still being drafted.
+    This {{election}} is still being drafted.
     All ballots will be counted as test votes and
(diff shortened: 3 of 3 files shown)


packages/backend/src/Controllers/Election/sendEmailController.ts
```diff
@@ -202,6 +203,23 @@ async function handleSendEmailEvent(job: { id: string; data: email_request_event
         throw new InternalServerError(`Couldn't send email: ${e}`);
     }
 
+    // Record the sent event in the email events table
+    const xMessageId = emailResponse?.[0]?.[0]?.headers?.['x-message-id'];
```
JonBlauvelt: Are these always guaranteed to be lists of length one?

ArendPeter:
> This will currently drop information about _ballot receipts_. Do we want to store that information in the database as well?

That's okay with me as a first pass. The recent Indivisible election has got me thinking about features to add for unauthenticated elections. One of those is to give a way to notify voters when the results are live, and that could require storing the email given with the ballot receipt, but we can think more on that later.

packages/backend/src/Controllers/sendGridWebhookController.ts
```diff
@@ -12,41 +12,85 @@ interface SendGridEvent {
...
+        const valid = crypto.verify('SHA256', Buffer.from(payload), publicKey, Buffer.from(signature, 'base64'));
+
+        if (!valid) {
+            Logger.warn(req, `SendGridWebhook: invalid signature`);
+            res.status(403).send('Invalid signature');
+            return;
+        }
+
+        const timestampAge = Math.abs(Date.now() / 1000 - Number(timestamp));
+        if (isNaN(timestampAge) || timestampAge > 300) {
```
ArendPeter: Could you elaborate on this? Could this 5 minute threshold cause us to miss out on events?

jacksonloper: The timestamp the timestamp of the moment when sendgrid POSTs to us.  It should be like 5 seconds before now at most.  Its part of the security, so there's no replay (though I can hardly imagine it mattering :joy: ).

jacksonloper: just in case sendgrid gets deeply backed up, i'll make it 24 hrs.  it doesn't really much matter anyway.

PR #1297, opened 2026-03-27
Add random tiebreak shuffle using tinyrand

### LINKED ISSUE #1296: Random Tiebreaker fix
## Random tie breaker bug
Another possibly related issue occurred when we did a 4-way tie scenario with STAR. Notice how the tabulation logs show d & b advancing but then the scoring round summary shows a and b advancing

Image

[Ballot Data - abcd-ykftw9.csv](https://github.com/user-attachments/files/26289591/Ballot.Data.-.abcd-ykftw9.csv)

This was prompted by the following slack thread https://starvoting.slack.com/archives/C01EBAT283H/p1774496241382089 

## Root Cause

Image

On line 160 we just completely ignore all the finalists we selected in the loop [facepalm]

## TinyRand Python disucssion

Original Tiebreaker response from Tim: https://discuss.python.org/t/finding-a-bloc-star-provider/73918/100
Motivation for tinyrand: https://discuss.python.org/t/finding-a-bloc-star-provider/73918/119

## Action Items

This prompts some immediate follow ups, as well as some longer term ones

- [ ] Fix the Arend's embarrassing bugs in the tabulator 😆, this should get random tiebreakers working again for STAR
- [ ] Apply the shuffleCandidates utility function for the other voting methods so that we get random results on all of them. 
- [ ] Adjust the seeding for shuffleCandidates to also incorporate the voting method. Perhaps translate the voting method to an index and then do method_index*1000 + tally_votes.length? That keeps the tiebreaker from resolving the same way everytime when we have a multi-method election.
- [ ] Update shuffleCandidates to use tinyrand (see the comment in the function). No need to import, it case just be copied from the source.  
- [ ] Consider how to edit sortCandidates, currently it always sorts to the highest value but I have a feeling that tieBreakOrder will be flipped

ArendPeter:
## Summary
- Introduces `tinyrand`-based deterministic shuffle for candidate tiebreak ordering
- Exposes `tieBreakOrder` field on candidates and surfaces tiebreaker info to the frontend
- Applies tiebreaker logic consistently across all voting methods (STAR, IRV, RankedRobin, AllocatedScore)
- Fixes the shuffle: `getTinyRand(0, seed).shuffle(candidates)` was commented out and never actually called

## Test plan
- [x] Verify tiebreak order is deterministic for a given election (same seed → same order)
- [x] Confirm tiebreak info displays correctly in election results UI
- [x] Run backend unit tests: `npm test -w @equal-vote/star-vote-backend`
- [x] Run E2E tests: `cd testing && npx playwright test --reporter=list`

## Demo
https://www.youtube.com/watch?v=N9ThcCQWhqk

fixes #1296


ArendPeter:
I checked the End-to-end tests failure. This is a race condition that's addressed in https://github.com/Equal-Vote/bettervoting/pull/1321

Issue #1302, opened 2026-04-01
Delay between keystrokes


jacksonloper:
At least in the wizard, and probably other places, it can take up to .5 seconds between keystrokes in an input box, at least in part because we're doing more cloning than probably strictly necessary.  It should be straightforward to make it much more responsive (e.g. only set onblur).

jacksonloper:
Immer could also speed it up.  But if we did immer and still tried to set on every keystroke it would still be kind of pointlessly expensive.  Only setting onblur seems better to me.  Just my 2 cents.

jacksonloper:
Immer would probably not speed it up actually because I think MOST of the time is going into rerunning useSubstitutedTranslation.  This probably needs to be memoized.

jacksonloper:
Closed by #1310

PR #1310, opened 2026-04-09
only trigger rerender after onblur, memoize i18next

jacksonloper:
Some text boxes are molasses, including the wizard (a whopping 1.5 seconds interaction-to-next-paint).  

Some of this is i18next, which can be better memoized, so I did this.

There are other things too though, including the measurement that MUI does for the text area resizing and so forth... an argument can be made that deepcopying the whole form is just too performance-intensive even though it does simplify the maintainability of the code.  That's a debate for another day.  

Anyway, in the short term, to avoid the most intense pain point, which is keystrokes, the most immediate solution is to not rerender the whole form on each keystroke, but only onblur of any textbox.   This PR does that at least for the wizard text boxes (`RaceForm`) so it is fast now (<50ms).    

If we like this pattern we can use it on `ElectionDetailsForm` as well. 

This PR is to address #1302 .

### DIFF
--- a/packages/frontend/src/components/util.tsx
+++ b/packages/frontend/src/components/util.tsx
@@ -8,7 +8,19 @@ import { AddCircleOutlineRounded, ArrowForwardIos } from "@mui/icons-material";
 import { getEntry } from "@equal-vote/star-vote-shared/domain_model/Util";
 import { createHash } from "crypto-browserify";
 import RemoveCircleOutlineRoundedIcon from '@mui/icons-material/RemoveCircleOutlineRounded';
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+// Holds a local copy of `value` for fast keystrokes, flushing to parent on blur.
+export const useLocalState = <T,>(value: T, onFlush: (v: T) => void): [T, (v: T) => void, () => void] => {
+    const [local, setLocal] = useState(value);
+    const localRef = useRef(local);
+    localRef.current = local;
+    useEffect(() => setLocal(value), [value]);
+    const flush = useCallback(() => {
+        if (localRef.current !== value) onFlush(localRef.current);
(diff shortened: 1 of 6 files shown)


jacksonloper:
Hmm.  investigating playwright tests... it seems the be that playwright can somehow fill a textbox without blurring it before it goes onto the next thing.  at the risk of sounding insane I feel like solution is to change the test here.  going with that.

packages/frontend/src/components/util.tsx
```diff
@@ -8,7 +8,19 @@ import { AddCircleOutlineRounded, ArrowForwardIos } from "@mui/icons-material";
 import { getEntry } from "@equal-vote/star-vote-shared/domain_model/Util";
 import { createHash } from "crypto-browserify";
 import RemoveCircleOutlineRoundedIcon from '@mui/icons-material/RemoveCircleOutlineRounded';
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+// Holds a local copy of `value` for fast keystrokes, flushing to parent on blur.
```
ArendPeter, line 13: Nice! I think I was trying to solve a similar problem in my PR by adding useSyncState https://github.com/Equal-Vote/bettervoting/pull/1321/changes#diff-dcb24686bec7b05857c6788f8749e5bad7794f37348c3408d3c00c12834b5abc

If I understand correctly both aim to make the experience more responsive by maintaining a local value, but useLocalState will sync based on onFlush, whereas useSyncedState uses a debounce timer and it also has a rollback in case the backend fails. 

Perhaps they could be combined in the future? Debounce can be the default behavior but it can also disabled if the user wants to control when the sync happens (like onBlur)?

Either they can co-exist for now and then we can have a follow up PR later

ArendPeter (approved):
Looks good! I just left a comment for follow up.

PR #1317, opened 2026-04-14
fix(wizard): replace voter list toggle with auto voter ID choice

### LINKED ISSUE #1307: Update voter list modes
Image

Currently, if an admin wants to use a voter list, we'll have them decide between an ID list or an email list. 

We should replace that toggle with a yes/no decision on "Automatically generate voter ID". I think that admins should be able to do any combination of IDs or emails, and IDs exist either way; the real decision is whether or not that ID is automatically generated or manually generated, so I think changing the option would make it more clear.

utsa-aaa:
Replaces the ID list vs email list toggle with a yes/no choice for automatically generating voter IDs.
Screenshot 2026-04-14 at 5 39 38 PM

Changes:
- Added auto_voter_id_yes and auto_voter_id_no mappings
- Updated wizard UI to reflect new choice
- Ensures different election configurations based on selection

Tested locally.

Closes #1307

### DIFF
--- a/packages/frontend/src/components/ElectionForm/Wizard/WizardExtra.tsx
+++ b/packages/frontend/src/components/ElectionForm/Wizard/WizardExtra.tsx
@@ -28,27 +28,27 @@ const templateMappers = {
             },
         }
     }),
-    'email_list': (election) => ({
+    'auto_voter_id_yes':(election:NewElection):NewElection => ({
         ...election,
-        is_public: false,
-        settings: {
+        is_public:false,
+        settings:{
             ...election.settings,
-            voter_authentication: {
+            voter_authentication:{
                 ...election.settings.voter_authentication,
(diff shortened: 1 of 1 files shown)


ArendPeter:
Hi @utsa-aaa, I really appreciate the contribution, and this looks like a good first step, but unfortunately I'm not sure if this is good first issue to start with as a full implementation would require a lot of extra context on the project. I'll close this PR, but I'd still love you for you to get involved if you're interested. I left a comment on https://github.com/Equal-Vote/bettervoting/issues/1307 some more details, but I'll paste them again here. 

I recommend following the steps at bettervoting.com/volunteer . That'll walk you through all the prerequisite steps for contributing to our project, and then it'll get you started on some of the good-first-issue items. I also recommend scheduling a meeting with me at starvoting.org/meeting . I can give more context on the project from there, and help find something that's a good fit for your class project.

PR #1322, opened 2026-04-22
fix sendgrid webhook message id

jacksonloper:
Sendgrid is sending some messaged with a '.recvd' suffix instead of '.filter' suffix.  Update logic to handle all suffixes.


ArendPeter (approved):
Looks good!

PR #1327, opened 2026-04-27
DBs with head=true pattern should never have two rows with head=true

jacksonloper:
Merge this before we launch the new admin panel so we don't have to do manual cleanup later if there are races.


ArendPeter (approved):
Looks good!

PR #1330, opened 2026-04-28
tell mobile browsers we don't do well with dark mode

### LINKED ISSUE #1308: Somehow dark mode made things awful
User reports seeing this. 

Image

We should figure out how it happened and make it not happen 😂 .

jacksonloper:
To fix #1308


ArendPeter (approved):
Love it! I didn't think the fix would be so simple :o

Issue #1316, opened 2026-04-14
Update Voter ID wording


ArendPeter:
The voter ID list question is confusion. Ultimately you still have email and voter id regardless of the option you pick, the question is more between whether the ID should be managed by BV or not, so it should be reworded accordingly

<img width="921" height="470" alt="Image" src="https://github.com/user-attachments/assets/e9fce9af-f8d5-4d6c-ac6e-d0039118e447" />

jacksonloper:
I think this is done.

Issue #1170, opened 2025-12-15
Rework Admin Experience


ArendPeter:
### Overview

The admin experience should be broken down into more steps and feel more like a wizard experience

### Action Items

Hierarchy is drafted in a [google doc](https://docs.google.com/document/d/1Ua7MmubgDcLiQf7slvjZEClWnnT5YT3pKyQYWw0G3Kk/edit?tab=t.nhgrqykb38ro) (in Election Builder tab)

Current Tabs
* Voting Page
* Admin Home
* Voters
* Ballots

New Tabs
* Admin Home
* Build Ballot
* Preview Ballot
* Manage Voters 
* Election Settings (Invisible until previous sections complete.) 
* Publish Election

- [ ] Move races section in to build ballot
- [ ] Rename Voting Page to Preview Ballot
- [ ] Move Security settings into Manage Voters (use a question with fade out panels)
- [ ] Move election Settings to it's own page
- [ ] Move finalize to it's own page
- [ ] And previous/next/skip to end to the bottom of each page
- [ ] Consider adding steps to AdminHome ?

### I have set labels for the following [https://github.com/hackforla/website/wiki/How-to-read-and-interpret-labels](more labels info)

- [x] Role
- [x] Complexity

ArendPeter:
Some spin off items
https://github.com/Equal-Vote/bettervoting/issues/1306
https://github.com/Equal-Vote/bettervoting/issues/1304
https://github.com/Equal-Vote/bettervoting/issues/1316

ArendPeter:
Notes from a review w/ Sara

Current pass (small change to admin rework)
- [x] Don't display the table if it's empty
- [x] Hovering on the sidebar uses a darker gray color, and the current gray color becomes the color for a selected page. The lightest gray can be a bit darker
- [x] The side bar should be sticky and stay there when you scroll. 

Uncategorized
- [ ] After adding email she needed to refresh the page to actually see them in the table
- [x] As a general rule, remove save buttons wherever possible. 
- [x] Allow for editable votes to be available on an election with public results. 
- [x] ~~settings to move, "election results are public", start and end time~~
- [ ] Elections in draft mode should have results be public regardless of the public results setting. 
- [x] Confirm that elections with candidates lower than the max ranking will only use up to that number of candidates. 	
- [ ] Embed the "How does Star Voting work" page as one of the links in the default election invitation template. starvoting.org/voters_guide . "Are you new to STAR Voting? Here's what you need to know"

[...]

PR #1332, opened 2026-04-28
wording for voter id elections

jacksonloper:
To address #1316


jacksonloper:
This would be my 2 cents on voter-managed vs BV-managed ids.  No need to actually use this code if it goes better in the #1170 , but this is what I'd say

ArendPeter (approved):
Works for me! We can always iterate on the wording and the key names further, but this looks like a big improvement over the previous approach :)

PR #1349, opened 2026-05-13
Allow ballot updates and public results to be enabled simultaneously

ArendPeter:
## Context

When doing the admin-rework, I decided to remove the UI logic for enforcing mutual exclusion between public_results and ballot_updates. I think we were over focused on secure elections when adding the ballot_updates feature and added more restrictions than needed. Removing them also makes our UI and backend a bit cleaner.

It's already removed on the frontend, and this change applies the corresponding backend changes. 

## Summary

- Removes the shared-layer compatibility validation that prevented `ballot_updates` and `public_results` from both being enabled when an election is in the `open` state
- Removes the auto-disable logic in `editElectionController` that silently cleared `public_results` when `ballot_updates` was enabled on an open election

## Test plan

- [x] Open an election with `ballot_updates` enabled and toggle `public_results` on — verify both remain set
- [x] Open an election with `public_results` enabled and toggle `ballot_updates` on — verify both remain set
- [x] Verify existing elections that had only one setting enabled are unaffected

### DIFF
--- a/docs/help/preliminary_results.md
+++ b/docs/help/preliminary_results.md
@@ -19,7 +19,7 @@ When the setting is on:
 - The tally updates as ballots come in. A new ballot changes the numbers immediately.
 - After the election closes, the same view becomes the final results.
 
-When the setting is off, the tally is hidden from everyone (including voters and the public) until the election is closed and the admin chooses to publish.
+When the setting is off, the tally is hidden from everyone (both admins and voters) until the election is closed and the admin chooses to publish.
 
 ## What admins can see during an election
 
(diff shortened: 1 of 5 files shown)


jacksonloper (commented):
Still to fix:
- https://github.com/Equal-Vote/bettervoting/blob/9db4e063b778309ab49e33ec406b16f8abb15b3c/packages/backend/src/Controllers/Election/setOpenStateController.ts#L35 enforces the mutual exclusion
- https://github.com/Equal-Vote/bettervoting/blob/9db4e063b778309ab49e33ec406b16f8abb15b3c/packages/frontend/src/i18n/en.yaml#L687 needs adjusting

jacksonloper:
I took the liberty of pushing a markdown file to your branch, preliminary_results.md in commit 09cb6ef as you see above.  We could link to it whenever preliminary results is active (or maybe if it has ever been active since the voting began?)

jacksonloper:
So is this ready to go?  Probably the preliminary info needs to be modified when we merge #1365

PR #1355, opened 2026-05-14
double voting fix

jacksonloper:
Merged Justin's double-voting-fix into the latest of main.  Don't squash this, we want to attribute Justin's great work in git blame.


ArendPeter (approved):
Thanks for updating!

PR #1361, opened 2026-05-15
fix 409s in admin panel

jacksonloper:
Helps ensures that users don't silently overwrite their own work due to weirdness in optimistic concurrency in frontend.  Done in part by insisting users only change one thing at a time on the admin panel (no crazy mind-bendiness where typical activity results in two POSTs in-flight at the same time).


jacksonloper:
Ah it reveals troubles.  The end-to-end test fails because of a real race condition between "set [max] number of rankings" the actual number itself being set.  They are separately debounced for some reason.

packages/frontend/src/components/Election/ElectionInFlightGate.tsx
```diff
@@ -0,0 +1,91 @@
...
+ * because every nav control inside the admin layout sits inside the gate.
+ */
+const GRACE_MS = 500;
+
+export default function ElectionInFlightGate({ children }: { children: ReactNode }) {
+    const { inFlight } = useElection();
+    const ref = useRef<HTMLDivElement | null>(null);
+    const [showOverlay, setShowOverlay] = useState(false);
+
+    useEffect(() => {
```
ArendPeter, line 21: Could this useEffect functions be combined since they're all triggering on inFlight?

packages/frontend/src/hooks/useOptimisticToggle.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useRef, useState } from "react";
+
+/**
+ * Optimistic toggle/select state synced with a backend resource.
+ *
+ * The setter (1) updates local state immediately so the MUI Switch/Radio
+ * animates without waiting on the network, (2) fires `commit` right away —
+ * no debounce — so the in-flight gate sees the write instantly, and
```
ArendPeter, line 8: Could you elaborate on why debounce is removed?

packages/frontend/src/components/Election/ElectionInFlightGate.tsx
```diff
@@ -0,0 +1,91 @@
+import { ReactNode, useEffect, useRef, useState } from 'react';
+import { Box, CircularProgress } from '@mui/material';
+import useElection from '../ElectionContextProvider';
+
+/**
+ * Disables interaction with its children while an admin write is in flight.
```
ArendPeter, line 6: I'm a little concerned about how this pattern would feel for the user. I'll check out the code to try for myself.

Stepping back, I worry that we're making the UX worse for 100% of users in exchange for solving race conditions that only would have impacted a small percentage of users.

ArendPeter: Thinking out loud, still looking for a way to solve the race conditions without hurting UX. This proposal is probably overcomplicating things but I wanted to write it out anyway.

**Optimistic Concurrency**: (the implementation in this PR) New requests pass an expected_update_date and then they only get written if the expected_update_date matches the version on the server. Therefore there can only be 1 request in flight at a time, since we need an update to resolve so we know the next expected_update_date. 

**Committer Context**:  Context of on the device id, and the client side timestamp for when the change was made. This is generated by the client, and stored in the database alongside the election.

**Naive Concurrency**: Only write if the committer_context.timestamp is more recent than the commiter_context in the database. This isn't as robust as optimistic concurrency since it doesn't scale for multiple timezones, and when there's multiple committers there's risk of overwriting each others changes. BUT the naive approach can handle stacked requests all coming from the same source. 

**Proposal**: If the db_entry.committer_context.device_id matches req.committer_context.device_id then use the naive approach. Otherwise use optimistic concurrency. 

**Motivation**: When many changes are coming from the a single client and they arrive out of order, the server can keep them from overwriting each other via the committer_context.timestamp. And if multiple clients are writing to the same data, the stricter optimistic concurrency will ensures that they aren't overwriting each other. This keeps the UX smooth while also addressing concurrency.

ArendPeter (approved):

ArendPeter:
image

Some styling notes
 - Can you "Election Support Email" font color consistent with the other fields?
 - Make the text input line wider
 - Make the text input line clickable (also opening the popup)
 - Make the text edit buttons darker, currently they appear disabled

I understand that the popup is there to avoid concurrency weirdness, but people will still want to click the text input, so if clicking the text input also triggered the popup that would make the experience intuitive.

packages/frontend/src/components/Election/ElectionInFlightGate.tsx
ArendPeter: Just tested it out, it feels bad if I'm spamming, but as long as I slow down a tad it's really good. 

That said, I still think it's worth thinking about the committer context pattern. Especially since it could help with race conditions on the E2E tests. If we decide to do it, we could either update the current PR, or implement it as a follow up

packages/frontend/src/hooks/useOptimisticToggle.ts
jacksonloper: Because as long as anything is in flight, I don't want the user editing anything *else*.  Because if they do we have to either merge or drop their other edits.  So the faster we get it done the more likely it is that they won't notice.

packages/frontend/src/components/Election/ElectionInFlightGate.tsx
jacksonloper: yep.  will fix.

packages/frontend/src/components/Election/ElectionInFlightGate.tsx
jacksonloper: We could also serialize edits on the frontend.

jacksonloper:
I think I addressed the stylin' notes.  Also made rank a buttongroup instead of radio (it was already forced to 3-10 anyway).

Screenshot 2026-06-05 at 2 28 12 PM

..... edit ...

Er.  After reading this some more, I see that the idea that we "don't" set max rank is not very useful for user.  If its not set... its 6.  So we should just always have this as a buttongroup instead of a toggle at all?  

Right now if we changed the REACT_APP_DEFAULT_BALLOT_RANKS environment variable, everybody's elections would I think spontaneously change, which seems... like not what we want.  I think we want the max ranks finalized when the election finalizes.

ah, and relatedly, the dedicated voter can POST any number of rankings (even if the frontend doesn't allow it) if the max rankings isn't set in the elections object.  that means that if you really had a lot of candidates you could actually do better in an STV/PR sort of situation because you could give the whole matrix and your ballot would never get exhausted. 

so.

I think max ranks should be finalized when election finalizes.

ArendPeter:
> Er. After reading this some more, I see that the idea that we "don't" set max rank is not very useful for user. If its not set... its 6. So we should just always have this as a buttongroup instead of a toggle at all?
> 
> Right now if we changed the REACT_APP_DEFAULT_BALLOT_RANKS environment variable, everybody's elections would I think spontaneously change, which seems... like not what we want. I think we want the max ranks finalized when the election finalizes.

Good points, I agree it makes sense to remove the toggle. The default can just be the initial value for new elections
 
> ah, and relatedly, the dedicated voter can POST any number of rankings (even if the frontend doesn't allow it) if the max rankings isn't set in the elections object. that means that if you really had a lot of candidates you could actually do better in an STV/PR sort of situation because you could give the whole matrix and your ballot would never get exhausted.
> 
> so.
> 
> I think max ranks should be finalized when election finalizes.

Makes sense as well. We should validate the number of rankings given on the backend, and also have it be finalized once the election is published. In the future this could be changed via a breakglass flow but that's not a high priority

ArendPeter (approved):
This PR looks good! Sorry for the delay

There's a few open questions, but I'm happy to leave it to your discretion which ones you'd like to address in this PR. 

Feel free to merge whenever you're ready :)

jacksonloper:
Max rankings stuff now looks like this:

Screenshot 2026-06-16 at 2 06 53 PM

ArendPeter (approved):
Awesome, thanks for all your work!

PR #1364, opened 2026-05-18
make STAR ballots more accessible

jacksonloper:
The semantics are generally radio groups, not buttons.  This PR makes that clear and reduces number of tabs a keyboard user must press.


jacksonloper:
@copilot+claude-sonnet-4.6 fix --- its just b/c 

```
await page.getByRole('button', { name: `Score ${vote.candidateName} ${vote.value}` }).click();
```

uses the old semantics

packages/frontend/src/components/Election/Voting/GenericBallotView/BubbleGrid.tsx
```diff
@@ -13,68 +13,333 @@ interface BubbleGridProps {
...
+      ? 'Score'
+      : 'Rank';
+
+  const semantic: BubbleSemantic = RADIO_PER_ROW_METHODS.has(votingMethod)
+    ? 'radio-per-row'
+    : votingMethod === 'Plurality'
+    ? 'radio-grid'
+    : votingMethod === 'Approval'
+    ? 'checkbox'
+    : 'pressed';
```
ArendPeter: what's pressed used for? I'm not sure how this line is reachable

ArendPeter (approved):
Thank you! This is a tedious change that we've been needing to do for a while.

The business logic looks correct to my eye, but it's a little hard to think through all the functionality cases, so I'll trust that the e2e tests are catching this.

ArendPeter:
Speaking of E2E tests, looks like they were failing. I restarted it in case it was a inconsistent error. Once e2e passes, feel free to merge

packages/frontend/src/components/Election/Voting/GenericBallotView/BubbleGrid.tsx
jacksonloper: Its a fallback in case someone adds a new valid voting method, leaving the buttons as simply pressed or not.  But maybe we'll just make it raise an error instead of having a fallback.  Let's do that.

PR #1365, opened 2026-05-18
Public audit log for finalized elections

jacksonloper:
## Summary

New `GET /Election/:id/history` returns a public, append-only event log of everything observable about an election since it was finalized. The point is that admins **cannot** suppress it — close/reopen cycles, preliminary-results toggles, ballot uploads and break-glass voter-ID reveals are all on the record.

Scope note: this PR previously also carried a `strict_ballot_privacy` election setting and a change to `ElectionRolls.update` error handling. **Both have been removed** so this is only the audit log. They can return as their own PRs.

### Events

Exactly six types:

| Event | Timestamp | Notes |
|---|---|---|
| `state_change` | exact | finalize, open, close, reopen, archive |
| `preliminary_results_change` | exact | |
| `ballots_milestone` | UTC day | ladder `{1, 5, 10, 25, 50, 100, 250, 500, 1000, 2500, …}` |
| `ballots_edited_milestone` | UTC day | finer ladder `{1, 2, 5, 10, 25, 50, 100, …}` |
| `upload_ballots` | UTC day | **new** — exact batch size |
| `voter_id_revealed` | UTC day | from the break-glass roll-history `action_type` |

Admin actions keep precise timestamps; ballot- and voter-related events are day-level so per-voter timing can't be cross-referenced against traffic logs to deanonymize. Filtering against the finalize cutoff uses raw ms, so precision is only lost in what's emitted, never in ordering.

### Changes since the last review of this branch

- **`upload_ballots` added.** Derived from `submitted_via_admin` entries in ballot history, clustered with a 5-minute gap so one bulk upload reads as one event rather than N. Batch size is reported exactly — it describes the admin's action, not any individual voter. Actions are deduped per `(ballot_id, timestamp)` because history is re-persisted on every ballot version.
- **`finalization_summary` removed.** The transition into `finalized` is now itself the first event (filter moved `>` → `>=`), so the log starts at finalization instead of summarizing it.
- **`rolls_milestone` removed.**
- **Day timestamps truncate instead of rounding to nearest.** Previously an event at 6pm was reported as the *following* day. The frontend also now renders these in UTC — `toLocaleDateString()` was shifting a UTC midnight backward a day in any negative-offset timezone.
- **`HistoryEvent` is now properly discriminated.** The two milestone types shared a single union member via `MilestoneType`, which broke `Extract<>` narrowing for consumers.
- **Cast-vs-edit split no longer depends on DB row order** — versions are grouped by `ballot_id` and the earliest is the cast.

### Testing

Derivation logic is extracted into a pure `buildHistory(electionRows, ballotRows, rollHeadRows)`. The mock `ServiceLocator` has no `database()`, so the controller can't be driven through the route in tests; the pure function is tested directly instead.

- [x] `npm test -w @equal-vote/star-vote-backend` — 22 suites, 156 tests passing, including 15 new in `electionHistory.test.ts`
- [x] `npm run build` for backend, frontend and shared; `swagger.json` regenerated and matches
- [ ] Manual smoke: finalize, cast, upload ballots, hit `/API/Election/:id/history`
- [ ] Playwright regression sweep on admin pages

Dev elections now get synthetic ballot history (wizardstar's last 2 ballots as an admin upload) so the view has an `upload_ballots` event to render locally.

### Assumptions verified against the models

These are load-bearing; each was checked rather than assumed:

- `electionDB` / `ballotDB` are append-only (old row `head=false`, new row inserted). Had they updated in place, `state_change` and edit milestones would silently emit almost nothing.
- Finalization **hard**-deletes all ballot rows, so draft-mode test ballots can't inflate post-finalize milestone counts.
- `getElectionByID` uses `executeTakeFirstOrThrow`, so a nonexistent ID 404s rather than 500ing on `req.election.election_id`.
- Ballot `history` is persisted by `makeSubmitBallotsQuery`, so `upload_ballots` works end to end.
- `update_date` is `varchar` ms-since-epoch — lexicographic `orderBy` is correct for equal-length 13-digit values.
- i18next resolves `{{count}}` without `_one`/`_other` variants (same pattern as `results_ext.column_distribution_num_avg`); passing `count` otherwise triggers plural lookup and would have silently broken three descriptions.

## Open questions

**1. The endpoint is fully public and ignores `public_results`.** It passes through `getElectionByID` and `electionSpecificAuth`, neither of which restricts access. So anyone holding an election ID can read turnout milestones and state changes **even when the admin has deliberately hidden preliminary results**.

This is arguably the whole point — an audit log admins can suppress is worthless — but it is a disclosure they cannot opt out of, on a system where hiding preliminary results is an explicitly supported setting. Worth a deliberate decision now rather than discovering it post-launch. Options: leave as is, gate ballot-count milestones behind `public_results`, or coarsen the ladder further while the election is open.

**2. Should `upload_ballots` report an exact count?** Every other ballot figure is laddered. The argument for exact is that batch size describes an admin action, not a voter; the argument against is consistency, and that a batch of 1 is a fairly precise signal.

**3. Minor:** `election_history.empty` is now unreachable — the finalize `state_change` always passes the `>=` filter, so `events` is never empty.

**4. Minor:** `ElectionHistoryView` renders "This election has not been finalized yet" for *any* fetch failure, including a 500 or network error. Pre-existing on this branch; would mislead during an outage.

Happy to address 3 and 4 in this PR if wanted.

## Notes

- `voter_id_revealed` events come from the head roll row's `history` JSON. If a future bug overwrites `history`, reveals would disappear from the log — worth a follow-up if it becomes a concern.
- Race/settings/`ballot_updates` mutations are already locked by `editElection`'s draft-state guard, so no audit event is needed for those.


packages/frontend/src/hooks/useAPI.ts
```diff
@@ -21,6 +21,23 @@ export const useElectionExists = (electionID: string | undefined) => {
...
+    | { type: 'preliminary_results_change'; timestamp: string; to: boolean }
+    | { type: 'ballots_milestone'; timestamp: string; count: number }
+    | { type: 'rolls_milestone'; timestamp: string; count: number }
+    | { type: 'ballots_edited_milestone'; timestamp: string; count: number }
+    | { type: 'voter_id_revealed'; timestamp: string };
+
+export const useGetElectionHistory = (electionID: string | undefined) => {
+    return useFetch<undefined, {
+        election_id: string,
+        finalized_at: string,
```
ArendPeter, line 35: Why is there a specific finalized_at field? Wouldn't this correspond to the timestamp on the finalization_summary with in the events?

packages/frontend/src/i18n/en.yaml
```diff
@@ -14,6 +14,39 @@ election_home:
   view_results: View Results
   drafted: This election is still being drafted
   archived: This election has been archived
+  strict_ballot_privacy: Results are hidden while voting is open, and closing the election is final.
```
ArendPeter: I feel mixed on the strict_ballot_privacy implementation. I think admins should always have the option to toggle open/close. Even if they think they won't need it, there's many unforseen circumstances that could cause them to need it

ArendPeter (changes requested):
Excited for the election history, but I'd like to talk about strict mode more. Could you do the following

1. Hide election history behind a beta flag for now so we can test it out first. 
2. Remove strict mode for now. We can chat about it, in stand up, and then bring it back in it's own PR

packages/frontend/src/i18n/en.yaml
ArendPeter: This could be something for a future PR, but it would nice to allow the admin to specify a reason for making certain changes so it doesn't look to scary to the voter

jacksonloper:
Per our discussion yesterday, plan going forward here is: 
* let admins leave note (also let them leave note for break glass voter id reveal), admin page has link for audit log, voter page doesn’t
* Make timing full-optional and changeable after finalizing

jacksonloper:
Merged latest `main` (191 commits behind). Three conflicts, all resolved:

- **`swagger.json`** — `main` now gitignores it (generated at build time), so accepted the deletion. The route’s swagger-jsdoc comment still feeds `generate:openapi`.
- **`Election.tsx`** — `main` wrapped the routes in `ElectionInFlightGate` while this branch added the `/history` route. Took `main`’s structure and added the route into it.
- **`ElectionHistoryView.tsx`** — `main`’s MUI upgrade dropped direct `display`/`justifyContent` props on `Box`. Moved them into `sx`; this was the last such instance in the codebase. This one was a *semantic* break that only surfaced on `tsc`, not a textual conflict.

Re-verified after merging, since `main` moved a lot: `electionDB`/`ballotDB` are still append-only, `getElectionByID` still throws (so missing IDs 404 rather than 500), finalization still hard-deletes ballots, and the `submitted_via_admin` / `VOTER_ID_REVEALED` action types are unchanged. All still hold.

Backend 22 suites / 155 tests passing (`main` deleted `register.test.ts` and added `anonymizedBallots.test.ts`); backend, frontend and shared all build.

One thing I deliberately left out: `npm install` rewrote the lockfile’s `optionalDependencies` from `"*"` to `"latest"`, because `main`’s committed `package-lock.json` has drifted from its `package.json`. That is a real pre-existing drift but unrelated to this PR, so I reverted it to keep the diff focused — worth a separate fix.

packages/frontend/src/components/Election/ElectionHistoryView.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { HistoryEvent, useGetElectionHistory } from "../../hooks/useAPI";
+
+const chipColor = (type: HistoryEvent['type']): "success" | "error" | "warning" | "info" | "default" => {
+    switch (type) {
+        case 'state_change': return 'info';
+        case 'preliminary_results_change': return 'warning';
+        case 'ballots_milestone': return 'success';
+        case 'upload_ballots': return 'info';
+        case 'ballots_edited_milestone': return 'warning';
+        case 'voter_id_revealed': return 'error';
```
ArendPeter: I think this should be warning, especially since we eventually want to show this to voters we don't want it to be over alarming when it's usually being revealed for a good reason

ArendPeter: Based on my experience so far, I think we'd expect any election with >100 voters to have at least a couple people with email issues, so these events will be expected for most larger elections

packages/frontend/src/components/Election/ElectionHistoryView.tsx
```diff
@@ -0,0 +1,140 @@
...
+                {data && data.events.length === 0 && (
+                    <Typography color="text.secondary">{t('election_history.empty')}</Typography>
+                )}
+
+                {data && data.events.length > 0 && (
+                    <>
+                        <Typography variant="caption" sx={{ display: 'block', mb: 2, color: 'text.secondary', fontStyle: 'italic' }}>
+                            {t('election_history.ladder_note')}
+                        </Typography>
+                        <TableContainer component={Paper} variant="outlined">
```
ArendPeter: Could this use EnhancedTable https://github.com/Equal-Vote/bettervoting/blob/main/packages/frontend/src/components/EnhancedTable.tsx ? That would be more consistent with the other tables on the site

packages/frontend/src/components/Election/Results/ViewElectionResults.tsx
```diff
@@ -112,6 +112,13 @@ const ViewElectionResults = () => {
             <a href="https://www.equal.vote/donate">
               {t("ballot_submitted.donate")}
             </a>
+            {election.state !== 'draft' && (
```
ArendPeter: I think it's a bit awkward coming after the donate link. Maybe it can be a button in the row above along with Download and Share Election? That's pretty much our row of election-wide actions (non-race specific)

ArendPeter: I'd also be more comfortable adding a feature flag check here. It would be a flags.isSet("AUDIT_LOG"). I expect we'll iterate more on the presentation of the information, and I want to give us a chance to access the /history link directly before making voters directly aware of it. Happy to discuss the value of feature flags more in stand up

https://github.com/Equal-Vote/bettervoting/blob/8d2b3f9a1dd41c15fcc1fe3ea47e5a2d6de29887/packages/frontend/src/components/Header.tsx#L232

ArendPeter (changes requested):
Excellent work! I particularly appreciated the detailed synthetic tests. I do think I'd like to have a feature flag for the button on the results page, but if that's done then I'd be happy for us to merge as and leave any other comments for a future PR.

packages/frontend/src/components/Election/Results/ViewElectionResults.tsx
jacksonloper: Great.  So /history will be live for all elections but nobody will know about it until we feel happy with it.

jacksonloper:
All three addressed in 2a2b49d2.

**Feature flag** — `flags.isSet("AUDIT_LOG")` now gates the results-page link, and the button moved into the election-wide actions row with Download and Share Election. Added `AUDIT_LOG` to `flagDefinitions` and `FF_AUDIT_LOG: 'false'` to `SharedConfig`, so it is off everywhere until switched on. To confirm your read: `/history` and its API stay live for every finalized election regardless of the flag — the flag only controls whether voters are pointed at it.

**EnhancedTable** — swapped in, which deleted most of the bespoke view. Event/description/when are now entries in the shared head cell pool.

**Reveal chip** — no longer red. Worth explaining where this landed, since it is not where you commented: EnhancedTable already auto-renders a chip for any group-filtered column and colours it via `getStateColor`, so the local `chipColor` function that your comment pointed at no longer exists. The audit log labels get cases in `getStateColor` instead, and none of them map to red — `Voter ID Revealed` is orange, the same colour `closed` uses with a comment noting it isn't a bad thing. That seemed like the right precedent for your point that these are expected on any election of size.

Two consequences of adopting EnhancedTable that you should sanity check, because both are visible to voters:

1. **Event chip labels are plain English, not translated.** EnhancedTable keys both its group filters and its chip colours off the formatted cell value, so a translated label would break filtering and colouring. Every other group-filtered column on the site works this way, so this is consistent — but it does mean the six event labels are not translatable until that architecture changes. The descriptions are still translated.

2. **The timestamp column is UTC**, rendered as `2026-01-05` / `2026-01-05 14:32` and headed "When (UTC)". EnhancedTable sorts on the formatted cell value, and a locale-formatted date neither sorts correctly nor round-trips through `Date()` outside en-US — so a browser in, say, de-DE would have silently mis-ordered the log. ISO-ish is the safe choice, and arguably right for an audit log, but it is less friendly than the previous localised date. Happy to revisit if you would rather have localised display and I will find another way to keep the sort honest.

Also note the table now defaults to newest-first. `EnhancedTable` hardcodes descending for any `defaultSortBy` other than `email`, so getting oldest-first would mean changing shared component behaviour — left alone deliberately.

jacksonloper:
Correcting and expanding **consequence 2** from my previous comment — it was vague, and wrong in two specific ways.

### Why the date format is a correctness question at all

Adopting `EnhancedTable` made the "When" column **sortable**. The old bespoke table wasn't — it just rendered the backend's chronological order, so the display format was purely cosmetic.

`EnhancedTable` sorts on the **formatted cell text**. `descendingComparator` only ever reads `row[orderBy]`, which is the string the formatter produced; the unformatted `raw` object is attached to every row by `formatTableData` but sorting never looks at it. So whatever we display *is* the sort key.

There are two ways it can interpret that string, and with a localised date **both are broken**.

### Without `isDate` — plain lexicographic on the displayed text

I previously said this only breaks outside en-US. That was wrong; it breaks in en-US too:

```
TRUE ORDER            finalized -> 1st ballot -> ballots reach 25 -> admin upload -> closed
lexicographic en-US   finalized -> 1st ballot -> admin upload -> closed -> ballots reach 25
```

`"2/3/2026"` sorts after `"10/2/2026"`, because `"2" > "1"`.

### With `isDate: true` — it calls `new Date()` on the displayed text

I said these wouldn't "round-trip". That undersells it: it does not throw or produce `Invalid Date`. It silently returns the **wrong date**, swapping day and month.

| locale | rendered | `new Date()` gives |
|---|---|---|
| en-US | `1/5/2026` | 2026-01-05 ✓ |
| pt-BR | `05/01/2026` | 2026-05-01 ✗ |
| es-ES | `5/1/2026` | 2026-05-01 ✗ |
| pl-PL | `5.01.2026` | 2026-05-01 ✗ |

`es`, `pl` and `pt-BR` are all shipped locales in this repo, so this is not hypothetical. What a pt-BR viewer would have seen:

```
TRUE ORDER            finalized -> 1st ballot -> ballots reach 25 -> admin upload -> closed
isDate pt-BR          admin upload -> ballots reach 25 -> finalized -> 1st ballot -> closed
```

An audit log that silently reorders itself according to the viewer's locale is worse than one that looks plain — it tells a different story about the same election, and there is nothing on screen to suggest anything is wrong.

### What shipped, and what it costs

The column renders `2026-01-05` / `2026-01-05 14:32` under the header "When (UTC)", where lexicographic order **is** chronological order in every locale.

The cost is real and I don't want to gloss it: it is less friendly than a localised date, and it shows UTC rather than the viewer's own timezone.

### If you'd rather have localised display

The fix belongs in `EnhancedTable`, not in this view: let a head cell supply a **sort key separate from its display value**. The `raw` row is already threaded through `formatTableData`, so it is a small change. The column could then render `t('listed_datetime', ...)` in local time while sorting on the ISO string — and it would fix the same latent problem for `create_date`, `update_date`, `start_time` and `end_time`, which all currently rely on `new Date()` parsing a localised string.

Happy to do that here or leave it as a follow-up — your call.

---

Also correcting **consequence 1**, which I overstated. Chip **colours** are fine in every locale: no group-filtered column has ever produced a translated value (`election_state` passes the raw `'draft'`, `has_voted` passes hardcoded `'Voted'`, and so on), so `getStateColor` always receives the key it expects. And the loss of translation isn't a live regression either — `es`, `pl` and `pt-BR` contain no `election_history` keys at all, so those chips were already falling back to English before this change.

The genuine limitation is that `EnhancedTable` derives both the label and the colour from a single value. `ElectionStateChip` already shows the fix — translate the label, pass the raw value to `makeChipStyle` — which would make every chip column on the site translatable, not just this one. Also worth flagging separately: `es`, `pl` and `pt-BR` are ~340 lines against `en.yaml`'s 1430, and none of them define `election_state`, so a fair amount of the site is falling back to English regardless.

ArendPeter (approved):
Looks good, thank you for the updates, and your patience!

PR #1376, opened 2026-05-21
update node react MUI

jacksonloper:
Dependabot upgrades will be less likely to explode if our node isn't years out of date.


ArendPeter (approved):
Looks good, thanks for doing the upgrade!

Issue #1090, opened 2025-11-13
'Equal Opposition' (0-score) vote mislabeled as 'Abstained' in UI and data export is ambiguous


masiarek:
Bug: 'Equal Opposition' (0-score) vote mislabeled as 'Abstained' in UI and data export is ambiguous

- election: https://bettervoting.com/jfrk9t
- video: https://youtu.be/nTlRqzVXlQY 
- doc - BV655 - “equal opposition” vote - the “Same-Score Ballots” : https://docs.google.com/document/d/1LPOOVztBjNdCTG9nQ2aLalD9UGEwTwqjUg2qbzBOddY/edit?tab=t.0

##two races:

<img width="822" height="249" alt="Image" src="https://github.com/user-attachments/assets/eb877bbe-03d2-4a3a-9bae-53fcd35b9975" />

### Race 1
<img width="623" height="624" alt="Image" src="https://github.com/user-attachments/assets/a5787d14-2c61-42ca-af92-0390574b1349" />

### Race 2

<img width="762" height="814" alt="Image" src="https://github.com/user-attachments/assets/a7a35305-4a19-406d-b7aa-fcd37029562d" />

## Incorrect confirmation window
<img width="739" height="482" alt="Image" src="https://github.com/user-attachments/assets/71a6735e-5c62-4106-a041-489a6d6c7be4" />

## Incorrect Ballot Download:

<img width="818" height="341" alt="Image" src="https://github.com/user-attachments/assets/c4ec3356-79e8-4abf-8da8-c6e40911ed03" />

## JSON is correct:

<img width="534" height="911" alt="Image" src="https://github.com/user-attachments/assets/7ca6b454-4a0d-40e5-a700-c80b8ad07648" />

Description: There are two related bugs in how the system handles 'Equal Opposition' (explicit 0-score) votes versus 'Abstain' (blank) votes.

UI Bug: The confirmation screen incorrectly labels an explicit 0-score vote as "Abstained."

Data Export Bug: The downloaded ballot CSV uses empty cells for abstentions, which is ambiguous. The data should clearly distinguish between an explicit 0 score and a NULL (no-vote) entry.

## Bug 1: UI Confirmation Screen
Steps to Reproduce:

On the ballot (see image_0ba9ad.jpg):

In Race 1 ("BV655 - “equal opposition” vote..."), give both options a score of 0 stars.

In Race 2 ("Abstain Vote"), leave both options blank.

Click "Submit."

### Actual Behavior:
[...]

Issue #791, opened 2025-02-15
Test BV20 - CSV Ballots Download - missing abstention


masiarek:
BV20 - Basic - 3 candidates - Approval - Abstention: https://bettervoting.com/bp7kwg/
More details: https://docs.google.com/document/d/1aDPqJUrQjf1_tyqd2T8UjxPqhO-iuiAnl7foeqzxqiI/edit?tab=t.0

<img width="785" alt="Image" src="https://github.com/user-attachments/assets/5343f198-c75c-4a30-a14a-092363fd262a" />

<img width="542" alt="Image" src="https://github.com/user-attachments/assets/a7943d1f-a86a-4ba1-aab8-bc17c5af76dc" />

We are missing a clear indicator flagging this ballot/race as "abstain".

masiarek:
Related issue - incorrect number of voters - compare to: https://github.com/Equal-Vote/star-server/issues/788

PR #1419, opened 2026-07-05
Clean up the ballot-data export (JSON v2 + CSV Raw/Official)

### LINKED ISSUE #1160: Add Dual Export Option (Processed vs. Raw Data) - Download ballots
### Overview

Download ballots - add "Dual Export Options" (Processed vs. Raw Data)

### Action Items

See details:

https://docs.google.com/document/d/19yfr5ktSECPalPoWEXi2h325ZDSQjKm0iF0QQP-B8ew/edit?tab=t.0

## 1. The Core Problem
In Star Voting, a voter can leave a candidate blank.
Database Reality (NULL): The database should store this as NULL to truthfully record that the voter did not interact with that candidate.
Mathematical Reality (0): To count the votes, Star Voting rules treat a blank as a 0 (no support).
The Conflict: If we only export raw data (NULL), calculation scripts crash. If we only export processed data (0), we destroy the audit trail of who actually abstained.

## 2. The Solution: Two Download Options
We should provide two separate buttons for downloading ballot data:
### Option A: "Official Count Data" (Recommended for most users)
What it does: Replaces all NULL (empty) values with 0.
Use Case: For users who want to copy-paste the data into a Star Voting calculator, Excel, or Python script to find the winner immediately.
Why: Removes friction. Users don't have to "clean" the data themselves; it just works.
### Option B: "Raw Audit Data" (For verification)
What it does: Exports the data exactly as stored, keeping NULL values empty (e.g., 5,,4).
Use Case: For auditors, data analysts, or election officials who need to verify exactly what the voter saw and did. It allows them to distinguish between a voter who hates a candidate (explicit 0) and a voter who ignored a candidate (implicit 0).

## 3. Example of the Difference
User Action: Voter scores "Bird" a 5 and leaves "Cat" blank.

Download A (Official Count / Processed):

Bird,Cat
5,0
(The software can instantly sum 5+0. The election runs smoothly.)

Download B (Raw / Audit):
Bird,Cat
5,

(The auditor sees the comma and knows: "The voter skipped Cat." This preserves the truth.)

## Summary for the Developer
"Please implement two export formats. The Standard Export must sanitize the data by filling all nulls with 0s so that 3rd-party counting tools don't crash. The Raw Export must preserve nulls so we maintain a perfect audit trail of voter intent."

# Before clean_up.csv

[after_cleanup.csv](https://github.com/user-attachments/files/24069836/after_cleanup.csv)
[before_cleanup.csv](https://github.com/user-attachments/files/24069835/before_cleanup.csv)

Image

# After clean_up.csv

Image

### Resources/Instructions

_No response_

- [x] Role
- [x] Complexity

masiarek:
## Description

Cleans up the results **ballot-data export** (`BallotDataExport.tsx`) — both the JSON and CSV downloads, which previously leaked the tabulator's internal object shape and couldn't distinguish a blank from an explicit `0`.

### JSON — versioned v2 format

The old JSON download was `JSON.stringify({ Election, Ballots, Results })` of the raw in-memory objects, which leaked:
- every candidate's O(n²) `votesPreferredOver` / `winsAgainst` maps (keyed by UUID, incl. self-vs-self), duplicated across `elected`/`tied`/`other`/`summaryData`
- `Results` in camelCase while `Election`/`Ballots` were snake_case
- inconsistent timestamps (ISO `create_date` vs epoch-ms-string `update_date`)

New `packages/shared/src/utils/exportFormat.ts` (`buildElectionExport`) produces a versioned v2 export: candidates listed once, a deduped `pairwise` matrix (self-pairs removed, keyed by name), `{id,name}` refs, snake_case throughout, ISO-8601 timestamps, `format_version: 2`. **Ballot scores are preserved verbatim, including `score: null`** — a blank (didn't score) stays distinct from an explicit `0` and from a NOTA vote. Measured 69–74% of legacy size on real exports.

### CSV — Raw vs Official Count

The single CSV mode wrote a blank as an empty cell, ambiguous vs an explicit `0`. Split into two menu items:
- **Download CSV (Official Count)** — fills blanks with `0` so Excel/Python counting tools don't choke.
- **Download CSV (Raw / Audit)** — keeps blanks empty (`5,,4`), preserving the audit trail.

Transform-only — no change to the tabulation engine, the results endpoint, or what the UI renders. The export is write-only (nothing re-imports it). Unit-tested in `packages/backend/src/test/exportFormat.test.ts` (incl. a regression test that `null` ≠ `0` ≠ dropped).

No visible page change; the outputs are the downloaded files. Download menu now offers: **CSV (Official Count)**, **CSV (Raw / Audit)**, **JSON**.

## Related Issues

- Addresses the export-shape cleanup in #1420.
- Implements the dual Raw-vs-Official export from #1160 (both JSON-null-preservation and the CSV split) — proposing `closes #1160`.
- Fixes the export-ambiguity half of #1090 and #791 (blank vs explicit `0` now distinguishable).

### DIFF
--- a/packages/backend/src/test/exportFormat.test.ts
+++ b/packages/backend/src/test/exportFormat.test.ts
@@ -1,0 +1,113 @@
+import { Star } from '../Tabulators/Star';
+import { mapMethodInputs } from './TestHelper';
+import { buildElectionExport } from '@equal-vote/star-vote-shared/utils/exportFormat';
+
+// Builds a realistic STAR result and checks the v2 export shape.
+describe('buildElectionExport (v2 export format)', () => {
+    const names = ['Allison', 'Bill', 'Carmen'];
+    const votes = [
+        [5, 2, 1],
+        [5, 3, 0],
+        [4, 0, 5],
+        [3, 5, 5],
+        [5, 1, 4],
+    ];
+    const results: any = Star(...mapMethodInputs(names, votes), 1);
+
+    // Minimal election whose race candidates line up with the tabulator names.
+    const election: any = {
+        election_id: 'testelec',
+        title: 'Export format test',
+        state: 'closed',
+        create_date: '2026-07-05T13:41:39.541Z',
+        update_date: '1783258899541', // epoch-ms string — must normalize to ISO
+        owner_id: 'owner',
+        audit_ids: null,
+        public_archive_id: null,
+        head: true,
+        races: [
+            {
+                race_id: 'r1',
+                title: 'Race 1',
+                voting_method: 'STAR',
+                num_winners: 1,
+                candidates: names.map((n) => ({ candidate_id: n, candidate_name: n })),
+            },
+        ],
+        settings: { public_results: true },
+    };
+
+    const ballots: any = votes.map((v, i) => ({
+        ballot_id: `b${i}`,
+        election_id: 'testelec',
+        precinct: null,
+        votes: [
+            {
+                race_id: 'r1',
+                scores: names.map((n, j) => ({ candidate_id: n, score: v[j] })),
+            },
+        ],
+    }));
+
+    const out: any = buildElectionExport(election, ballots, [results]);
+    const json = JSON.stringify(out);
+
+    test('is versioned and self-describing', () => {
+        expect(out.format).toBe('bettervoting-export');
+        expect(out.format_version).toBe(2);
+        expect(typeof out.exported_at).toBe('string');
+        expect(Number.isNaN(Date.parse(out.exported_at))).toBe(false);
+    });
+
+    test('drops the O(n^2) pairwise maps from candidate objects', () => {
+        expect(json).not.toContain('votesPreferredOver');
+        expect(json).not.toContain('winsAgainst');
+    });
+
+    test('pairwise matrix is deduped with no self-pairs', () => {
+        const pw = out.results[0].pairwise;
+        for (const name of names) {
+            expect(pw[name]).toBeDefined();
+            expect(pw[name][name]).toBeUndefined(); // no self-vs-self
+        }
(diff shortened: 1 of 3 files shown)


masiarek:
Verified end-to-end in the Docker stack (`docker compose up`) against a live STAR election, on a ballot deliberately carrying **both** a blank and an explicit `0`, plus an all-blank and an all-`5` ballot:

**JSON** — preserves intent: the untouched candidate is `"score": null`, the explicit-0 candidate is `"score": 0`. Format is v2 (snake_case, deduped `pairwise`, `{id,name}` refs, `format_version: 2`).

**CSV (Official Count)** — blanks filled with `0` (`3,0,2,0`), so Excel/Python can sum directly.

**CSV (Raw / Audit)** — blanks kept empty (`3,,2,0`), the explicit `0` stays `0`; an all-blank ballot is an all-empty row. Audit trail intact.

The full E2E suite also passed on this branch (19/19).

One extra commit here (`516fbd69`): the CSV download previously set React state and then synchronously clicked a hidden `<CSVLink>`, so each click emitted the *previous* click's data/mode (the Raw item produced the Official file, filename always `-official`). Rebuilt the CSV to serialize + download via a Blob (same approach as the JSON export), so the mode is applied on the spot — no react-csv, no stale state. Verified: the Raw and Official files now differ correctly and the filename carries the mode.

masiarek:
Follow-on: #1432 (tie-break transparency) proposes extending this same `buildElectionExport` with per-round `logs` + `tieBreakType`, so a tied result's resolution path is auditable in the JSON. Out of scope for this PR — noting it here since #1432's JSON ask builds directly on the v2 export scaffolding added here.

jacksonloper:
Good questions!  We will merge your excellent bug fixes #1428 and the rest needs some contemplation.

PR #1423, opened 2026-07-07
Enable brotli/gzip response compression

jacksonloper:
## Problem

API responses are served with no `Content-Encoding` at all, even when the client advertises `accept-encoding: gzip, deflate, br, zstd`. Verified directly against prod:

```
$ curl -sD - https://bettervoting.com/API/Election/v3hw4g/anonymizedBallots -H 'accept-encoding: gzip, deflate, br, zstd' -o /dev/null
HTTP/2 200
content-type: application/json; charset=utf-8
content-length: 3312        <- raw bytes, no content-encoding header
```

For large elections the `anonymizedBallots` / results payloads run to many megabytes of highly repetitive JSON, all going over the wire uncompressed.

## Fix

Add the standard [`compression`](https://github.com/expressjs/compression) middleware to `makeApp()`. As of 1.8.x it negotiates brotli (preferred, quality 4 by default — appropriate for dynamic responses) with gzip fallback, and skips responses under 1 KB. Repetitive ballot JSON typically compresses ~10-20x.

This doesn't change backend memory usage (the JSON is still fully serialized before compression), just transfer size and time.

## Testing

- `npm run build -w @equal-vote/star-vote-backend` passes
- `npm test -w @equal-vote/star-vote-backend` — 135 tests pass (suites exercise `makeApp()` with the middleware in place)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1424, opened 2026-07-07
Stream anonymizedBallots from a DB cursor in random order

jacksonloper:
## Problem

`GET /API/Election/:id/anonymizedBallots` loads every ballot into memory (`selectAll()`), then `.filter().map()`, then `secureShuffle`, then `res.json` — the full dataset is held roughly 3x at peak. For large elections (e.g. ~31k ballots ≈ 14 MB of JSON, ~30–50 MB of heap to build) this is one of the endpoints that OOMKills the pod and 502s users. Worse, V8 never returns the heap high-water mark to the OS, so one big request permanently inflates the pod's baseline.

## Change

Stream the ballots row-by-row from a server-side Postgres cursor straight to the response, so peak memory is O(batch) instead of O(ballots):

- **`ServiceLocator`**: pass `pg-cursor` to the Kysely `PostgresDialect` (this is what enables kysely's `.stream()`; no behavior change for existing queries).
- **`BallotsDB.streamSubmittedBallotsByElectionID`**: new model method — cursor-backed (`.stream(500)`), filters `head = true AND status = 'submitted'` in SQL.
- **Random order moves into the database**: the response order must not reveal ballot submission order (an admin could zip it against roll timestamps to deanonymize votes), but you can't shuffle a stream. The query orders by `gen_random_uuid()`, which is backed by Postgres's strong RNG (`pg_strong_random`), so this keeps a CSPRNG-grade shuffle — it just happens in Postgres memory instead of Node. ⚠️ Flagging explicitly since it swaps the privacy primitive from `secureShuffle` to the database RNG.
- **Controller**: serializes `{"ballots":[...]}` incrementally via an async generator (batched into ~64 KB chunks) piped to the response with `stream.pipeline`, which handles backpressure (a slow client throttles the cursor rather than ballooning buffers) and destroys the cursor if the client disconnects mid-stream.
- **`AnonymizedBallot.precinct`** typed `string | null` to match what the API has always returned (`"precinct": null`).
- Mock ballot store now sets `head = true` on insert like the real store, and implements the new stream method.

## Response shape

Unchanged: `{"ballots": [{ballot_id, election_id, precinct, votes}, ...]}`. The response is chunked (no `Content-Length`), which clients don't observe through `fetch`/axios.

## Error semantics tradeoff

Since streaming sends the `200` and body incrementally, a mid-stream DB failure can no longer produce a 500 — the connection is aborted instead (client sees a network error rather than truncated-but-parseable JSON). Auth/permission errors are unaffected (checked before any bytes are written).

## Testing

- New `anonymizedBallots.test.ts`: returns all submitted ballots anonymized (exact key set per ballot), empty election, and 401 for non-public open elections — 139 backend tests pass.
- `npm run build` passes for shared, backend, and frontend.

## Relationship to other work

Independent of #1423 (response compression) — they compose (compression middleware compresses streamed chunks as they flow) and any merge conflict is trivial. The new cursor plumbing (`pg-cursor` dialect option + the model stream method) is also the foundation the planned compact tabulation loading (docs/ram-reduction.md §2.7) can reuse.


ArendPeter (approved):
Looks great! I hadn't used the streaming libraries before, fascinating stuff

PR #1426, opened 2026-07-08
Fix flaky Firefox e2e: make /new_election scroll to wizard reliable

### LINKED ISSUE #1393: Flaky Firefox e2e: "Create Election" nav to /new_election doesn't reliably scroll wizard into viewport
## Summary

On Firefox, the `create-election.spec.ts` Playwright tests intermittently fail because clicking the **Create Election** nav (which now targets `/new_election`) does not reliably scroll the wizard into the viewport in time. Chromium passes consistently.

Observed on branch `JacksonLoper/updatenode` (Node 24 / React 19 / MUI 9 upgrade), but the root cause is a navigation change, not the dependency upgrade.

## Symptoms

Running `cd testing && npx playwright test --reporter=list`:

- `create-election.spec.ts:45 › From About Us, Election, More than one race, BetterVoting-managed voter IDs` — **fails consistently on Firefox** at:
  ```
  await expect(electionButton).toBeInViewport({timeout: 2000});
  ```
  The `Election` radio resolves in the DOM but reports `viewport ratio 0` (rendered below the fold, not scrolled into view within 2s).
- `create-election.spec.ts:118 › Poll, Multi Race, One vote per Device` — **flaky on Firefox** (fails some runs, passes others), same area.

Chromium variants of both tests pass. All other tests (17) pass.

## Root cause

The "Create Election" nav target was changed from the URL anchor `/#wizard` to the route `/new_election` (`packages/frontend/src/components/Header.tsx`, `createWizardNav`).

- `/#wizard` is a native URL fragment → the browser jumps to the wizard **instantly and reliably**.
- `/new_election` has no anchor. Instead the scroll is done in JS (`packages/frontend/src/components/LandingPage.tsx`):
  ```ts
  useEffect(() => {
      if (checkUrl.pathname === "/new_election") {
          scrollToElement(document.querySelector(`.wizard`))
      }
  }, [checkUrl]);
  ```
  and `scrollToElement` (`packages/frontend/src/components/util.tsx`) waits **250ms** then does `window.scrollTo({ behavior: 'smooth' })`.

On Firefox, the 250ms delay + smooth-scroll animation — compounded by landing-page reflow as the carousel / featured elections load and push `.wizard` further down — means the wizard isn't in the viewport within the test's 2000ms assertion. Chromium is fast enough to pass.

This is both a **test-reliability regression** and a minor **UX regression** (on Firefox the "Create Election" nav may land short of the wizard).

## Possible fixes

1. **Harden the test (lowest risk):** raise the `toBeInViewport` timeout on `create-election.spec.ts:52` (e.g. 2000 → 6000ms). The subsequent `.check()` auto-scrolls the radio into view, so only the early viewport assertion is failing.
2. **Make the `/new_election` scroll reliable:** scroll instantly (`behavior: 'auto'`) for this path, and/or scroll after layout settles (e.g. once `.wizard` is laid out) rather than a fixed 250ms timeout, so it isn't beaten by post-load reflow.
3. Revert the nav back to the native `/#wizard` anchor if the route change isn't otherwise needed.

## Repro

```
cd testing
npx playwright test tests/create-election.spec.ts --project=firefox --reporter=list
```

jacksonloper:
Fixes #1393

## Problem

On Firefox, `create-election.spec.ts` intermittently failed after the Create Election nav was changed from the native `/#wizard` anchor to the `/new_election` route, where the scroll to the wizard is done in JS. Two distinct races (both rooted in `scrollToElement`'s fixed 250ms delay + smooth scroll):

1. **`/new_election` full page load** (test :45): the one-shot delayed smooth scroll could land short of the wizard when async landing-page content (carousel, images) reflowed the page, and in Firefox dev mode the page takes ~1.5s+ to render at all, leaving no headroom in the test's 2s `toBeInViewport` budget.
2. **Landing-page nav click** (test :118): the click handler scheduled the scroll 250ms later, so Playwright would validate the `Poll` radio as stable and click exactly as the scroll kicked in — the click missed (`Clicking the checkbox did not change its state`).

## Fix

This combines options 1 and 2 from the issue:

- **`LandingPage.tsx`**: on `/new_election`, position at the wizard instantly (like the old native `/#wizard` anchor did) and keep re-aligning via a `ResizeObserver` on `document.body` while content above the wizard loads and reflows. Re-aligning stops once layout settles (2s) or as soon as the user scrolls/interacts, so we never fight the user. This also fixes the minor UX regression where the nav could land short of the wizard on Firefox.
- **`Header.tsx` / `util.tsx`**: the in-page wizard-nav click keeps its original feel (250ms pause, then smooth scroll). The race is closed instead with a new `cancelOnUserInput` option on `scrollToElement`: if the user (or the test) interacts during the delay, the still-pending scroll is dropped, so it can never yank the page out from under a later click. `scrollToElement`'s defaults are unchanged for all other callers.
- **`create-election.spec.ts`**: raise the `toBeInViewport` timeout from 2s to 10s for the test that does a full page load of `/new_election` (matching the 10s timeouts already used elsewhere in the spec).

## Testing

Against the local dev stack:

- `npx playwright test tests/create-election.spec.ts --project=firefox --repeat-each=3` — 15/15 pass (previously test :45 failed most runs and :118 failed ~half)
- Tests :6 and :118 repeated 5x each on Firefox — the wizard-scroll flakes are gone
- Full spec on Chromium + Firefox and the full suite — all pass
- Instrumented Firefox runs confirm the page scrolls to the wizard as soon as it renders and stays there (no snap-back)
- `npm run build -w @equal-vote/star-vote-frontend` passes; eslint reports no new issues on the touched files

https://claude.ai/code/session_01KJ6tU9igwGYGrFGumeCA7d

### DIFF
--- a/packages/frontend/src/components/Header.tsx
+++ b/packages/frontend/src/components/Header.tsx
@@ -20,9 +20,9 @@ export const createWizardNav = (heading: string, isLandingPage: boolean) => {
     return isLandingPage ?
         {
             text: heading,
-            // no delay: the wizard is already rendered, and starting the scroll
-            // immediately avoids it kicking in mid-interaction with the wizard
-            onClick: () => scrollToElement(document.querySelector(`.wizard`), { delay: 0 }),
+            // cancelOnUserInput so the delayed scroll can't kick in mid-interaction
+            // if something else gets clicked before it starts
+            onClick: () => scrollToElement(document.querySelector(`.wizard`), { cancelOnUserInput: true }),
         } : {
             text: heading,
             href: '/new_election',
@@ -201,7 +201,7 @@ const Header = () => {
                                 {t('nav.your_account')}
                             </MenuItem>
                             {isLandingPage && 
-                            <MenuItem component={Link} onClick={() => scrollToElement(document.querySelector(`.wizard`), { delay: 0 })}>
+                            <MenuItem component={Link} onClick={() => scrollToElement(document.querySelector(`.wizard`), { cancelOnUserInput: true })}>
                                 {t('nav.new_election')}
                             </MenuItem>
                             }
--- a/packages/frontend/src/components/Header.tsx
+++ b/packages/frontend/src/components/Header.tsx
@@ -20,7 +20,9 @@ export const createWizardNav = (heading: string, isLandingPage: boolean) => {
     return isLandingPage ?
         {
             text: heading,
-            onClick: () => scrollToElement(document.querySelector(`.wizard`)),
+            // no delay: the wizard is already rendered, and starting the scroll
+            // immediately avoids it kicking in mid-interaction with the wizard
+            onClick: () => scrollToElement(document.querySelector(`.wizard`), { delay: 0 }),
         } : {
             text: heading,
             href: '/new_election',
@@ -199,7 +201,7 @@ const Header = () => {
                                 {t('nav.your_account')}
                             </MenuItem>
                             {isLandingPage && 
-                            <MenuItem component={Link} onClick={() => scrollToElement(document.querySelector(`.wizard`))}>
+                            <MenuItem component={Link} onClick={() => scrollToElement(document.querySelector(`.wizard`), { delay: 0 })}>
                                 {t('nav.new_election')}
                             </MenuItem>
                             }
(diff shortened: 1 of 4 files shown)


ArendPeter (approved):
Thanks for fixing this! This was a tricky one

PR #1427, opened 2026-07-09
Fix STV crash when the final winner's surplus has no candidates left

jacksonloper:
## Summary

`GET /API/ElectionResult/:id` returns a 500 for any STV race where the last remaining candidate wins the final seat via quota with surplus votes. This is live in prod: election `h93tm4` has been 500ing on its results page (`TypeError: Reduce of empty array with no initial value`).

## Mechanics

With h93tm4's 8 ballots (3 candidates, 1 winner, quota 5):

1. First preferences split 3/3/2 — nobody reaches quota, lowest is eliminated.
2. Redistribution leaves 4/4 — still under quota, another elimination.
3. The last remaining candidate now holds all 8 votes ≥ quota. The winner branch `shift()`s them out of `remainingCandidates` (leaving it empty) and calls `distributeVotes()` on the surplus ballots.
4. `distributeVotes` reduces over the empty `remainingCandidates` with no initial value → `TypeError` → 500.

Plain IRV can't hit this because its winner branch rebuilds the candidate pool instead of shifting.

## Fix

Guard in `distributeVotes`: when no candidates remain to receive votes, all ballots are exhausted — splice them out and return, matching the existing exhausted-ballot behavior.

## Test

New `stvSurplus.test.ts` supertest reproduces the h93tm4 ballot pattern end-to-end (create election → submit 8 ballots → fetch results). It failed with the 500 before the fix and passes after. Full backend suite passes (21 suites, 138 tests). Also verified the fix against the actual h93tm4 production ballots via the anonymized ballots endpoint.

https://claude.ai/code/session_01S14fmwy1Hsw57x83HzSYmr


packages/backend/src/test/stvSurplus.test.ts
```diff
@@ -0,0 +1,101 @@
...
+                    scores: rankings.map((rank, i) => ({ candidate_id: String(i), score: rank })),
+                }],
+            } as NewBallot;
+            const response = await th.submitBallot(election.election_id, ballot, testInputs.user1token);
+            expect(response.statusCode).toBe(200);
+        }
+        th.testComplete();
+    });
+
+    test("Results tabulate when the last remaining candidate wins with surplus", async () => {
```
ArendPeter: Could this be moved to STV.test.ts under tabulators?

ArendPeter (changes requested):
Thanks for addressing this so quickly, I left a comment to keep the convention consistent

packages/backend/src/test/stvSurplus.test.ts
jacksonloper: Yes!  Smaller tests for the win.

ArendPeter (approved):
Thanks for the update!

PR #1428, opened 2026-07-09
Fix CSV ballot export: escape fields and download directly via Blob

jacksonloper:
## Description

Extracts the CSV bug fixes from #1419 (authored by @masiarek — the commit carries his authorship) without the new JSON v2 export format or the Raw/Official menu split, plus a fix for the Download menu collapsing on first open.

**Bug 1 — corrupt CSV when a field contains a comma.** The CSV download rendered through a hidden `<CSVLink>` with `enclosingCharacter={''}`, which disabled quoting entirely: any candidate name, race title, or precinct containing a comma (or quote/newline) shifted every subsequent column. Verified end-to-end on the old code: a candidate named `Smith, John` splits into two header columns, so every later score sits under the wrong candidate. Fields are now escaped per RFC 4180 (quoted only when needed).

**Bug 2 — first "Download CSV" click downloads nothing.** The old path set React state and then synchronously clicked the hidden link — which doesn't exist until the *next* render. Verified end-to-end: the first click produces no download, and a second click delivers whatever the first click staged. The CSV string is now built directly and downloaded via a Blob, the same approach the JSON download already used.

**Bug 3 — the Download menu collapses while ballots load.** The component rendered two separate MenuButtons (loading vs. loaded). The first click opens the loading menu and starts the ballots fetch; when the ballots arrive, that subtree unmounts and the open menu vanishes, forcing a second click on the button. Now a single MenuButton stays mounted and only its items swap, so the menu opened by the first click stays up and CSV/JSON appear in place. Net effect with Bug 2's fix: **one click on Download, one click on the format, one correct file** — previously it took four clicks and produced a corrupt file.

Output is otherwise unchanged: same columns, same filename, blanks (not-scored) stay blank, and the JSON download still emits the existing `{ Election, Ballots, Results }` shape. Removes the now-unused `react-csv` dependency.

No visible page change — the Download menu offers the same items (CSV / JSON); it now stays open while ballots load, and the downloaded CSV bytes differ when fields need escaping.

## Related Issues

- Extracted from #1419


ArendPeter (approved):
Looks great, thanks @jacksonloper and @masiarek !

Issue #1390, opened 2026-06-02
STAR Equal Support graph doesn't always match the total count


ArendPeter:
https://bettervoting.com/gvdy42/results

On that election we saw 124 votes of equal support, but only 5 votes in the graph. Maybe blanks and zeros were different

masiarek:
similar issue: https://github.com/Equal-Vote/bettervoting/issues/1384

jacksonloper:
I believe we're talking about this graph, which is indeed mysterious

<img width="879" height="536" alt="Image" src="https://github.com/user-attachments/assets/892e06d7-f6c8-4332-b34b-798a7de0ae4d" />

jacksonloper:
Would be closed by #1431

PR #1431, opened 2026-07-10
Fix Distribution of Equal Support graph dropping skipped-candidate ballots

jacksonloper:
## Problem

To address #1390 

On https://bettervoting.com/gvdy42/results (2026 CA Governor, 302 voters), the "Distribution of Equal Support" graph in Stats for Nerds shows a single 100% bar computed from only 5 ballots, even though the runoff chart reports 124 equal-support ballots (41%).

Anonymized ballots store `null` for candidates a voter skipped. `STAREqualPreferencesWidget` compared raw scores:

- 117 ballots that skipped **both** finalists passed `null == null` and incremented a phantom `equalPreferences[null]` bucket, so they silently vanished from the chart.
- 2 ballots with `null` for one finalist and an explicit 0 for the other failed `null == 0` and were excluded, even though the tabulator counts them as equal support (it treats null as 0).

## Fix

Coerce skipped scores to 0 when reading front-runner scores, matching the tabulator:

```ts
getEntry(scores, frontRunners[i].id, 'candidate_id').score ?? 0
```

## Verification

- `npm run build -w @equal-vote/star-vote-frontend` passes.
- Ran the fixed widget logic (including the `ballotsForRace` filtering) in Node against the live gvdy42 anonymized ballots: it now yields `[124, 0, 0, 0, 0, 0]` across 0–5⭐, matching the tabulator's 124 equal-support count exactly.


ArendPeter (approved):
Looks good!

PR #1441, opened 2026-07-15
Tweak results barchart to make the heights more consistent

ArendPeter:
## Description
2 changes
- Remove the space after the ⭐ to avoid excess line wrapping (see screenshot)
- Add height padding to account for the legend height. This ensures that the individual bar heights are consistent across the scoring and runoff round

**Before**
image

**After**
image


packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
```diff
@@ -136,7 +138,7 @@ const [rawNumbers, setRawNumbers] = useState(false);
   const adjustedAxisWidth = useTickMargin ? axisWidth + tickOffset : axisWidth;
   return (
     <div style={{width:'100%', maxWidth: '600px', marginLeft: 'auto', marginRight: 'auto'}} onMouseEnter={() => setRawNumbers(true)} onMouseLeave={() => setRawNumbers(false)}>
-    <ResponsiveContainer width="90%" height={50 * data.length} style={maxBarSize ?
+    <ResponsiveContainer width="90%" height={50 * data.length + (majorityLegend == undefined ? 0 : 32)} style={maxBarSize ?
```
ArendPeter, line 141: The 32 came from inspecting the UI. I feel a bit bad about the magic number, but it works and I don't think there's a way to infer it. Hopefully the purpose is reasonably clear from the ternary condition.

Long term I have a dream of migrating from recharts to writing the bar charts ourselves. They're simple enough that I think the added flexibility would be worth it

jacksonloper:
Chatbot agrees with you @ArendPeter "The short version: you're using an SVG charting library to draw horizontal bars, which is the one chart form that CSS does natively and better. Nearly every hack in ResultsBarChart.tsx is Recharts fighting you, and this PR is patching a symptom rather than the cause."

packages/frontend/src/components/Election/Results/components/ResultsBarChart.tsx
JonBlauvelt: Why home-rolled over another, better chart lib?

JonBlauvelt (approved):

PR #1445, opened 2026-07-17
Add yearly, per-method breakdown to GlobalElectionStats API

### LINKED ISSUE #1443: Add yearly, per-method breakdown to GlobalElectionStats API
## Problem Statement

BetterVoting/Equal Vote want to show stakeholders how the platform has grown over time — specifically, how many elections have been created and how many votes have been cast, and whether adoption of individual voting methods (STAR, STAR-PR, Approval, Ranked Robin, RCV, Plurality, STV) is trending up or down year over year. The current `GlobalElectionStats` API only reports all-time cumulative totals (plus an all-time per-method breakdown) — there's no way to see the trend, so a stakeholder looking at the numbers today can't tell whether growth is accelerating, flat, or whether a particular voting method is gaining or losing traction. Separately, the REST form of this endpoint (`GET /API/GlobalElectionStats`) is currently broken — it responds with an unresolved promise instead of the computed stats — so any consumer that isn't the landing page's socket connection gets no usable data from it today.

## Solution

Extend the existing `GlobalElectionStats` API (both the `GET /API/GlobalElectionStats` REST route and the `updated_stats` socket event it shares logic with) in place, rather than adding a new endpoint. Add a `by_year` field: an object keyed by calendar year (as a string, e.g. `"2024"`), where each entry mirrors the existing top-level shape for elections-created and votes-cast, including the full per-voting-method breakdown, using the same field-naming convention already in use at the top level (e.g. `star_elections`, `star_votes`, ..., `multi_method_elections`, `multi_method_votes`). The set of years is contiguous, spanning from the earliest year with any qualifying election through the current (possibly still in-progress) year, with years that had no qualifying activity present and zero-filled. Legacy/classic pre-migration counts (which have no per-election date or per-method data behind them) are not represented in `by_year` at all and remain solely in the existing top-level `legacy_elections`/`legacy_votes` fields. As part of this change, also fix the un-awaited REST route so it actually returns data.

## User Stories

1. As a dashboard maintainer, I want a `by_year` breakdown of elections created and votes cast, so that I can chart BetterVoting/Equal Vote's growth over time.
2. As an Equal Vote stakeholder, I want to see year-over-year adoption of each voting method (STAR, STAR-PR, Approval, Ranked Robin, RCV, Plurality, STV), so that I can tell whether a particular method's usage is growing or shrinking.
3. As a dashboard maintainer, I want years with zero qualifying elections to still appear as zero-filled entries in a contiguous range, so that I can build a chart without having to backfill missing years myself.
4. As a dashboard maintainer, I want the current, still-in-progress year included in the breakdown, so that I can show up-to-date momentum rather than waiting until year-end.
5. As a dashboard maintainer, I want the earliest year in the range determined dynamically from real election data, so that the range doesn't need to be manually maintained as a hardcoded constant as the platform ages.
6. As a dashboard maintainer, I want legacy/classic-site lump-sum counts kept out of the yearly breakdown, so that a single bulk historical number isn't misrepresented as if it happened in one specific year.
7. As an API consumer developer, I want `GET /API/GlobalElectionStats` to actually return data instead of an empty object, so that I can integrate with the endpoint directly over HTTP without depending on the socket-push path.
8. As a BetterVoting engineer, I want the yearly breakdown to apply the same filtering rules as the existing top-level totals (excluding dev/test elections, excluding bulk-imported "prior_election" elections, requiring at least 2 votes), so that yearly and all-time numbers stay internally consistent.
9. As a BetterVoting engineer, I want the year-bucketing/aggregation logic isolated as a pure, independently callable function, so that multi-year and edge-case behavior can be verified without seeding a real database with backdated elections.
10. As a frontend maintainer, I want the TypeScript type describing this API's response kept in sync with its actual shape, so a future consumer of the typed hook isn't misled by a stale type.
11. As an API documentation reader, I want the Swagger docs for `GET /API/GlobalElectionStats` updated to describe `by_year`, so the documented contract matches the real one.
12. As a dashboard maintainer, I want each year's per-method fields to use the same key naming as the top-level fields, so I can reuse the same parsing logic for both the all-time totals and any given year.
13. As a BetterVoting engineer, I want an election whose races mix more than one voting method to continue being bucketed under `multi_method` within its year, consistent with existing top-level behavior, so method-specific growth numbers aren't polluted by mixed elections.
14. As a dashboard maintainer, I want `by_year`'s keys to be plain year strings with no synthetic non-year keys, so I can safely sort/iterate the keys as years without special-casing.
15. As an Equal Vote board member reviewing growth reports, I want confidence that yearly totals reflect genuinely qualifying elections (not test/dev noise or near-empty elections), so the reported growth trend is credible.

## Implementation Decisions

- Extend the existing `GlobalElectionStats` response type — shared by the `GET /API/GlobalElectionStats` route and the `updated_stats` socket event — with a new `by_year` field. No new endpoint or API version is introduced.
- `by_year` is an object keyed by stringified calendar year (e.g. `"2024"`). Each value mirrors the top-level fields for elections/votes and the full per-method breakdown, but does **not** include `legacy_elections`/`legacy_votes` — those stay exclusively at the top level, unchanged.
- An election's year is derived from its `create_date`. Its voting method is derived the same way it is today (parsed from its races), with elections whose races span more than one method bucketed as `multi_method`, matching current top-level behavior.
- The same three inclusion filters already applied to the top-level totals apply per-election before it's bucketed into a year: exclude dev/test-owned elections (unless explicitly whitelisted), exclude elections sourced from a prior election (bulk-imported legacy data), and require at least 2 votes.
- The year range in `by_year` is contiguous: from the minimum year among qualifying elections' `create_date` through the current calendar year (computed dynamically, not hardcoded). Any year in that range with zero qualifying elections is still present, fully zero-filled.
- The current, in-progress calendar year is included even though it's partial — the API doesn't withhold or flag partial-year data; a consumer decides how to present it.
- Legacy/classic pre-migration counts (currently sourced from environment variables, representing activity before this system's data existed) are unchanged and are not represented anywhere inside `by_year`.
- The year-bucketing/aggregation computation is extracted into its own function that takes already-fetched data (the election records and ballot counts the endpoint already fetches today) and returns the `by_year` object, decoupling it from the database-fetching code so it can be exercised independently.
- Fix the existing bug where the REST route responds before its own stats computation resolves, so `GET /API/GlobalElectionStats` returns real data. (The socket-push path used by the landing page already awaits correctly and is unaffected.)
- Update the frontend's currently-unused, stale TypeScript type for this endpoint's response so it matches the extended shape, including `by_year`. No behavioral or UI change to the landing page component that already consumes this data via the socket — it only reads `elections`/`votes` today and continues to do so.
- Update the Swagger/OpenAPI documentation for `GET /API/GlobalElectionStats` to describe `by_year`.

## Testing Decisions

- A good test here asserts on the shape and values of the computed stats object for known inputs — it should not depend on the internal structure of the aggregation loop, only on its observable output.
- The year-bucketing/aggregation function is extracted specifically to make it unit-testable: tests feed it fabricated election/ballot-count data (arbitrary `create_date` years, voting-method mixes, vote counts) and assert on the resulting `by_year` object. Coverage should include: an election in a past year vs. one in the current (partial) year, a multi-method election bucketed as `multi_method`, the min-2-votes/dev-user/prior_election filters correctly excluding within a year, and the zero-filled contiguous year range including at least one gap year with no activity.
- This is a new, lower-level unit-testing seam for this module. The existing convention across this repo's backend tests (`packages/backend/src/test/*.test.ts`, e.g. `createElection.test.ts`) is HTTP-level integration testing via `TestHelper`/`supertest` against a real test database. That seam was considered and ruled out for the year-bucketing behavior specifically: `create_date` is set server-side to "now" at creation time and can't be controlled through the API, so there's no way to backdate a test election and exercise multi-year bucketing through the HTTP seam alone.
- In addition to the unit tests, keep one HTTP-level integration test, following the existing `TestHelper` pattern, that hits `GET /API/GlobalElectionStats` for real. This confirms the endpoint returns data end-to-end (regression coverage for the `await` fix) and that an election created through the API is correctly reflected in the current year's bucket and the top-level totals.
- There is no prior test coverage for this endpoint today (`GlobalElectionStats` currently has zero tests), so there's no existing behavior to preserve beyond what's specified above.

## Out of Scope

- Any metric beyond elections-created and votes-cast (e.g. unique voters, ballot completion rates, election-state breakdowns) — explicitly deferred to a future change.
- Any change to how classic/legacy pre-migration counts are computed or sourced — they remain the existing environment-variable-based lump sum, untouched.
- Any change to the landing page's stats component or its socket-based consumption of this data — it continues to read only `elections`/`votes` and is unaffected by this change.
- Adding raw database access to `TestHelper` to support backdating elections for integration-style multi-year testing — explicitly ruled out in favor of the pure-function unit-testing seam described above.
- Any time granularity other than calendar year (e.g. monthly, weekly).
- Any new REST endpoint or API version — this extends the existing `GET /API/GlobalElectionStats` route and `updated_stats` socket event in place.

## Further Notes

- There is no existing frontend consumer of `by_year` or the per-method breakdown today — the real intended consumer is a separate, external dashboard repository that will call this API directly. This means there's no internal backwards-compatibility constraint shaping the response design above.
- The `GlobalElectionStats` type currently lives inline in the backend controller, with an independently-maintained (and currently stale) duplicate in the frontend's API hook file. This change keeps both in sync but doesn't restructure where these types live (e.g. moving either into the shared package).
- Timezone handling for converting `create_date` to a calendar year isn't specified beyond "pick one consistent convention" (e.g. UTC) — `create_date` is stored as an ISO timestamp string, so the implementer should be deliberate about this to avoid off-by-one-year edge cases for elections created near a year boundary.

ArendPeter:
## Summary
- Add `by_year` to the `GlobalElectionStats` response (shared by `GET /API/GlobalElectionStats` and the `updated_stats` socket event): a contiguous, zero-filled, per-calendar-year breakdown of elections/votes with the same per-method fields used at the top level (`star_elections`, `star_votes`, ..., `multi_method_elections`, `multi_method_votes`)
- Year bucketing/aggregation is extracted into a pure function so multi-year and edge-case behavior can be tested without seeding a real database
- Applies the same inclusion filters as the existing totals (exclude dev/test elections, exclude bulk-imported `prior_election` elections, require ≥2 votes); legacy/classic lump-sum counts stay top-level only, never in `by_year`
- Fixes `GET /API/GlobalElectionStats` responding before its stats promise resolved (the socket path was already correct)
- Updates the frontend's stale response type and the Swagger docs to describe `by_year`

Closes #1443

## Test plan
- [x] `npm test -w @equal-vote/star-vote-backend` — new coverage in `globalElectionStats.test.ts` for multi-year aggregation, zero-filled years, filter exclusions, and multi-method bucketing
- [x] Manually hit `GET /API/GlobalElectionStats` and confirm `by_year` is present and matches expectations
- [x] Confirm landing page (socket-driven `updated_stats`) still renders unchanged


jacksonloper (changes requested):
Maybe its out of scope for this PR but I think it would be a good place to fix the RAM issue here.  As is, we're still going to end up calling getBallotCountsForAllElections on every vote cast (and every homepage load once we put that back in).  We need to decouple how often these stats are read from how often they are computed.  I think this PR would be a reasonable place to do it.  

I would
* remove the castVoteController call sites (they're dead weight anyway right now b/c sockets are disabled)
* put in a module-level cache value `export const getCachedGlobalElectionStats = async...` that recalculates at most every 5 minutes.  
* then we can uncomment the global stats display from the landing page

PR #1446, opened 2026-07-17
Fix/publish now ownerless

clarissa-garcia:
**Problem**

Clicking “Publish Now” while creating an election without signing in leaves the temporary user with administrative access after the election is published. As a result, owner-only navigation—such as the voting page, admin home, and ballots links—remains visible even though an anonymously published election should not have an owner.

The publishing behavior is handled in the election creation flow, while ownership assumptions are also used across shared election types, validation, and backend controller checks.

**Cause**

The anonymous publish flow finalized the election without first clearing the temporary user’s owner_id. That meant the election continued to reference the temporary account created during election setup, so existing ownership checks still treated the anonymous publisher as an administrator.

Several parts of the application also assumed owner_id was always a string. Shared TypeScript interfaces and validation did not allow an ownerless value, while backend checks called string methods or compared owner IDs without first accounting for null. Simply clearing owner_id in the frontend would therefore create type, validation, and runtime problems elsewhere in the application.

**Fix**

Set owner_id to null when an unauthenticated user chooses “Publish Now,” so the finalized election no longer belongs to the temporary account.

Update the shared TypeScript election model and validation rules to allow owner_id to be nullable, then make backend ownership checks null-safe before performing string comparisons or temporary-user checks. Existing UI authorization behavior can then recognize that the election has no owner and omit the administrative sidebar. Signed-in publishing and elections with assigned owners remain unchanged.


packages/backend/src/Controllers/Election/getElectionsController.ts
```diff
@@ -112,7 +112,7 @@ const innerGetGlobalElectionStats = async (req: IRequest): Promise<GlobalElectio
     const electionMethodMap: Record<string, ElectionMethodKey> = {};
     electionRaces?.forEach(e => {
         const methods = new Set((e.races as Race[]).map(r => r.voting_method));
-        if(sharedConfig.DEV_USERS.includes(e.owner_id) && !sharedConfig.REAL_ELECTIONS_FROM_DEVS.includes(e.election_id)) {
+        if(e.owner_id == null || (sharedConfig.DEV_USERS.includes(e.owner_id) && !sharedConfig.REAL_ELECTIONS_FROM_DEVS.includes(e.election_id))) {
```
ArendPeter: I think it's okay to leave this off. The global election stats is meant to measure how much usage we're seeing across the site, and to filter out any "fake traffic" like test elections from devs. 

A election without owner_id (i.e. someone who created the election with the "publish now" flow) should still be counted toward the site usage

packages/frontend/src/components/ElectionForm/Wizard/Wizard.tsx
```diff
@@ -80,7 +80,7 @@ const Wizard = () => {
             submitTempID = makeID(ID_PREFIXES.VOTER, ID_LENGTHS.VOTER);
             setCookie('temp_id', submitTempID);
         }
-        election.owner_id = authSession.isLoggedIn() ? authSession.getIdField('sub') : submitTempID;
```
ArendPeter, line 83: I think this breaks the election claim flow. Can confirm that the following flow would still work?

1. Start as a logged out user
2. Fill out the election wizard
3. Select "See more options"
4. Complete the wizard
5. Sign in via the "Unlock more with a free bettervoting account!" popup
6. The election should be assigned now be assigned to the user. 

This flow is important for delaying account creation in the flow. We risk scaring off users if they need to create an account too early. Unfortunately this also makes our creation logic much more complicated to reason about 😅

ArendPeter: So we only want null to be set if the user selects "Publish Now" in the flow

ArendPeter (changes requested):
Thanks for taking this on Clarissa!

There's several edge cases to consider with our log in flow, but hopefully they shouldn't be too bad to fix. I'm really glad you're working on this, as it's been a bit, and this will reduce confusion for users without accounts.

packages/backend/src/Controllers/Election/getElectionsController.ts
clarissa-garcia: Updated. I removed the filtering of ownerless elections and kept a null guard before includes() so it type-checks with owner_id: string | null. Let me know if you had anything else in mind.

packages/frontend/src/components/ElectionForm/Wizard/Wizard.tsx
clarissa-garcia: I went through and verified that the new changes keep the flow, made only the "Publish Now" ownerless, and let the others have a temp id that can be changed to a new owner after logging in.

ArendPeter (approved):
Looks great, thanks Clarissa!

ArendPeter (changes requested):
Oops, actually looks like the end to end tests are failing. I'll double check just in case it's something unrelated

clarissa-garcia:
Thanks for checking Arend! I appreciate it. If you see anything else you think needs changing, let me know.

ArendPeter (approved):
Build works again (I didn't do anything, the node installer link was temporarily down for some reason)

PR #1448, opened 2026-07-22
Fix landing page/wizard header text color and favicon.ico regression

ArendPeter:
## Summary
- Landing page and election-creation wizard headings were rendering black text on dark/colored backgrounds (e.g. "Vote in a Featured Public Election"). MUI's `Typography` `color` prop stopped resolving dotted palette paths like `'darkShade.contrastText'` after the MUI v5→v9 upgrade — it now only maps top-level palette keys to `.main`. Moved these 12 usages into `sx={{ color: '...' }}`, which still resolves nested theme paths.
- The site favicon had reverted to the generic ICO instead of the intended PNG. Rsbuild auto-injects its own `<link rel="icon">` from the first of `favicon.ico`/`favicon.png`/`favicon.svg` it finds in `public/`, silently overriding the hand-written favicon link in `index.html` ever since the Vite→Rsbuild migration. Renamed the full-color icon to `favicon-prod.png` and made the grayscale icon the default `favicon.png` (used for local/dev builds and Netlify PR previews); `replace.sh` now copies `favicon-prod.png` over `favicon.png` only in the real production build, replacing the old sed-based string rewrite that Rsbuild's auto-injected tag was bypassing.
- Update the favicon graphics

## Test plan
- [x] `tsc --noEmit` passes
- [x] Built the frontend and confirmed both the hand-written and Rsbuild-injected `<link rel="icon">` tags resolve to the same `/favicon.png`
- [ ] Visually confirm the landing page section headings and wizard title render in the correct (white/contrast) color
- [ ] Confirm production build shows the full-color favicon and local/dev shows the grayscale one


packages/frontend/public/bettervoting-logo.pdn
ArendPeter, line 1: This is the save file from paint.net, my image editing tool. I figured it makes sense to have this in source control

SaraWolk (approved):
Looks good!

PR #1449, opened 2026-07-22
Phased lint/a11y/security linting, opt-in Husky hooks, secret scanning

ArendPeter:
## Summary

Adds automated feedback systems intended to catch mistakes (including AI-agent mistakes) before they land, without repeating a prior failed attempt to lint the whole codebase in one pass:

- **ESLint for backend and shared** — neither had any config before this. A shared base config (`eslint.base.mjs`, at the repo root) is extended by all three packages' own configs.
- **`eslint-plugin-jsx-a11y`** (frontend) — static accessibility checks.
- **`eslint-plugin-security`** (backend) — static security-footgun checks.
- **`eslint-plugin-jest`** (backend) — test-hygiene checks (`.only`, empty tests, etc.), scoped to `*.test.ts`.
- **Diff-scoped enforcement, not whole-repo** — `lint-staged` (Husky pre-commit) and a new `npm run lint:diff` (CI) only ever check changed files, never pre-existing code. Both are blocking. See `dev-docs/adr/0001`.
- **Husky hooks are on by default** — `pre-commit` runs `lint-staged`; `pre-push` runs the build and unit tests. E2E is left out of both (CI already covers it on every PR). See `dev-docs/adr/0002`.
- **gitleaks secret scanning is CI-only** — `gitleaks/gitleaks-action` runs as a blocking step on every PR and push. Deliberately not also wired into a local hook: that would require every contributor to install the `gitleaks` binary just to commit at all. (Also deliberately not an npm package — gitleaks has never published one; the packages that show up searching npm for "gitleaks" are unofficial single-maintainer wrappers around the binary and were not used.)
- CI (`node.js.yml`) now fetches full history (`fetch-depth: 0`) so the diff/secret-scan steps have a base commit to compare against, and adds blocking "Lint changed files" + "Scan changed files for secrets" steps on both the `pull_request` and `push` triggers.

## Test plan

- [x] Verified each of the three ESLint configs (`frontend`, `backend`, `shared`) loads and lints real files without crashing.
- [x] Verified `jsx-a11y`, `eslint-plugin-security`, and `eslint-plugin-jest` rules are actually active (checked via `--print-config` / targeted lint runs).
- [x] Verified `lint-staged` blocks and reverts on a deliberately introduced lint error in a staged file.
- [x] Verified `npm run lint:diff` against real historical diffs across all three packages, confirms non-zero exit on errors.
- [x] Verified `node.js.yml` YAML parses correctly.
- [ ] Confirm the CI run on this PR passes the new steps as expected.


ArendPeter:
The build is failing because gitleaks requires a license? I filled out the google form at https://gitleaks.io/, we'll see where it goes. If it takes a while for them to respond, then I may remove gitleaks from the PR

dev-docs/adr/0001-diff-scoped-lint-and-secret-enforcement.md
```diff
@@ -0,0 +1,12 @@
+# Diff-scoped lint and secret-scan enforcement, not whole-repo
```
ArendPeter, line 1: This get automatically added when using the /grill-with-docs skill

packages/frontend/eslint.config.js
```diff
@@ -1,16 +1,16 @@
-import js from "@eslint/js";
+import base from "../shared/eslint.base.mjs";
 import globals from "globals";
-import tseslint from "typescript-eslint";
```
ArendPeter, line 3: This has been moved to the shared lint config

dev-docs/adr/0001-diff-scoped-lint-and-secret-enforcement.md
ArendPeter: https://www.youtube.com/watch?v=6BB6exR8Zd8

eslint.base.mjs
JonBlauvelt, line 1: I'm not sure that it actually matters, but I wonder whether it makes more sense philosophically for the base config to exist in the root of the monorepo rather in the shared lib and then be/fe/sharedlib can all use/extend from there. then perhaps eslint can be configured to look to root dir first rather than an import.

scripts/lint-diff.mjs
JonBlauvelt, line 1: Alternatively:
1. We could just do one commit where we run the fixer and push that.
2. Whatever cannot be auto-fixed, we can just set the failure thresholds based off of that. 

This would also allow us to enforce standards going forward and as we resolve existing issues we could just lower the thresholds accordingly.

.gitignore
```diff
@@ -16,3 +16,8 @@
 
 # Generated build artifacts
 packages/backend/src/OpenApi/swagger.json
+
+# Husky hooks are opt-in — only the .sample files are committed. Activating a
```
JonBlauvelt: Why do we not want to have this enabled for folks by default?

JonBlauvelt (changes requested):
This is going to be awesome to finally have. I just added some thoughts that are not necessarily show stopping but lmk what you think.

eslint.base.mjs
ArendPeter: Makes sense, will fix!

scripts/lint-diff.mjs
ArendPeter: Also makes sense, but I think I'd prefer to keep that out of scope of this PR for now. I did make a issue and added it to prioritized backlog though: https://github.com/Equal-Vote/bettervoting/issues/1593

.gitignore
ArendPeter: Initially I thought I'd make it opt in since people may have different preferences, but now I agree it makes sense to have a default. I'm leaning towards having lint + gitleaks on pre-commit, and then run build + unit tests on pre-push. E2E tests can be left out of the hooks

JonBlauvelt (approved):

PR #1462, opened 2026-07-29
Let admins clear the voter roll of a draft election

### LINKED ISSUE #1363: Add endpoint to clear election rolls for draft elections
## Problem

In Manage Voters, the voter-access and voter-id/email radios are disabled once `electionRollData.length > 0`, even for elections still in `draft` state:

```ts
disabled={election.state !== 'draft' || electionRollData.length > 0}
```

If an admin has added a roll and then realizes they picked the wrong auth mode (e.g. picked email list but wanted id list, or vice versa), there's no in-app way to undo that without recreating the election from scratch — there's no DELETE route on the roll router at all (the model has `ElectionRolls.delete()`, but no controller/route exposes it).

## Proposal

Add a backend endpoint that, for `draft` elections only, deletes all rows in the election roll for that election. After it runs, the auth/access radios re-enable per the existing condition.

- Endpoint: `DELETE /API/Election/:id/rolls` — guarded by `canAddToElectionRoll` (same permission used to add voters).
- Refuses for any non-draft state, returns 409 or 400.
- Frontend: a "Clear all voters" button on the Manage Voters page (draft only), behind a `useConfirm` dialog with explicit copy ("this removes all N voters and lets you change voter auth settings — proceed?").

## Possibly related

Per-voter DELETE is also not exposed (model method exists, no route). Worth considering whether to add `DELETE /API/Election/:id/rolls/:voter_id` in the same PR, or split.  

## Proposal

Don't actually delete rows, just mark as old

jacksonloper:
Closes #1363

## Problem

Adding the first voter locks the voter-access and voter-id/email radios on Manage Voters:

```ts
disabled={election.state !== 'draft' || electionRollData.length > 0}
```

That lock is permanent even while the election is still a draft, so an admin who picked the wrong auth mode (email list when they wanted an ID list, or vice versa) had no in-app way to undo it short of recreating the election. There was no DELETE route on the roll router at all.

## What this does

Adds `DELETE /API/Election/:id/rolls`:

- Guarded by `canAddToElectionRoll` — the same permission used to add voters, on the theory that an admin who can create the lock should be able to undo it.
- Refused with a 400 for any non-draft election.
- **Archives rather than deletes**, per the last note on the issue. Every `head=true` row is demoted to `head=false`, so all the model's read paths (which already filter on `head`) stop seeing them and the roll reads as empty, but the rows and their `history` survive for auditing. `update_date` is untouched, so there's no collision with the `(election_id, voter_id, update_date)` primary key, and the partial unique index is on `(election_id, voter_id) WHERE head`, so the same voter_ids can be re-added afterwards.

Frontend gets a `useClearRolls` hook and a "Clear Voter List" button on Manage Voters, rendered only in draft with a non-empty roll and behind a confirm dialog that names the voter count. On success it refetches the roll, which drops the count to 0 and re-enables the radios through the existing condition.

The mock roll store now honors `head` in its reads and its conflict check, matching the real store.

## Testing

- New `clearElectionRoll.test.ts` — 9 tests: non-owner → 401, owner → 200 with the archived count, roll empty afterwards, no-op on an already-empty roll, the same voters re-added successfully, and 400 once the election is finalized.
- Full backend suite passes (22 suites / 149 tests), which is what confirms the mock's new `head` filters didn't disturb existing tests.
- Frontend `tsc --noEmit` is clean and the touched files add no lint errors.
- Added an E2E case to `election-with-rolls.spec.ts` that clears the roll and then flips to the other voter-ID scheme. **This one is unverified** — I didn't have a live stack to run Playwright against, so it needs a run in CI or locally before merge.

## Deliberately out of scope

The issue's "possibly related" per-voter `DELETE /rolls/:voter_id` — flagged there as maybe-split, so it's not in this PR. Happy to add it here if you'd rather have both at once.

Only `en.yaml` gained strings; the other locales don't carry the neighboring `add_first_voter_roll_confirm` keys either and fall back to English.

### DIFF
--- a/packages/backend/src/Controllers/Roll/clearElectionRollController.ts
+++ b/packages/backend/src/Controllers/Roll/clearElectionRollController.ts
@@ -1,0 +1,40 @@
+import ServiceLocator from "../../ServiceLocator";
+import Logger from "../../Services/Logging/Logger";
+import { permissions } from '@equal-vote/star-vote-shared/domain_model/permissions';
+import { expectPermission } from "../controllerUtils";
+import { BadRequest } from "@curveball/http-errors";
+import { IElectionRequest } from "../../IRequest";
+import { Response, NextFunction } from 'express';
+
+const ElectionRollModel = ServiceLocator.electionRollDb();
+
+const className = "VoterRolls.Controllers";
+
+// Draft-only escape hatch for admins who picked the wrong voter auth mode: once a roll exists
+// the auth/access radios lock, and there was no in-app way to undo that short of recreating the
+// election. Clearing archives the rows (head=false) rather than deleting them, so the audit
+// trail survives and the radios re-enable because the roll now reads as empty.
+const clearElectionRoll = async (req: IElectionRequest, res: Response, next: NextFunction) => {
+    expectPermission(req.user_auth.roles, permissions.canAddToElectionRoll)
+    Logger.info(req, `${className}.clearElectionRoll ${req.election.election_id}`);
+
+    if (req.election.state !== 'draft') {
+        Logger.info(req, `Refusing to clear roll, election state=${req.election.state}`);
+        throw new BadRequest("The voter list can only be cleared while the election is in draft")
+    }
+
+    const cleared = await ElectionRollModel.archiveRollsByElectionID(
+        req.election.election_id,
+        req,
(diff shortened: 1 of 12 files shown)


ArendPeter (approved):
Looks great, thank you for adding this!

PR #1463, opened 2026-07-29
Remove unsubscribe links from transactional email

jacksonloper:
Our ballot invites and receipts are transactional, but they were arriving with an "Unsubscribe From This List" footer. Turns out we ask for it.

Every outbound message was tagged with a SendGrid unsubscribe group via `SENDGRID_GROUP_ID` (`43160` in prod), and the html template hardcoded `<%asm_group_unsubscribe_url%>` in its footer. Raw source of a real receipt shows the group was doing three separate things:

1. `List-Unsubscribe` + `List-Unsubscribe-Post: List-Unsubscribe=One-Click` headers, which make Gmail render its own native unsubscribe button next to the sender — independent of anything in the body
2. substituting our footer tag in place in the html part
3. appending its own `Unsubscribe From This List … | Manage Email Preferences …` footer to the plain text part, which we never wrote

So removing just the template tag wouldn't have worked — the headers would have survived, and the html part would most likely have picked up an appended footer in place of the in-place one. This drops the group as well as the tag.

## Changes

- `EmailTemplates.ts` — remove the `asm` group from `emailSettings` and `<%asm_group_unsubscribe_url%>` from the template footer; fix the template-regeneration comment so the tag doesn't get pasted back
- `IEmail.ts` — drop the now-unused `asm?` field
- `sample.env`, `docker-compose.yml` — drop `SENDGRID_GROUP_ID`, nothing reads it now

## Behavior change worth flagging

Group suppressions apply only to messages sent *with* that group id, so this bypasses group 43160's suppression list wholesale rather than clearing it. Anyone who previously clicked unsubscribe starts receiving ballot invites and receipts again. For transactional election mail that seems right, but it's a real change and worth a look at the list size before merging. The entries aren't deleted — if the group is ever re-added with the same id, they all take effect again.

SendGrid's global unsubscribe list and bounce/spam suppressions are unaffected and still apply.

## Not needed

- **argocd** — `SENDGRID_GROUP_ID` in `star-server` values becomes inert; no deploy ordering concern, can be deleted whenever
- **webhook** — `sendGridWebhookController.ts` stores `event_type` generically, so `group_unsubscribe` events simply stop arriving and nothing downstream breaks

## Verification after deploy

`EMAIL_TEST_MODE` uses sandbox mode and won't produce a deliverable message, so this needs a real send: trigger a receipt from a test election and check the raw source for absence of `List-Unsubscribe` and a clean plain text part. If a footer still appears, account-level Subscription Tracking is enabled (Settings → Tracking) and needs turning off there. The html part of the sample message shows no appended footer, which suggests it's off, but that's an inference rather than a check.

Backend suite passes locally (21 suites, 140 tests).


packages/backend/src/Services/Email/EmailTemplates.ts
```diff
@@ -203,7 +204,7 @@ body, #bodyTable { background-color: rgb(244, 244, 244); }.mceText, .mcnTextCont
...
...padding-right:0;padding-left:0" align="center" valign="top"><table width="100%" style="border:0;border-radius:0;border-collapse:separate"><tbody><tr><td style="padding-left:16px;padding-right:16px;padding-top:12px;padding-bottom:12px" class="mceTextBlockContainer"><div data-block-id="8" class="mceText" id="dataBlockId-8" style="display:inline-block;width:100%"><p class="last-child"><%asm_group_unsubscribe_url%></p></div></td></tr></tbody></table></td></tr><tr><td class="mceLayoutContainer" align="center" valign="top"><table align="center" border="0" cellpadding="0" cellspacing="0" width="100%" role="presentation" data-block-id="-2"><tbody><tr class="mceRow"><td style="background-position:center;background-repeat:no-repeat;background-size:cover;padding-top:0px;padding-bottom:0px" valign="top"><table border="0" cellpadding="0" cellspacing="24" width="100%" role="presentation"><tbody></tbody></table></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table></td></tr></tbody></table><!--[if (gte mso 9)|(IE)]></td></tr></table><![endif]--></td></tr></tbody></table>
```
ArendPeter, line 206: I did a diff checker to see exactly what was changed and it's just ``<p class="last-child"><%asm_group_unsubscribe_url%></p>`` being removed (as expected)

ArendPeter (changes requested):
This looks good, however for the emails to be considered transactional I think https://github.com/Equal-Vote/bettervoting/issues/1219 needs to be done first. I've assigned that one to me and I'll try to prioritize it so that this PR can be unblocked

PR #1464, opened 2026-07-29
Stream ballots into a compact per-race projection for tabulation

### LINKED ISSUE #1425: Reduce tabulation memory: stream ballots from a cursor and project to compact tabulator input
## Background

Tabulating a large election currently loads every ballot as a full verbose object graph (~75–150 MB heap for a ~31k-ballot election) before handing it to the tabulators. This is one of the main drivers of star-server OOMKills: V8 grows the heap to serve one big tabulation and never returns it to the OS, permanently inflating the pod's baseline.

Tabulators (STAR/IRV/STV/etc.) are batch algorithms — they genuinely need the whole election at once, so the *algorithm* can't stream. But "the whole election" doesn't have to mean "30k verbose JSON ballot objects."

## Proposal

Stream ballots from a server-side Postgres cursor and project each row down to the minimal numeric form the tabulator needs (small int arrays per race), discarding the verbose row immediately. You still hold everything, but a *lean* everything — a few MB of typed arrays instead of a ~100 MB object graph.

The plumbing already exists after #1424:

- `pg-cursor` is wired into the Kysely dialect, so `.stream()` works on any query.
- `BallotsDB.streamSubmittedBallotsByElectionID()` streams head/submitted ballots from a cursor. This work likely wants a narrower-select variant (tabulation doesn't need `ballot_id`, `precinct`, timestamps, etc. — select only `votes`, and whatever filtering columns apply).

The new work is the projection layer: cursor row → per-race compact representation → tabulator input.

## Why this is its own issue (and needs a high review bar)

Unlike the endpoint streaming in #1424 (where a bug mis-serves a response — loud, shallow, recoverable), a bug in the projection layer **silently miscounts elections**: the tabulators would happily process mangled input and render plausible-looking wrong results. The projection re-encodes the semantic content of every ballot, so it must preserve, per voting method:

- score vs. null/skipped distinctions (STAR, Approval)
- ranking order and skipped/duplicate-rank edge cases (IRV, STV, Ranked Robin)
- write-in candidate mapping (approved vs. unapproved, aliases)
- per-race candidate indexing, multi-race ballots, precinct filtering

**Testing bar:** golden-result tests asserting that, for each voting method, tabulation over the compact projection produces byte-identical results to tabulation over the current verbose path (fixtures with write-ins, skipped ranks, nulls, multi-race ballots).

## Related

- #1424 — cursor streaming for `anonymizedBallots` (provides the plumbing this reuses)
- #1423 — response compression (wire size; orthogonal)
- Longer-term: results caching + single-flight recompute (ram-reduction Tier 3) bounds *how often* tabulation runs; this issue bounds *how much memory* each run takes. They compose.

jacksonloper:
Closes #1425.

Tabulating a large election loaded every ballot as a full verbose row before handing it to the tabulators — one of the main drivers of star-server OOMKills. The algorithms are batch, so they genuinely need the whole election at once, but "the whole election" doesn't have to mean 30k verbose JSON ballot objects.

## What changed

`getElectionResults` now streams ballots from a DB cursor (selecting only the `votes` column) and projects each row into a `CompactVoteStore` — flat typed arrays holding one race's marks positionally against its candidate order. The verbose row is dropped immediately, and the tabulator's own input is expanded from the store one race at a time, with the store freeing itself block by block as it's consumed.

Tabulator inputs and outputs are unchanged; this is purely how the input gets built.

## DRY with the bulk-upload format

The positional layout is the same one bulk uploads already use, so it's now defined once in `shared/domain_model/OrderedVoteCodec`:

- the frontend uploader encodes with it (was an inline array literal in `UploadElections.tsx`)
- `castVoteController`'s `mapOrderedNewBallot` decodes with it, including the length validation and its exact error messages
- the projection stores it unrolled into typed arrays

`OrderedVote` widened from `number[]` to `(number | null | undefined)[]`, which is what it already carried at runtime — `JSON.stringify` writes `null` for unmarked candidates and absent ranks.

## Memory

Peak retained (heap + external), synthetic 31k ballots, 10 candidates/race:

| | before | after |
|---|---|---|
| 1 race | 36.5 MB | 5.3 MB |
| 3 races | 70.5 MB | 11.6 MB |

Two things mattered beyond streaming. The store keeps marks in fixed-size blocks and drops each as it's read, so the store and its expansion never both sit at full size. And marks are handed back as int32s rather than raw `Float64Array` doubles — V8 boxes a double stored into an object property as a HeapNumber, which was 310k HeapNumbers per race and more than doubled the expanded cvr.

## Testing

A bug in a projection layer miscounts elections silently rather than failing loudly, so `BallotProjection.test.ts` pins the new path against a verbatim copy of the old inline projection: for every voting method, the tabulator inputs must match and tabulating both must give identical results. Fixtures cover null vs. absent marks, write-ins (approved, unapproved, unrecognized, aliased, shared aliases), duplicate scores, `overvote_rank`, `has_duplicate_rank`, skipped ranks, multi-race ballots, and ballots spanning three storage blocks.

The distinction that most needed pinning: **a candidate the ballot never mentions is not the same as one left explicitly blank.** STAR treats an all-equal ballot as an abstention, so filling an approved write-in's missing slot with 0 would silently reclassify those ballots. That's why marks carry a three-state tag rather than a NaN sentinel.

Each encoding rule was mutation-tested — collapsing absent→null, null→0, last-alias-wins, dropping the unprocessed-write-in counter, and both block-indexing off-by-ones each fail the suite.

## Worth a reviewer's attention

- **`streamVotesByElectionID` deliberately matches `getBallotsByElectionID`'s filter** (head only, *any* status) rather than the submitted-only filter used by `streamSubmittedBallotsByElectionID`. Narrowing it here would silently change election results.
- **The candidate order is snapshotted at construction.** `shuffleCandidatesForRandomTiebreak` sorts `candidates` in place, and an earlier revision read the store's mark→candidate mapping off that array — the two calls were in the only order that works, and swapping them misattributed every ballot's marks with the entire suite still green. `candidateIds` is now immune to that.
- **Connection hold time** is longer than before: the cursor holds a pool connection for query + projection, where the old query released it as soon as it returned. Projection is O(n) and fast, and #1424's `anonymizedBallots` already holds one across a whole HTTP response, but it's a new way for concurrent results requests to contend.
- `getWriteInNamesController` still loads all ballots verbosely. It's admin-only and needs `write_in_name`, which the narrow select drops, so it wants its own query — left out of scope.


packages/backend/src/Tabulators/BallotProjection.test.ts
```diff
@@ -0,0 +1,514 @@
...
+// pin it against `referenceProjection` below — a verbatim copy of the inline
+// projection getElectionResultsController used before the streaming rewrite —
+// and assert both that the tabulator inputs match and that tabulating each
+// voting method over the two produces identical results.
+
+// ---------------------------------------------------------------------------
+// Reference implementation (the pre-streaming code path, unchanged)
+// ---------------------------------------------------------------------------
+
+const referenceProjection = (race: Race, ballots: BallotVotes[]) => {
```
ArendPeter, line 28: At first glance I feel like there's too much logic here. Like the test is confirming the production logic by reimplementing the production logic

packages/backend/src/Tabulators/BallotProjection.test.ts
```diff
@@ -0,0 +1,514 @@
...
+
+// deterministic RNG so a failure is always reproducible
+const mulberry32 = (seed: number) => () => {
+    seed = (seed + 0x6D2B79F5) | 0;
+    let t = Math.imul(seed ^ (seed >>> 15), 1 | seed);
+    t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
+    return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
+};
+
+const shuffled = <T,>(items: T[], rand: () => number): T[] => {
```
ArendPeter: We have a couple of adhoc shuffle functions spread around. I think the TinyRand implementation should be the default. That could also be cleaned up in more places, but that's probably out of scope for the PR

https://github.com/Equal-Vote/bettervoting/blob/15289d30b7cca60c936391eb9532fd40973e78ab/packages/backend/src/Tabulators/tinyrand.ts#L30

packages/backend/src/Tabulators/BallotProjection.test.ts
```diff
@@ -0,0 +1,514 @@
...
+
+    const cvr = projection.takeRawVotes();
+    // toEqual ignores key insertion order, which the tabulators are also
+    // insensitive to (they only ever read marks by candidate id, or reduce over
+    // all of them with order-independent operations)
+    expect(cvr).toEqual(reference.cvr);
+    return {reference, cvr, candidates: projection.candidates};
+};
+
+const compareTabulation = async (race: Race, ballots: BallotVotes[]) => {
```
ArendPeter, line 212: I wonder if we can just remove compareProjection and referenceProjection and just do compare tabulation 👀

packages/backend/src/Tabulators/BallotProjection.test.ts
ArendPeter: NOTE: after reading the production logic, I see how this is much smaller, so I'm fine with this as is

packages/backend/src/Tabulators/BallotProjection.ts
```diff
@@ -0,0 +1,218 @@
...
+        });
+
+        store.commitRow(vote.overvote_rank, vote.has_duplicate_rank);
+    }
+
+    /**
+     * Expand this race's compact store into the tabulator's input and release
+     * the store. Called one race at a time so only a single race's verbose
+     * tabulator input is ever alive; consumeMarks frees the store block by block
+     * as it goes, so the two don't both sit at full size.
```
ArendPeter, line 160: Could you clarify if this optimization would help for single race polls? I'm thinking of the Indivisible CA race with 11k votes

https://bettervoting.com/3vjtj4/results

ArendPeter (approved):
Looks great, thanks for doing this! I was impressed by the optimizations in the BallotProjection & CompactVoteStore pattern.

I left some clarifying questions and optional fixes, but none of them are blockers. Approved! Feel free to merge whenever you're ready

packages/backend/src/Tabulators/BallotProjection.test.ts
jacksonloper: Fixed!

packages/backend/src/Tabulators/BallotProjection.test.ts
jacksonloper: I don't think so.  Like consider the case that we screwed up copying numUnprocessedWriteIns.  The tabulation results would be the same, but the intermediate representation wouldn't be the same.

packages/backend/src/Tabulators/BallotProjection.ts
jacksonloper: Yes it helps single race for sure.  We no longer hold all rows in memory for any race.  Thats the streaming part.

Issue #1350, opened 2026-05-13
Add a disclaimer related to preliminary results


ArendPeter:
On the ballot, there should be a link to a separate article whenever there are preliminary results. This article would clarify all of the anonymity and security concerns that arise when there are preliminary results. We want to make sure this is transparent to the voter before they've cast their ballot. 

There should be an additional layer of warning when it's a closed list elections, because election admins are able to see who has voted and it's extra trivial for them to reveal what those votes are.

masiarek:
## Progress

Spent time mapping this against the codebase before writing any code. Three findings that change the shape of the work, then proposed wording and some open questions.

**1. The article already exists.** 
https://docs.bettervoting.com/help/preliminary_results.html 
`docs/help/preliminary_results.md` is on `main` (`nav_order: 8`) and already covers everything this issue asks the article to cover — what a live tally exposes, that admins can always see who voted, delta-analysis de-anonymization (L26), the editable-ballot interaction, and that turning the setting back off doesn't un-reveal (L44-50). It is referenced from **nowhere** in `packages/` — the only `preliminary_results` hits are the i18n label key and its one consumer in `ElectionSettings.tsx`. So this issue is a **linking + copy** task, not a writing task. Nice — that's most of the work already done.

**2. Where the notice has to go, and one trap.** The right insertion point is `VotePage.tsx` (~L259-265), where `<DraftWarning/>` and `<ElectionStateWarning state="archived">` already stack above the ballot. That spot is the only one that renders once per ballot rather than once per race **and** covers `DraggableIRVBallotView`, which bypasses `GenericBallotView` entirely and re-implements its own instructions block with no footer. If the notice goes in the ballot footer instead, ranked elections silently don't get it.

**3. A second trap worth flagging before anyone writes the link.** `util.tsx` L235 renders markdown links as `target={v['newWindow'] ? '_blank' : '_self'}` — same-tab by default — and `ElectionStateWarning` calls `t(title)`/`t(description)` with no values object, so it can't pass `newWindow`. On the ballot a same-tab navigation destroys the in-progress ballot (`VotePage`'s races are React state with no draft persistence). So the ballot-side link should be an explicit `<Link target='_blank' rel='noreferrer'>` in `ElectionStateWarning`'s `children` slot, not a markdown link in an i18n value. The admin-side link can just use `learn_link:`, which already renders a `_blank` "Learn More" anchor via `styles.tsx` L33.

## Proposed wording

[...]

PR #1465, opened 2026-07-30
fix(i18n): rewrite the Public Results tip and link the help article

masiarek:
Part of #1350.

Three changes to `tips.public_results`, all in `packages/frontend/src/i18n/en.yaml`:

**1. Reworded**, per the issue's Should-be text — with one deliberate deviation. The issue quotes the *rendered* string, so pasting it verbatim would hard-code "election" and regress poll terminology. This keeps the `{{election}}` / `{{elections}}` interpolation and adds `{{ballot}}`, so as a poll it now reads "…after submitting their **response**" (`keyword.poll.ballot` is `response`).

**2. Block style `|` → `>`.** `applyLineBreaks` turns a literal block's newlines into `<br/>`, so the old tip rendered as three separate lines in the tooltip. The replacement is a single paragraph and now renders as one. (Screenshot of the current three-line rendering to follow.)

**3. Added `learn_link`.** `docs/help/preliminary_results.md` has been in the repo for a while but is linked from nowhere in `packages/` — a grep for `preliminary_results` returns only the label key and its one consumer. It already covers what #1350 asks an article to cover: what a live tally exposes, that admins can always see who voted, delta-analysis de-anonymization, the editable-ballots interaction, and that turning the setting off doesn't un-reveal. `styles.tsx` renders `learn_link` as a "Learn More" anchor with `target='_blank'`, so this needs no component change.

Verified `https://docs.bettervoting.com/help/preliminary_results.html` returns 200.

---

**Scope:** this is the admin-side copy half of #1350 only. The on-ballot notice for voters, the closed-list warning layer, and the qualification to the article's own L26 claim are separate and still open — the closed-list wording in particular needs someone to sign off on what it may claim (see my comment on the issue).

**Translations:** en-only. `es` / `pl` / `pt-BR` carry no `tips:` block at all, and `tips:` sits in PRIORITY 4, so this creates no translator obligation.

**Noticed while in this region, not touched here:** `results.admin_results_toggle` and `disabled_msgs.ballot_updates_when_open` both appear to have no consumer outside `en.yaml`. Happy to remove them in a separate cleanup PR rather than widen this diff.

### DIFF
--- a/packages/frontend/src/i18n/en.yaml
+++ b/packages/frontend/src/i18n/en.yaml
@@ -736,10 +736,11 @@ tips:
 
   public_results:
     title: Public Results
-    description: |
-      Allows voters to see the results of the {{election}}.
-      If enabled while voting is open then voters will be shown to the preliminary results after completing their ballot.
-      High profile {{elections}} will usually keep the results hidden, and then reveal them after the {{election}} is closed.
+    description: >
+      Controls whether voters can see {{election}} results.
+      When enabled during an open {{election}}, voters will see preliminary results after submitting their {{ballot}}.
+      High-profile {{elections}} typically keep results hidden until the {{election}} closes.
+    learn_link: https://docs.bettervoting.com/help/preliminary_results.html
 
   voter_groups:
     title: Enable Voter Groups


ArendPeter (changes requested):
I think the PR is missing the preliminary_results.md file

ArendPeter (approved):
Actually, I see that the preliminary results article is already published https://docs.bettervoting.com/help/preliminary_results.html

Great work!

Issue #1086, opened 2025-11-12
UI displays 'STAR Voting' instead of 'Bloc STAR' and has incorrect help link


masiarek:
- Test case: **BV129 - 3 cand - 2 winners (Bloc STAR) - score tiebreaker** - https://docs.google.com/document/d/1I_8NZlcAPwJcb_n5TqTN5V9queHN5CR0SSaZkVOJmpI/edit?tab=t.0
- video: https://youtu.be/WL6RIVaTBn4

There are several UI inconsistencies when running a Bloc STAR election (e.g., a multi-winner race using "Basic Multi-Winner"). The UI incorrectly displays the voting method as "STAR Voting" instead of the more specific "Bloc STAR Voting" and provides an incorrect explanatory link.

This issue was observed in test election BV129 - 3 cand - 2 winners (Bloc STAR).

Problem Areas
### 1. "Edit Race" Modal
When editing a Bloc STAR race, the UI is incorrect:

<img width="593" height="756" alt="Image" src="https://github.com/user-attachments/assets/2a995764-89c3-46ec-946b-6e08fef4a7bd" />

Location: "Edit Race" modal > "Which Voting Method?" field.

Actual Behavior: The field displays "STAR Voting".

Expected Behavior: The field should display "Bloc STAR Voting" to accurately reflect the multi-winner method being used.

### 2. Election Results Page
The results page shows two separate issues:

<img width="1002" height="379" alt="Image" src="https://github.com/user-attachments/assets/471d63d8-c86e-4bb5-a9e9-12bb377c400d" />

**Location 1**: Voting method label (below the winner announcement).

Actual Behavior: The label reads Voting Method: STAR Voting.

Expected Behavior: This label should read Voting Method: Bloc STAR Voting.

**Location 2:** Help link.

Actual Behavior: An incorrect link is displayed: How STAR voting works.

Expected Behavior: This link should be removed or, preferably, replaced.
Link URL: Point to the correct external resource: https://electowiki.org/wiki/Bloc_STAR_Voting

masiarek:
Test case "BV1820 - method names - STAR - STAR Bloc - STAR PR" - https://github.com/Equal-Vote/bettervoting/issues/904

masiarek:
The result is correct — this is only the exported method **name**, but it's a slippery one. 

The method is stored as plain `"STAR"` in two places, with the seat count in a separate field:

```json
Election.races[0].voting_method : "STAR"     num_winners : 2
[...]

PR #1474, opened 2026-08-04
docs: add a Bloc STAR Voting help page, and link it from the Basic Multi-Winner tip

masiarek:
## Description

BetterVoting can already run Bloc STAR — **STAR** + **Basic Multi-Winner** — but there's nothing in the docs that explains it. `docs/help/` has no page for any voting method's multi-winner behaviour, the ballot text exists only inside `paper_ballots.md`, `faq.md` never mentions multi-winner, and `tips.bloc_multi_winner` is the only method tip in `en.yaml` with no `learn_link`, so its tooltip renders without a **Learn More** at all (`tips.proportional_multi_winner` has one, to equal.vote/pr).

Two commits:

**1. `docs/help/bloc_star.md`** (new, nav_order 9). Covers:

- what Bloc STAR is, and that "Multi-Winner STAR Voting" unqualified means Bloc STAR (technical specifications 1.c / 1.e)
- that the 0–5 ballot doesn't change with the seat count, with the same ballot text already used in [Paper Ballots](https://docs.bettervoting.com/help/paper_ballots.html)
- the count: scoring round → automatic runoff → **remove the winner and repeat**, and that removal doesn't reweight or spend anything (the contrast with Proportional Multi-Winner)
- a fully worked 3-ballot / 3-candidate / 2-seat example — including a **1–1 runoff tie at seat 2 broken by score**, which walks a reader straight into [Ties](https://docs.bettervoting.com/help/ties.html)
- how to set the race up in the admin UI
- **when not to use it**: not proportional, a cohesive majority sweeps every seat, and the at-large / Section 2 Voting Rights Act warning from [starvoting.org/multi_winner](https://www.starvoting.org/multi_winner) — that one is currently written down nowhere on the site, and it's the guidance an admin most needs before choosing Basic Multi-Winner for a representative body
- voter-facing strategy: score honestly, and specifically *don't* pad five stars out to the number of seats

**2. `en.yaml`** — one line, `learn_link` on `tips.bloc_multi_winner` pointing at the new page. `styles.tsx` renders it as the standard "Learn More" anchor, so no component change is needed; this is the same move as #1465. Drop this commit if you'd rather land the page on its own first.

Notes:

- The worked example is engine-checked, not hand-waved: tabulated with Larry Hastings' `starvote` (`method=bloc, seats=2`), winners **A then C**, seat 2 decided by the higher-score tiebreaker exactly as `docs/help/ties.md` specifies. BetterVoting's ladder gives the same answer.
- I phrased the VRA sentence more carefully than starvoting.org does. The site says at-large bloc voting "was banned in the Voting Rights Act"; what's accurate is that at-large bloc systems have repeatedly been struck down **under Section 2** where they diluted minority voting strength. The recommendation is unchanged.
- **Translations:** en-only. `es` / `pl` / `pt-BR` carry no `tips:` block, so this creates no translator obligation.
- New page will publish to `https://docs.bettervoting.com/help/bloc_star.html`.

No visual change beyond the tooltip gaining a "Learn More" anchor, in the same position and style as the Proportional Multi-Winner tip already has.

## Related Issues

Related to #1086 — it asks for the results-page help link to stop pointing at "How STAR voting works" for a Bloc race, and suggests electowiki as the target. This gives BetterVoting a first-party page to point at instead, but doesn't touch the results page, so I'm leaving #1086 open.


masiarek:
Pushed a third commit after actually reading the specification rather than the campaign pages.

`starvoting.org/technical_specifications` 301s to a scanned PDF (v1.3, published 2024-12-20) with no text layer, which is presumably why the two wordings have been allowed to drift. Extracted from the file, **§3.d prescribes a different multi-winner explanation than the paper-ballot text this repo carries:**

| | technical specifications §3.d | `docs/help/paper_ballots.md` |
|---|---|---|
| who wins a seat | "**the finalist** preferred by the most voters wins" | "**the candidate** with the most votes is elected" |
| the ballot goes to | "your vote" | "your **full** vote" |
| the winner count | above the race (§3.c) | folded into the first instruction |

The first row is the one worth caring about. *"The candidate with the most votes"* invites a voter to read stars as votes and conclude the top scorer is elected — exactly what Bloc STAR doesn't guarantee, and the thing this page has a whole section warning about. *"The finalist preferred by the most voters"* names the field and the currency and closes it.

§3.e permits paraphrase "as long as they are presented with the meaning unchanged", so the circulated version isn't off-spec — but that clause is where the meaning is arguably not unchanged. The page now prints both and says to use the specification's on a real ballot.

Two other things from the same read, for whoever maintains the paper-ballot doc: §3.a makes printing the instructions and the method explanation **mandatory**, not advisory; and §2.c defines Bloc as single-winner STAR "with an additional Automatic Runoff round conducted for each seat" — terse, but correct, because the removal step leaves every remaining candidate's score untouched, so only the runoff genuinely repeats.

Happy to make the same correction in `docs/help/paper_ballots.md` in a separate PR if you want the two docs consistent — I left it out of this one to keep the diff to a new file plus one line.

masiarek:
Correction on my side: my previous commit wrote the whole file from a copy taken before fe6c1773, so it reverted that denominator fix and put the wrong claim back on the page for about ten minutes. Restored verbatim in 745a4a3 — the bar labels divide by every ballot cast, only the majority-threshold marker drops the Equal Support row. The specification-wording section is unaffected.

PR #1477, opened 2026-08-04
Don't store voter emails/IPs on rolls for elections that don't authenticate

jacksonloper:
## The bug

`getOrCreateElectionRoll` derives two locals from the election's `voter_authentication` settings, correctly leaving them `null` when the election doesn't authenticate on them:

```ts
const require_ip_hash = (election.settings.voter_authentication.ip_address ? ip_hash : null) ?? null;
const email = election.settings.voter_authentication.email ? req.user?.email : null
```

The roll it then builds ignored both and copied the raw request values in:

```ts
email: req.user?.email ? req.user.email : undefined,
ip_hash: ip_hash,
```

The guard that decides whether to persist the roll *does* consult the settings, which is why this looked safe. It isn't. The cast-vote path reaches the "not persisted" `else` branch and returns the roll anyway, and `CastVoteStore.submitBallotEvent` inserts whatever roll it is handed:

```ts
const { address: _address, registration: _registration, ...rollToInsert } = event.roll;
await trx.insertInto('electionRollDB').values(rollToInsert).execute();
```

So on the voting path that guard is effectively dead code — it only protects a branch nothing reaches.

Net effect: an **`open_open`** election — `voter_access: 'open'` + `voter_authentication: {}`, the mode the UI presents as requiring no authentication at all — stores the voter's email address and an IP hash next to their `ballot_id`. The email lands whenever the voter happens to have a session cookie, which they often do; the IP hash lands always, session or not. Neither is anything the election asked for.

`ElectionRoll.ip_hash` is already documented in the domain model as *"sha256(req.ip); set when `voter_authentication.ip_address` is enabled."* After this change that's actually true.

## Scope on prod

Querying `electionRollDB` joined against head elections with `voter_authentication = '{}'` and `voter_access = 'open'`:

| | rows |
|---|---|
| ballots with an `ip_hash` stored | **937** |
| …of which also have an email | **456** (70 elections) |

Every one of those rows has both `ballot_id` set and `submitted = true` — i.e. all of them came through `castVote`, none from an admin adding voters. Still happening; most recent was yesterday.

Not proposing a data migration here — happy to do that separately.

## How exposed is it, really?

Lower than it first looks, and worth stating plainly so this gets prioritised correctly:

- `GET /Election/:id/rolls` **refuses outright** for `voter_access: 'open'` elections, owner included — so admins can't list this today.
- `GET /Election/:id/rolls/:voter_id` scrubs `ballot_id` and `ip_hash`, but **not `email`**, and has no `expectPermission` call. Voter IDs aren't enumerable, and the party most likely to hold one is the voter themselves — but a forwarded ballot-update link (`/:election_id/id/:voter_id`, which receipt emails contain when `ballot_updates` is on) would disclose that voter's email. I've left that alone here; it's a separate fix.

So the practical case for this PR is data minimisation rather than an open door: the rows shouldn't exist, they're in every backup, and the only thing standing between them and disclosure is one `if` in one controller that any future endpoint or CSV export has to remember to reproduce.

## What's in here

- `voterRollUtils.ts` — use the already-computed `email` / `require_ip_hash` instead of the raw request values. Also corrected the comment on the `else` branch, which claimed nothing reaches it.
- `__mocks__/CastVoteStore.ts` — the mock only called `rollStore.update()`, which no-ops when there's no head row to update. That's exactly the `open_open` case, so the mock silently dropped every roll the cast-vote path writes and this entire class of bug was invisible to the suite. It now inserts when there's no head row, matching the real store.
- `openElectionRollPrivacy.test.ts` — `open_open` stores neither field; `{email: true}` still stores the email; `{ip_address: true}` still stores the hash and nothing else. Verified these fail without the `voterRollUtils` change.

The test asserts against the roll store rather than `GET /rolls`, since that endpoint won't list open elections — which is a decent part of why nobody noticed. The row was being written; nothing was reading it back.

## Not affected

Receipt emails. `castVoteController` already falls back through `event.roll?.email ?? extractUserFromRequest(req)?.email ?? req.body.receiptEmail`, so a signed-in voter on an `open_open` election still gets their receipt from the second link in that chain.

Full backend suite: 158/158.


ArendPeter:
We chatted about this one in stand up, in the future we may have the ability for admins to live switch between admin methods (probably only for casual elections, that would be weird on official ones). Logging the ip_hash and email lets us future proof, so I'll close this PR

PR #1490, opened 2026-08-07
Restore the publicly-searchable toggle on the election settings page

jacksonloper:
## What happened

The **"Make {{election}} Publicly Searchable"** (`is_public`) setting silently disappeared from the admin settings page. Two branches collided:

- **`4b4aec34`** (Apr 28) — *"fix make election publically searchable toggle"*. The checkbox had been writing to `election.settings.is_public`, but the field actually lives at the **top level** of `Election`. That commit added separate `editedIsPublic` state wired to `election.is_public`.
- **`5808a623`** (authored Mar 31, merged later) — the checkbox→`SwitchSetting` refactor. It converted `random_candidate_order`, `ballot_updates`, `require_instruction_confirmation`, `draggable_ballot`, and `max_rankings` into switches, but `<CheckboxSetting setting='is_public'/>` was deleted and never carried over.

Because the refactor branch was authored *before* the fix but landed *on top of* it, both the fix and the setting went away in one merge. Nothing in that diff suggests it was intentional — the other four toggles all survived.

## Impact

Everything below the UI still worked:

- `packages/shared/src/domain_model/Election.ts:27` — `is_public?: boolean`, still validated at `:148`
- `packages/backend/src/Models/Elections.ts:103` — `getOpenElections` still filters `.where("is_public", "=", true)`
- `packages/frontend/src/i18n/en.yaml` — the label and tip text were still there, orphaned

…but the only remaining writer in the frontend was `Wizard.tsx:33` hardcoding `is_public: false`. So **no election created since that merge could ever appear on `/browse`** — the browse tab has only been showing elections whose flag was set before the regression.

## The change

Restored the setting on `/admin/settings`, next to the other toggles.

It uses a bare `SwitchSetting` + `useOptimisticToggle` rather than the `ElectionSwitchSetting` helper, because that helper writes into `election.settings[key]` and `is_public` is a top-level column — same reason the `public_results` switch is hand-rolled.

One addition beyond a straight revert: the switch is **disabled with an explanation on voter-list elections**. `getOpenElections` also filters on `settings.voter_access == 'open'`, so on a closed election turning `is_public` on would have no visible effect. New string `disabled_msgs.is_public_when_voter_list`.

## Verification

- `tsc --noEmit` on the frontend: clean
- `eslint` on the changed component: clean
- `en.yaml` parses

Not covered: there's no existing E2E coverage of the settings page (`testing/tests/` has no `/admin/settings` spec), and adding one needs the full docker stack to run against, so I didn't add a regression test here. Worth a follow-up — a switch-level assertion on that page would have caught this.

## Related loose end (not fixed here)

`en.yaml:923` `public_description` — *"{{capital_election}} will be open to anyone via the 'Browse Polls' page"* — is now unreferenced anywhere in the frontend. It was attached to the wizard's old "public" voter-access preset, which the auth-modes refactor (`7d43150b`) removed. Worth deciding whether the creation wizard should offer public listing up front, rather than only via settings after the fact.


ArendPeter (approved):
Thank you!!

PR #1491, opened 2026-08-07
Make MockEventQueue run jobs inline; drop test-suite leak workarounds

### LINKED ISSUE #1481: Backend test suite flakes ~1 run in 12: shared process + leaked MockEventQueue timers
## Summary

The backend suite fails roughly **1 run in 12** (measured: **4 failures in 50 consecutive full-suite runs**). The failing test is different every time, and every affected file passes cleanly in isolation. The failures are transport/routing-level — not assertion bugs and mostly not timeouts — and they trace back to the whole suite sharing a single Node process with leaked async work.

This is long-standing and independent of the Jest version. It was found while investigating #1467 (the `sanitize-html` bump), but it is not caused by that PR.

## Evidence

Measured on `main` + the `sanitize-html@2.17.6` bump, Node 24.15.0, macOS:

| runner | failing runs | rate |
| --- | --- | --- |
| jest 29.7.0 (current) | 4 / 50 | 8% |
| jest 30.4.1 | 2 / ~65 | 3% |

Failure signatures across the 4 jest 29 failures (de-duplicated; Jest prints each block twice):

- **7 × HTTP 404** — election not found, for elections the test had just created
- **2 × HTTP 400** — from `createElection`, with **no `"Invalid Election"` line in the server log**, i.e. the request never reached `electionValidation`
- **1 × `read ECONNRESET`**
- **1 ×** genuine `Exceeded timeout of 5000 ms`

Tests observed failing, one per run: `editElection › Election data provided › responds with 200 status`, `customAuthKey › User with custom key CAN request ballot`, `idRoll › (whole file)`, `finalizeElection › Finalize Election`.

Ruled out by experiment: CPU contention (5/5 clean under 16 synthetic load hogs), cold caches (3/3 clean after `--clearCache`), and any single slow test (at `--testTimeout=4100` all 149 tests still pass). Files run in isolation do not reproduce it — `editElection` alone is 10 tests in 2.7 s; `customAuthKey` alone was 15/15 clean.

## Root cause

Three things combine:

1. **The whole suite runs in one process.** `npm test` passes `--detectOpenHandles`, and `@jest/core` forces `runInBand` whenever that flag is set:
   ```js
   // detectOpenHandles makes no sense without runInBand, because it cannot detect leaks in workers
   if (runInBand || detectOpenHandles) {
   ```
   So all 22 files and 149 tests share one process, with no worker isolation.

2. **Every file stands up its own Express app in that shared process.** The `TestHelper` constructor calls `makeApp()`, and supertest spins an ephemeral server per request. Multiple live apps coexist. This is directly visible: `ERROR: already have handler for queue sendEmailEvent` appears **6 times in every run** (passing and failing alike) — multiple apps registering handlers on the same shared queue.

3. **`MockEventQueue` leaks live timer chains across files.** It drains jobs on a real 1-second timer that re-arms after each job:
   ```ts
   private _delay = 1000;
   ...
   await new Promise(r => setTimeout(r, this._delay));
   await this.processNextJob();   // → triggerJobs() again → another 1000 ms
   ```
   Jest resets the module registry per file, so each file gets a *fresh* queue while the previous file's chain keeps running — still executing handlers, still logging. The logs show `MEQ: Processing job` for one file's election while a later file is already starting. This is exactly the `Cannot log after tests are done → "Event queue empty"` spam, and why `--forceExit` is needed for `npm test` to terminate at all.

Net effect: requests occasionally die on the socket (`ECONNRESET`) or land somewhere that doesn't have the test's data (404, or a 400 that leaves no server-side trace). The precise mechanism that selects 404 vs 400 vs `ECONNRESET` was not pinned down — that needs instrumented reproduction at a ~1-in-12 rate.

## Related: `finalizeElection` has under 1 second of timeout margin

`finalizeElection.test.ts:45` sleeps a hardcoded 4 seconds:

```ts
//Wait a few seconds for job queue to process emails, I imagine there's a better way to do this.
await new Promise(resolve => setTimeout(resolve, 4000));
```

Measured across 10 runs, that test takes a rock-steady **4022–4028 ms against the default 5000 ms timeout**. It has already failed once on a timeout (jest 29, run 46). It is the only test anywhere near the limit.

## Suggested fix

1. **Make `MockEventQueue._delay` ~0 under test.** This removes the leaked timer chains, allows the 4-second sleep in `finalizeElection.test.ts` to be deleted, and cuts roughly 11 s off the suite (`finalizeElection` ~4.0 s, `anonymizedBallots` ~3.0 s, `writeIns` ~3.0 s, `emailRoll`/`idRoll` ~1.0 s each are all queue-drain time).
2. **Then drop `--detectOpenHandles`** from the `test` script so files get worker isolation again — which removes the shared-process collisions entirely. `--forceExit` currently masks the leak and can go too once (1) lands.

Doing (1) first is what makes (2) safe: the open handles that `--forceExit` is papering over are the queue's timer chains.

## Repro

```bash
cd packages/backend
for i in $(seq 1 50); do npm test > /tmp/run$i.log 2>&1; grep -E '^Tests: ' /tmp/run$i.log; done
grep -l "failed" /tmp/run*.log
```

Expect roughly 4 failures in 50, each naming a different test.

jacksonloper:
Fixes #1481.

## What was wrong

`MockEventQueue` drained its jobs on a self-rearming 1-second timer that nothing awaited:

```ts
private async triggerJobs(){
    ...
    await new Promise(r => setTimeout(r, this._delay));  // 1000ms
    await this.processNextJob();   // → triggerJobs() again
}
```

Jest isolates module registries but **not the event loop**. Each file got a fresh queue, fresh mock DBs and a fresh app, but the previous file's chain kept running — against the previous file's objects — for roughly one second per queued job. That's the leak `--forceExit` was papering over, and the reason the whole suite ran in one process: `--detectOpenHandles` forces `runInBand` (`@jest/core/build/testSchedulerHelper.js:29`).

**The leaked jobs also never did their work.** Once a file has torn down, `console.info` throws, and the `"MEQ: Processing job"` log sits inside `processNextJob`'s `try`. So roughly 4 jobs per run aborted before their handler ran and were silently discarded, and the `catch`'s own log threw again. The timer wasn't modelling async delivery — it was randomly dropping about a third of the jobs.

## What changed

**Jobs run inline.** `publish()` resolves once its handlers have finished. Tests that need to observe the gap between "enqueued" and "handled" use `pause()` / `resume()`, which is deterministic and needs no timer. This is no less faithful to production than before: pg-boss polls at `newJobCheckInterval: 500` with `teamSize: 5` (`PGBossEventQueue.ts:39`), so the mock's strictly-serial 1000 ms modelled neither its latency nor its concurrency.

**Removed the waits this made unnecessary** — the six `waitUntilJobsFinished()` call sites and `finalizeElection`'s 4-second sleep.

Every one of them was already a no-op, which I verified rather than assumed: with the old 1 s delay left intact and `waitUntilJobsFinished` neutered to a no-op, all 47 tests in the four affected files still passed. The reason is structural — ballots are persisted synchronously at `castVoteController.ts:256`, before the response; the queue only mails receipts.

`finalizeElection`'s sleep was worse than decorative. Its comment said "Wait a few seconds for job queue to process emails", but finalize publishes nothing at all — invites go out through `POST /Election/:id/sendInvites`. It burned 4 s of a 5 s timeout waiting for jobs that were never created, and it's the one test the issue records failing on a genuine timeout.

**Added `ballotReceiptEmail.test.ts`.** Nothing in the suite had ever asserted on the queue's output: the mock `EmailService` records into `sentEmails` and `TestHelper:47` wires it up, but no test read it, and both `emailService.clear()` calls were commented out. So `handleCastVoteEvent` — the only consumer the suite exercises at all — had zero coverage. The receipt is worth pinning down because `castVoteController` deliberately scrubs `ballot_id` from the HTTP response, making the email the only channel that carries it back to the voter.

Both tests are mutation-checked, so they aren't vacuous:

| mutant | test 1 | test 2 |
| --- | --- | --- |
| drop the `sendEmails` call in `handleCastVoteEvent` | ✗ fails | ✗ fails |
| send the receipt inline instead of publishing | ✓ passes | ✗ fails |

**Dropped both jest flags.** `--detectOpenHandles` was reporting nothing while serialising 22 files into one process; `--forceExit` is no longer needed because nothing is left scheduled.

**Switched `emailRoll` to `th.expressApp`** instead of a second `makeApp()`, which was re-registering all three handlers and printing three `already have handler` errors every run. (This was cited in the issue as evidence of cross-file collision — it isn't; it's one file calling `makeApp()` twice.)

## Results

Measured locally, 50 consecutive runs:

| | before | after |
| --- | --- | --- |
| wall time | 15.3 s | **2.8 s** |
| failures | 4 / 50 (per issue) | **0 / 50** |
| "Cannot log after tests are done" | 9 per run | 0 |
| `already have handler` errors | 3 per run | 0 |
| exits without `--forceExit` | no | **yes** |

151 tests, 23 suites, `tsc` clean.

One honest caveat on the flake number: 0/50 is consistent with the fix working, but if the true rate were still 8% you'd see a clean 50 about 1.5% of the time. The mechanism that selected 404 vs 400 vs ECONNRESET was never pinned down in #1481 and isn't pinned down here either — what this PR proves is that the leak is gone and the suite exits on its own. If a flake resurfaces, the remaining suspect is supertest's one-ephemeral-listener-per-request pattern (`request(app)` calls `http.createServer` on every call), not the queue.

## Follow-ups not in this PR

- `handleSendInviteEvent` and `handleSendEmailEvent` are subscribed and **never invoked by any test** — no test hits `/sendInvites`, `/sendEmails`, or `/sendInvite/:voter_id`. Worth covering with the same drain pattern.
- When you do, note that `sendInvitesController.ts:136` reads `emailResponse.length` and `emailResponse[0][0].statusCode` from `EmailService.sendEmails`, but the mock returns `undefined`. That path will throw on the mock the moment a test drives it — the fake's return contract never matched what the production code expects, because nothing ever exercised it.
- `MockEventQueue.ts` still lives beside the real implementation in `src/Services/EventQueue/` rather than a `__mocks__/` directory, so it compiles into `build/`. Left alone here to keep the diff focused.

https://claude.ai/code/session_01ECGsUoD9JCefu95Aw7HieW

### DIFF
--- a/packages/backend/src/Services/EventQueue/MockEventQueue.ts
+++ b/packages/backend/src/Services/EventQueue/MockEventQueue.ts
@@ -1,6 +1,5 @@
 import { randomUUID } from "crypto";
-import { ILoggingContext } from "../Logging/ILogger";
-import { EventHandler, IEventQueue, JobInsert } from "./IEventQueue";
+import { EventHandler, IEventQueue } from "./IEventQueue";
 import { QueueName } from "./QueueName";
 
 type Job = {
@@ -9,18 +8,23 @@ type Job = {
     data: object
 }
 
+/**
+ * In-memory stand-in for PGBossEventQueue. pg-boss keeps its jobs in Postgres,
+ * so the backend unit tests — which run entirely against in-memory fakes — need
+ * a substitute rather than the real thing.
+ *
+ * Jobs run inline: publish() does not resolve until its handlers have finished.
+ * That keeps tests deterministic and leaves nothing scheduled once a test file
+ * ends. When a test needs to observe the state *between* "job enqueued" and
+ * "job handled" — e.g. to check that a response is sent without waiting on the
+ * queue — use pause() / resume() rather than a sleep.
+ */
 export class MockEventQueue implements IEventQueue {
 
-    private _delay = 1000;
-    
     private _handlers:Map<QueueName,EventHandler> = new Map();
     private _pendingJobs:Array<Job> = [];
-    private _working:boolean = false;
+    private _draining:boolean = false;
     private _paused:boolean = false;
-
-
-    constructor(){
-    }
 
     public subscribe(queue:QueueName, handler:EventHandler):void {
         if (this._handlers.has(queue)){
@@ -30,54 +34,46 @@ export class MockEventQueue implements IEventQueue {
     }
 
     public async publish(queue:QueueName, data:object):Promise<string> {
-        var j = {
+        const job = {
             queue: queue,
             data: data,
             id: randomUUID()
         }
-        this._pendingJobs.push(j);
-        this.triggerJobs();
-        return j.id;
+        this._pendingJobs.push(job);
+        await this.drain();
+        return job.id;
     }
 
     public async publishBatch(queue:QueueName, data:object[]):Promise<object> {
-        var j = data.map(d => ({
+        const jobs = data.map(d => ({
             queue: queue,
             data: d,
             id: randomUUID()
         }))
-        this._pendingJobs.push(...j);
-        this.triggerJobs();
-        return j;
+        this._pendingJobs.push(...jobs);
+        await this.drain();
+        return jobs;
     }
 
-    private async triggerJobs(){
-        if (this._working){
+    /**
+     * Runs every pending job to completion. Does nothing while paused, and is
+     * re-entrant safe: a handler that publishes another job leaves it for the
+     * loop already running rather than starting a nested one.
+     */
+    public async drain():Promise<void> {
+        if (this._paused || this._draining){
             return;
         }
-        if (this._paused){
-            return
-        }
-        this._working = true;
-        await new Promise(r => setTimeout(r, this._delay));
-        await this.processNextJob();
-    }
-
-    private async processNextJob(){
-        var j = this._pendingJobs.shift();
-        if (!j){
-            console.info("Event queue empty");
-            this._working = false;
-            return;
-        }
+        this._draining = true;
         try {
-            console.info("MEQ: Processing job: " + JSON.stringify(j));
-            await this.doJob(j);
-        } catch (e:any) {
-            console.info("MEQ: Exception handling job: " + JSON.stringify(j));
+            var job = this._pendingJobs.shift();
+            while (job){
+                await this.doJob(job);
+                job = this._pendingJobs.shift();
+            }
+        } finally {
+            this._draining = false;
         }
-        this._working = false;
-        this.triggerJobs();
     }
 
     private async doJob(job:Job){
@@ -86,16 +82,26 @@ export class MockEventQueue implements IEventQueue {
             console.info("ERROR: no handler for queue "+job.queue);
             return;
         }
-        await h(job);
+        try {
+            await h(job);
+        } catch (e:any) {
+            // pg-boss retries a failed job; the mock just reports it, so that a
+            // broken handler shows up in the test output instead of vanishing.
+            console.info(`MEQ: Exception handling job ${job.id} on ${job.queue}: ${e?.stack ?? e}`);
+        }
     }
 
     public pause(){
         this._paused = true;
     }
 
-    public resume(){
+    public async resume(){
         this._paused = false;
-        this.triggerJobs();
+        await this.drain();
+    }
+
+    public pendingJobCount():number {
+        return this._pendingJobs.length;
     }
 
     public async clearStorage():Promise<void> {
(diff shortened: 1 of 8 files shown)


packages/backend/src/Services/EventQueue/MockEventQueue.ts
```diff
@@ -30,54 +34,46 @@ export class MockEventQueue implements IEventQueue {
...
-            console.info("Event queue empty");
-            this._working = false;
+    /**
+     * Runs every pending job to completion. Does nothing while paused, and is
+     * re-entrant safe: a handler that publishes another job leaves it for the
+     * loop already running rather than starting a nested one.
+     */
+    public async drain():Promise<void> {
+        if (this._paused || this._draining){
             return;
```
ArendPeter, line 65: I wonder if it's better to throw an error here instead of returning?

After drain is awaited the caller expects the pending jobs to be empty, however if drain() is called while draining then the early return would violate that assumption.

I'm not sure if we even want to support having drain() be called while it's draining so I think throwing an error makes sense.

ArendPeter (changes requested):
Very exciting, I'm looking forward to the faster, less flaky tests! I left a comment for clarification before approving

PR #1493, opened 2026-08-08
docs: fix spelling and grammar errors on paper ballots page

masiarek:
## Description

Copy-editing pass over the [Paper Ballots](https://docs.bettervoting.com/help/paper_ballots.html) help page (`docs/help/paper_ballots.md`). Text-only, no code touched. Nine fixes on eight lines, in three commits.

### Spelling (2)

1. **"Propositional STAR" → "Proportional STAR"** — the heading directly above this sentence and the bullet directly below it both already say "Proportional", so this line was the odd one out.
2. **"pupulation" → "population"** — Eugene, OR, in the Gravic/Remark section.

### Grammar, agreement, and word choice (7)

3. **"each score read is correctly" → "each score is read correctly"** — transposed words in the hand-count observer roles.
4. **"your voters personal identifying information" → "your voters' personal…"** — missing possessive apostrophe.
5. **"every voter receive a hard copy receipt" → "receives"** — subject–verb agreement.
6. **"a voter's name and email addresses confirmed" → "email address"** — number agreement with the singular "a voter's".
7. **"verify a voter's addresses and then use their address" → "a voter's address"** — same, and the rest of the sentence already treats it as singular.
8. **"a post marked piece of mail" → "a postmarked piece of mail"** — closes the compound, matching the correctly-spelled "postmarked" two paragraphs later.
9. **"(which runs elections for the City of Eugene,)"** — dropped the stray comma before the closing parenthesis.

Fix 1 seemed worth more than ordinary typo hygiene: this is the page people copy ballot instructions off of when printing real paper ballots, and "propositional" is a plausible-looking word a reader could carry onto a printed ballot.

Deliberately left alone: the page's conversational register ("regardless if you are using STAR Voting or not", "All you'll need is a paper, pencil, and a calculator", "you'll have your winner!") and its hyphenation style ("full scale pilot", "time consuming", "3 person teams"). Those read as the author's voice and house style rather than errors, so they seemed yours to decide rather than mine.

No reflowing or whitespace changes — every hunk is a word-level edit on an otherwise untouched line.

### One thing I noticed but did not change

The page ends with a dangling cross-reference:

> More info on STAR Voting implementation costs and options for governmental elections can be found here:
> STAR Voting Cost of Implementation

"STAR Voting Cost of Implementation" looks like it was meant to be a link but has no URL, so it renders as plain text. I did not know the intended target — happy to add it in this PR if you point me at it.

N/A — documentation text only.

## Related Issues

None.

---

**Update (review):** the "Advanced registration" → "Advance registration" change has been reverted in 694de5ad, per @ArendPeter's comment — the word stands as the page had it, and this PR no longer touches that line.


docs/help/paper_ballots.md
```diff
@@ -25,7 +25,7 @@ Use the chart above to plan out the flow of your election. You should have a pla
 ## Step 2: Identify your voters
 Does your election have a finite number of possible voters? Do you know their identities? Is your election open to all who show up? Regardless, registering or credentialing your voters is key to ensuring that each individual is who they claim to be and ensuring that individuals only vote once.
 
-Advanced registration or in-person credentialing at the time of your election are two good options for confirming voter identity.
```
ArendPeter, line 29: I think advanced is correct here. Advance implies voting "in advance", but we're saying it's an advanced feature

ArendPeter (changes requested):
Left a small comment, otherwise looks good!

docs/help/paper_ballots.md
masiarek: Reverted in 694de5ad — the line reads "Advanced registration or in-person credentialing at the time of your election…" again, exactly as it did before this PR.

(For the record on what prompted it: the page's own heading further down says "Remote registration **without advance registration**", and the sentence contrasts registering beforehand with credentialing "at the time of your election". But it's your page's voice, so I've left the word as you had it.)

Every other fix in the pass is unchanged; I've updated the description so it lists nine fixes rather than ten.

ArendPeter (approved):
Thanks for making the changes!

PR #1501, opened 2026-08-08
docs: let contributors preview the documentation site locally

masiarek:
Editing through the GitHub web UI stays the right default for most doc changes — a typo, a sentence, a link fix. Nothing here changes that flow, and [Adding Documentation](https://docs.bettervoting.com/contributions/writers/4_adding_documentation.html) still leads with it.

This adds a second option for the larger cases: creating new pages, or moving things around. There, three things aren't visible in the browser:

- **Front matter can't be checked.** `parent:` has to match the parent page's `title:` character-for-character, emoji included — the guide already warns "Make sure this matches exactly otherwise your file won't be accessbile!" You currently find out by pushing and waiting for Pages.
- **GitHub's Preview tab isn't this site.** It renders GitHub-flavored Markdown, not Jekyll + just-the-docs — no theme, no nav placement, no Liquid.
- **Nav ordering is only visible rendered**, since it's relative to every sibling page.

So: a one-command local preview, for when you want it.

```bash
cd docs
docker run --rm -v "$PWD":/site -w /site -p 4000:4000 ruby:2.7 \
  sh -c "bundle install && bundle exec jekyll serve --host 0.0.0.0 --port 4000"
```

Then open <http://localhost:4000>. Edits rebuild automatically.

Worth noting the docs aren't part of `docker-compose.yml` — that runs the app (backend, database, Keycloak). Pages builds these straight from `docs/`, so previewing them means running Jekyll separately.

## What's here

**`docs/Gemfile`** — installs the `github-pages` gem, the same one Pages runs, so a local build matches production. `ffi` is pinned `< 1.17` because newer versions require Ruby >= 3.0 and won't resolve against the Ruby 2.7 that `github-pages`' Jekyll 3.9 expects.

**`docs/_config.yml`** — adds `repository:`. `jekyll-github-metadata` refuses to build without knowing which repo it is, and normally reads that from the `origin` remote; the command above mounts only `docs/`, so `.git` sits outside the container. Setting it in config makes the preview work with no env var. Inert in production — Pages already knows the repo, and `site.github` isn't referenced anywhere in these templates. *(Happy to drop this and document `-e PAGES_REPO_NWO=...` on the command instead, if you'd rather keep published config untouched.)*

**`docs/.gitignore`** — keeps `_site/`, `.jekyll-cache/` and `Gemfile.lock` out of the repo.

**`4_adding_documentation.md`** — a "Previewing your changes locally" section, added after the existing guidance rather than replacing it: the command, a named-volume variant that caches gems between runs, and troubleshooting for the three errors you actually hit (the missing repo name, the `ffi`/Ruby conflict, and the unmaintained `jekyll/jekyll` images whose expected tags no longer exist).

## Second commit: a small bug this turned up

The "Bold" example on *Updating Website Text* currently renders on the live site as:

```
voting_end: '**Voting ends on **'
```

The `{{datetime, datetime}}` is missing. Jekyll runs Liquid over page source before Markdown, and that happens inside fenced code blocks too — so Liquid parses the `{{...}}` as an output tag, fails on the comma, and substitutes nothing. Since that page is what teaches contributors how templated text works, the example is worth having survive.

The same page already handles this correctly in its other two examples (lines 31 and 38), which are wrapped in `{% raw %}` — this applies the same wrapper to the third. Jekyll prints `Liquid Warning: Liquid syntax error (line 62): Expected end_of_string but found comma` on every build.

Split into its own commit in case you'd prefer it as a separate PR.

## Verified

Full site builds and serves at `localhost:4000` with the theme, logo, colour scheme, search and nav intact, and every existing page returning 200. Cold start installs the gem set (a few minutes); with the named-volume cache, rebuilds are ~7 seconds.


ArendPeter (approved):
Looks good!

PR #1542, opened 2026-08-17
docs: answer "is my vote secret?" from what the code actually does

masiarek:
## Description

Adds `docs/help/is_my_vote_secret.md`.

This is the question voters ask most, and the site had no answer — so voters were left to assume. The answer turns out to be **good, and specific enough to state plainly** rather than reassure vaguely. All of it read out of the controllers:

- **`getBallotsByElectionIDController` scrubs `user_id`, `ip_hash`, `date_submitted`, `create_date`, `update_date` and `history`** from every ballot before returning it. Those fields aren't hidden from an admin — they aren't sent.
- It then calls **`secureShuffle`**, Fisher–Yates over `crypto.randomInt` (not `Math.random`). The comment in `controllerUtils.ts` states the threat model outright: without it an admin could correlate ballots to voters by submission time using the roll's own timestamps.
- **`getAnonymizedBallotsByElectionIDController`** emits only `ballot_id`, `election_id`, `precinct`, `votes`, from a cursor that returns rows in random order for the same reason.
- **`getBallotByBallotIDController`** applies the same scrubbing, so a voter opening their receipt sees their ballot without identity attached.
- Access is gated **twice**: `permissions.canViewBallots` (system_admin, owner, admin, auditor — notably *not* credentialer) **and** a state condition, since ballots are unavailable while an election is open unless public results are on. An organiser cannot quietly watch votes arrive.
- What an organiser **can** see is *who* has voted (`VoterAuth.has_voted`). That's not a leak — it's how reminders to non-voters work — and the page says so rather than glossing it.

The page also states the **limits** honestly, because a secrecy page that only reassures isn't worth reading: an email receipt shows your own ballot to anyone with your inbox; a three-voter race decided 3–0 reveals everyone's vote by arithmetic; and none of this touches what an organiser learns by asking you in person.

**No secrecy weakness was found while writing this.** The design is sound and visibly deliberate — the shuffle exists specifically to close the correlation attack.

**Merge ordering: no dependency.** Links only to pages on `main`.

n/a — documentation.

## Related Issues

None.


docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
+---
+layout: default
+title: Is My Vote Secret?
+nav_order: 8
+parent: BetterVoting Documentation
+---
+
+# Is My Vote Secret?
+
+It's the question voters ask most, and it deserves a straight answer rather than a reassuring one.
```
ArendPeter, line 10: This sentence feel too AI coded, I recommend removing it

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+**The short version: nobody — including the person running your election — can see how *you* voted. In an election with a voter list, they can see *that* you voted.** Those two things are separated deliberately, and this page explains how.
+
+## What the organiser can see
+
+**They can see who has voted.** In an election with a voter list, each voter is marked as having voted or not. This isn't incidental — it's what makes reminders work, so that only people who haven't voted get chased.
```
ArendPeter, line 16: Replace "This isn't incidental — it's what makes reminders work, so that only people who haven't voted get chased." with "This allows the admin to send reminder to those who haven't voted yet"

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+**They can see the ballots.** Once an election has closed (or if the organiser turned on public results), they can look at the ballots that were cast.
```
ArendPeter, line 18: Replace

> They can see the ballots.

with

> They can see anonymized ballot.

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+**They cannot connect the two.** When ballots are handed out for viewing, the following are stripped from every one of them:
+
+* the voter's user ID
+* the hash of the voter's IP address
+* the time the ballot was submitted
+* the ballot's creation and update times
+* the ballot's edit history
```
ArendPeter, line 26: I think this is TMI, and opens more questions for the voters. I'd do "When ballots are handed out for viewing any personal identifiable information is stripped and what's left is the votes themselves. Nothing says who cast the votes"

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+What's left is the votes themselves, and nothing that says whose they are.
+
+## Why the order is shuffled
+
+Removing names isn't enough on its own, and BetterVoting doesn't stop there.
+
+If ballots came back in the order they were cast, an organiser could line that order up against the voter list — which records when each person voted — and work out who cast which ballot. Names would be gone, but anonymity wouldn't be.
```
ArendPeter, line 34: Replace

> If ballots came back in the order they were cast, an organiser could line that order up against the voter list — which records when each person voted — and work out who cast which ballot. Names would be gone, but anonymity wouldn't be.

with

> If ballots came back in the order they were cast, a bad actor could try to infer the voter identity from that order.

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+So **ballots are deliberately returned in a random order**. The shuffle uses a cryptographic random number generator, not an ordinary one, specifically so the sequence can't be predicted or reconstructed. The same protection applies to the anonymised ballot data that gets published with public results, which carries only a ballot ID, the election, the precinct, and the votes.
```
ArendPeter, line 36: Replace

>. The shuffle uses a cryptographic random number generator, not an ordinary one, specifically so the sequence can't be predicted or reconstructed. The

with

> , and the

For a lay reader, I think just mentioning shuffling is sufficient

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+Not everyone with access to an election can.
+
+| Role | Can view ballots? |
+|:---|:---|
+| Owner, Admin, Auditor | Yes — after the election closes, or if public results are on |
+| Credentialer | No |
+| Everyone else | Only if the organiser turned on public results |
+
+And there's a timing rule on top: **the full set of ballots is not available to anyone while an election is still open**, unless the organiser has deliberately made results public. An organiser cannot quietly watch the votes arrive during a live election.
```
ArendPeter, line 48: I'd remove this "Who can look at ballots at all" section. The rules are the same for everyone, the ballots aren't available until the results are public, but that's already mentioned earlier, so this section doesn't add very much

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+## What this does and doesn't protect against
+
+Being straight about the limits is part of the answer.
+
+**It does protect against** the organiser — or an admin or auditor — looking up how a particular person voted. The identifying fields aren't hidden from them; they aren't sent at all.
```
ArendPeter, line 54: Replace

> the organiser — or an admin or auditor — 

with

> a bad actor

docs/help/is_my_vote_secret.md
```diff
@@ -0,0 +1,68 @@
...
+
+**It doesn't make you anonymous to yourself.** If you got an email receipt, that link shows you your own ballot. Anyone who gets into your email could see it too, so treat the receipt like any other private message.
+
+**It doesn't cover a very small electorate.** If a race has three voters and the result is 3–0, everyone knows how everyone voted. That's arithmetic, not a software failure, and no system can prevent it.
+
+**It doesn't change what your organiser knows outside the system.** If they asked you in person how you were going to vote, BetterVoting can't help with that.
```
ArendPeter, line 60: I'd remove this sentence

ArendPeter (changes requested):
Left some requested changes :). I've started using the "replace x with y" format to make reviewing faster, but let me know if you'd like more context on any of the decisions.

ArendPeter (approved):
Comments have been fixed, will merge!

Issue #1556, opened 2026-08-17
Documentation program: ~20 new help pages for voters and administrators (28 PRs)


masiarek:
## What this is

A documentation and localisation program for the help site and the app's own text, submitted as a batch of pull requests. This issue is the index: what each PR does, a suggested merge order, the structure they assume, and the decisions taken along the way.

**The short version:** the help site had 7 pages, all written for the person *running* an election. There was nothing addressed to voters — who are the larger audience, and who the app links into these docs from `en.yaml`. This adds roughly 25 pages, gives both audiences a front door, documents features that had no written explanation anywhere, and takes the app from 4 languages to 7.

Nothing here needs to merge as a set. Every page PR links only to pages already on `main`, precisely so merge order never matters.

---

### 1. Two one-line config changes — merge first, they make everything cheaper

| PR | What | Why first |
|---|---|---|
| #1535 | enable `jekyll-redirect-from` | Without it, `redirect_from:` is silently discarded and **renaming any published page is a permanent 404**. Eight doc URLs are hardcoded in the app. This is the safety net for every future change. |
| #1536 | enable mermaid | One key. Lets a diagram render instead of publishing as its own source text. |

Neither installs anything — both ship inside the `github-pages` bundle already in use.

### 2. Structure

| PR | What |
|---|---|
| #1543 | Gives the site a voter door and an admin door. Front-matter only — **no file moves, no URL changes**. Supersedes #1505, which can be closed. |

### 3. Small fixes

| PR | What |
|---|---|
| #1493 | spelling and grammar on the paper ballots page |
| #1503 | the Montserrat font URL is malformed, so headings never use it |
| #1501 | lets any contributor preview the docs site locally |
| #1557 | **the Adding Documentation page tells contributors to copy a `parent:` value that doesn't work** — see below |

### 4. Pages for voters

| PR | Page |
|---|---|
| #1539 | **How to Vote** · **After You Vote** |
| #1542 | **Is My Vote Secret?** |
| #1550 | **Voting Problems** — symptom → cause → fix |
| #1544 | **How Voters Get Access** |
| #1561 | **Using BetterVoting in Another Language** |

### 5. Pages for administrators

| PR | Page |
|---|---|
| #1537 | **Election or Poll?** |
| #1538 | **Choosing a Voting Method** |
| #1540 | **Before You Open Your Election** · **Electing More Than One Winner** |
| #1546 | **Managing Your Voters** |
| #1551 | **Emails to Voters** |
[...]

PR #1557, opened 2026-08-17
docs: fix the parent example that makes a new page vanish

masiarek:
## Description

**The Adding Documentation page warns:** *"Make sure this matches exactly otherwise your file won't be accessible!"* — and then gives an example that doesn't match.

The Writers section's real title is emoji + **space** + variation selector + `Writers`. Both places this page shows the value — the example front-matter block (line 31) and the bullet explaining it (line 40) — omit that space.

Byte comparison:

| | bytes |
|---|---|
| Real title in `3_writers.md` | `e2 9c 8d` **`20`** `ef b8 8f` … |
| The documented example | `e2 9c 8d` `ef b8 8f` … |

So a contributor who follows the documentation **exactly** gets a page that silently disappears from the navigation — no build error, nothing to search for. It's the precise failure the page warns about two paragraphs earlier, caused by the page itself.

Both occurrences now carry the byte-identical value, taken from this same file's own working front matter. The bullet also gains the rule that prevents the whole class of bug rather than this one instance: **copy the parent's title rather than typing it**, because titles can contain emoji, variation selectors and spacing that are invisible on screen.

### Three additions, from things that have actually caught people out

- **Links between pages.** Use the `.md` filename — a plugin rewrites it to `.html` and it keeps working when read on GitHub. The catch worth stating: **the rewrite only fires when the target exists.** Link a page that hasn't merged and the raw `.md` href ships to the published site. So add cross-links *after* a page merges.
- **Renaming and moving.** Think twice once published — the address comes from the filename and several doc URLs are hardcoded in the app — but since #1535 it is **not a one-way door**: the section shows the `redirect_from:` front matter that keeps the old address alive, and names the two things a redirect does not cover (a renamed *heading*, since two app links point at one; and cross-links inside other doc pages). And **reorganising the sidebar is free**, because the nav is built from `parent`/`nav_order` rather than file location. Worth saying, since that's the thing people actually want and assume they can't have.
- **Literal curly braces need a `raw` block.** The template engine reads them before rendering and swallows the text. This has silently broken a published page before — including, it turned out, this section's own first draft (see below).

Also repoints the docs directory link from `star-server` to `bettervoting`, the repository's current name.

**Verification.** Built locally with the `github-pages` gem (Jekyll 3.9), the toolchain from #1501:

- the page renders under the Writers section, and the build is warning-free;
- `redirect_from: [/help/tie_breakers.html]` on `ties.md` writes `_site/help/tie_breakers.html` containing `<meta http-equiv="refresh" content="0; url=/help/ties.html">`, which is the behaviour the new section documents.

n/a — documentation.

## Related Issues

Part of the documentation program indexed in #1556.

---

**Update (review), two commits:**

- **Merged `main`** — the file had gone conflicting against the "Previewing your changes locally" section added by #1501. That section is byte-unchanged and stays where it is; this branch's sections follow it.
- **Corrected two of the new sections.** *Renaming and moving pages* claimed there is no redirect from an old address to a new one; #1535 made that false, so it now documents `redirect_from:` instead. And *Curly braces* fell into the trap it documents — Liquid ate its own example, publishing "literal double curly braces — — wrap that part in and ````" and logging a syntax warning. It is now escaped so the tags survive, with a comment in the source explaining why it looks the way it does.


docs/contributions/writers/4_adding_documentation.md
```diff
@@ -37,9 +37,25 @@ These variables indicate the title of the document as well as where it's located
...
 
+## Links between pages
+
+Write links to other documentation pages using the **``.md`` filename**, like ``[Ties](ties.md)`` — a build plugin rewrites those to the published ``.html`` address automatically, and writing them this way means the link also works when someone reads the file on GitHub.
+
+One catch worth knowing: that rewrite only happens when the target file **exists**. If you link to a page that has not been added yet — including one that is still sitting in an unmerged pull request — the link is left exactly as you wrote it and will not work on the published site. So add cross-links to a new page *after* it has been merged, not before.
+
+## Renaming and moving pages
+
+**Avoid it if the page is already published.** A page's address comes from its filename, several documentation links are built into the BetterVoting app itself, and there is currently no redirect from an old address to a new one — so a rename turns every existing link into a dead one.
```
ArendPeter: Is this still true with the jekyll-redirect-from PR?

ArendPeter (changes requested):
Looks good, just a minor comment

docs/contributions/writers/4_adding_documentation.md
masiarek: Not any more — good catch. That paragraph was written before #1535 merged, and #1535 is exactly what makes it false.

Rewritten in 293b1e73. The section now shows the front matter that keeps an old address alive:

```yaml
redirect_from:
  - /help/tie_breakers.html
```

…and names the two things a redirect still does **not** cover: a renamed **heading** (two of the app's links point at one — `ties.html#random-tie-breakers` is shown to voters when a tie is broken), and cross-links written inside other doc pages, which are worth repointing anyway.

**Verified against a real build** (`github-pages` gem, Jekyll 3.9, the local preview from #1501): adding that key to `ties.md` writes `_site/help/tie_breakers.html` containing `<meta http-equiv="refresh" content="0; url=/help/ties.html">`.

Two other things rode along in this push:

**1. Merged `main` in.** The file had gone conflicting — `main` has since gained the "Previewing your changes locally" section from #1501, in the same place this branch adds its sections. Main's section is byte-unchanged and stays where it is; this branch's three sections follow it. PR is mergeable again.

**2. Fixed the "Curly braces" section, which fell into the trap it documents.** Written plainly, Liquid ate its own example before Markdown ever saw it. The published sentence came out as:

> If you need to write literal double curly braces — — wrap that part in and \`\`\`\`.

with `Liquid Warning: Liquid syntax error (line 97): Unexpected character ` in "{{`` and ``}}"` in the build log — the braces gone, and `raw`/`endraw` swallowed by the very tags they name. The example is now escaped so the tags survive to the page, with a comment in the source saying why it's written that way, and the build is warning-free. The published block now reads:

```
{% raw %}
Your total is {{ total }}
{% endraw %}
```

Worth knowing for anyone hitting this later: the obvious escape `{{ "{% raw %}" }}` does **not** work on Jekyll 3.9 — Liquid 4 matches `{%…%}` before `{{…}}`, so it fails the build with "Variable '{{ "{% raw %}' was not properly terminated". Only the opening brace can be emitted as a string.

ArendPeter (approved):
Thanks for making the change!

PR #1561, opened 2026-08-17
docs: explain how to use BetterVoting in another language

masiarek:
## Description

Adds `docs/help/using_another_language.md`.

**BetterVoting ships four languages and there is no way to discover that from the interface.** `i18n.ts` uses the browser language detector, and nothing in `packages/frontend/src` offers a language switcher — `i18n.language` is used only for list formatting. The sole way to choose a language is the `?lng=` parameter, which is documented for *translators* in the contribution guide and nowhere a voter would ever look.

Covers: detection is automatic and most people need nothing; the codes for the four current languages; how to append the parameter when the link already has a query string; and how to help add a language.

### Three things the page is careful about

**The choice is cached.** `i18next-browser-languagedetector`'s default order writes the selection to localStorage and a cookie, so it persists after the parameter is gone — and getting back to English needs an explicit `?lng=en`, not just dropping the parameter. This gets its own warning callout; it's the most confusing behaviour here and the one most likely to generate a support message.

**Mixed English is correct, not broken.** Only Priority 0 is translated and `fallbackLng` is `'en'`, so a page can legitimately be part-translated. Saying so plainly stops a voter reporting it as a fault.

**The election's own content is never translated.** Candidate names, race titles, descriptions and the organiser's emails appear exactly as written. A Spanish interface can carry an English ballot question — that's the system working correctly, and it isn't obvious.

The admin section draws the practical conclusion: sending `?lng=` links is worth doing for a known-language group, but it translates the buttons *around* your text, not the text itself — so a genuinely bilingual electorate is better served by writing the race description in both languages.

**Language list.** Lists only the four currently on `main`. German is in review (#1560) and can be added in a one-line follow-up once it merges.

**Verification.** Local build (`github-pages`, Jekyll 3.9): renders, sits under the `BetterVoting Documentation` parent, no unrewritten `.md` hrefs.

n/a — documentation.

## Related Issues

Part of the documentation program in #1556. Complements #1560 (German translation).

### DIFF
--- a/docs/help/using_another_language.md
+++ b/docs/help/using_another_language.md
@@ -17,11 +17,11 @@ So most people never need this page. If the site is already showing your languag
 
 ## Choosing a language yourself
 
-There is no language menu on the site. To pick a language, **add `?lng=` and the language code to the end of the address**:
+There is no language menu on the site. To pick a language, **take the address you already have and add `?lng=` and the language code to the end of it**.
 
-```
-https://bettervoting.com/your-election-id?lng=es
-```
+So if your voting link is `bettervoting.com/a1b2c3`, you'd open `bettervoting.com/a1b2c3?lng=es` for Spanish.
+
+The examples below use `a1b2c3` to stand in for your own election's address — replace it with yours. Opening the examples as written will give you an "Election not found" error, because no such election exists.
 
 | Language | Code | Add to the address | |
 |:---|:---|:---|:---|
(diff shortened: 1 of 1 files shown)


docs/help/using_another_language.md
```diff
@@ -0,0 +1,76 @@
...
+
+## It usually happens by itself
+
+**BetterVoting reads the language setting from your browser and matches it if it can.** If your device is set to Spanish and Spanish is available, you'll see Spanish without doing anything.
+
+So most people never need this page. If the site is already showing your language, nothing here applies to you.
+
+## Choosing a language yourself
+
+There is no language menu on the site. To pick a language, **take the address you already have and add `?lng=` and the language code to the end of it**.
```
ArendPeter, line 20: I can get a language selector added soon. I'll block this PR behind that feature

PR #1562, opened 2026-08-17
Fill Priority 0 translation gaps in es, pl, and pt-BR

masiarek:
## Description

The three existing translation files had drifted behind `en.yaml` and were missing Priority 0 strings. This PR brings **es**, **pl**, and **pt-BR** up to full Priority 0 coverage (255 leaf strings), the same coverage the new German translation (#1560) ships with. Per the translation guide, only Priority 0 was touched — nothing after the `#### PRIORITY 1` banner.

**Before → after (Priority 0 keys present):**

| file | before | after | added |
|---|---|---|---|
| es.yaml | 212 | 255 | 44 |
| pl.yaml | 205 | 255 | 51 |
| pt-BR.yaml | 205 | 255 | 51 |

What was added in all three: the whole new `election_history` section (15 keys), the draggable-ballot UI strings, `dialog_abstention`, `share.button_results`, the `ballot_submitted` status/receipt block, `results.admin_results_toggle` / `not_enough_candidates` / `win_long_title_prefix` / `admin_title` / `rcv.bloc_results_title` / `star.runoff_no_preference_footnote`, the third `star.tiebreaker_note_text` paragraph, and the `keyword` `races` / `race_title` / `votes` / `close` vocabulary. pl and pt-BR additionally gained the `ballot.methods.stv` block and `results.stv.table_title`.

**Interpolation-token repairs to existing strings** (tokens now match English exactly in every key): `!tip(random_tie_order)` restored to `results.random_tiebreak_subtitle`, `{{n}}` to `results.star_pr.chart_title`, and `{{candidate}}` to the first star_pr instruction bullet in all three files; the Spanish `ranked_robin.footer_single_winner` had a duplicated `{{candidate}}` (now "el oponente"). Polish `number.rank_ordinal_*` entries were still English-shaped (`1st`/`2nd`/`3rd`) and now use the Polish ordinal form `{{count}}.` (es and pt-BR ordinals were already correct for their languages).

**Verification** (script flattens each file and diffs key paths and sorted token lists against Priority 0 of `en.yaml`):

```
en Priority-0 leaves: 255
es:    has 256 | missing 0 | extra 1 | token mismatches 0
pl:    has 256 | missing 0 | extra 1 | token mismatches 0
pt-BR: has 256 | missing 0 | extra 1 | token mismatches 0
```

The 1 "extra" in each file is `results.single_vote`, a stale leftover: it no longer exists in `en.yaml` and nothing in the frontend references it. It was deliberately left in place rather than deleted as a side effect of this PR — happy to drop it (in all three files) if you prefer. Key insertion order matches `en.yaml`, so the files stay diffable against the source; keys already translated were not re-translated.

### Honesty note — these are AI-produced translations

This PR was produced with AI assistance (Claude). The translation guide asks for a native-speaker proof-reader, and that review has **not** happened yet — please treat every added string as needing native review before these are relied on. Specific places a reviewer should look first:

- **Polish grammatical case:** interpolated nouns arrive in nominative, but several English templates use them as objects. `ballot.methods.stv.instruction_bullets[0]` ("Uszereguj {{candidates}} …") needs accusative *kandydatów*, so it will read wrong once `keyword.election.candidates` is translated; bullets 1–2 and `results.admin_results_toggle` were rephrased ("{{capital_election}} - wyniki są publiczne") to keep the noun in nominative. `share.button_results` uses a colon for the same reason. A native speaker may find better constructions.
- **Gender agreement (es/pt-BR):** the star_pr bullet now reads "su(s) {{candidate}}(s) favorito(s)" / "seu(s) {{candidate}}(s) favorito(s)" — masculine agreement is correct for *candidato* but not for the poll vocabulary's *opción/opção*. English's own "(s)" hedge was mirrored; there is no phrasing that agrees with both vocabularies.
- **New terminology minted here:** "Registro de Auditoría / Registro de Auditoria / Dziennik audytu" for *Audit Log*; "Recinto" (es) / "Zona eleitoral" (pt-BR) / "Obwód" (pl) for *Precinct*; "herramienta de emergencia / ferramenta de emergência / narzędzie awaryjne" for the *break-glass tool*; `keyword.election.race_title` ("elected office title") in all three. These had no precedent in the files.
- **Polish `keyword` and other pre-existing untranslated strings:** much of pl.yaml's older content is still English placeholder text; per the no-re-translation scope, only the missing keys were added (in Polish), so some sections are now mixed-language until someone translates the rest.
- The es/pt-BR third `tiebreaker_note_text` paragraph slightly overlaps their older two-paragraph note, which already mentions the tiebreaker-protocol link; English's current three-paragraph text repeats it, and list lengths must match.

Text-only translation changes; no visual/layout changes. (Strings render in the same components as the existing English keys.)

## Related Issues

- Follows up on #1560 (German translation, full Priority 0 coverage) — this brings the three pre-existing languages up to the same coverage.

### DIFF
--- a/packages/frontend/src/i18n/es.yaml
+++ b/packages/frontend/src/i18n/es.yaml
@@ -15,6 +15,26 @@ election_home:
   drafted: Esta elección está en fase de elaboración
   archived: Esta elección se ha archivado
 
+election_history:
+  title: Registro de Auditoría de la Elección
+  subtitle: Un registro público de todo lo que ha ocurrido con esta elección desde que fue finalizada.
+  ladder_note: 'Nota: los conteos de boletas se reportan solo en hitos específicos (1, 5, 10, 25, 50, 100, 250, ...) para proteger la privacidad de los votantes, y las ediciones de boletas usan una escala más fina (1, 2, 5, 10, 25, ...). Por la misma razón, los eventos relacionados con boletas se reportan con precisión de día.'
+  link: Registro de Auditoría
+  loading: Cargando el registro de auditoría…
+  empty: Aún no se han registrado eventos.
+  not_finalized: Esta elección aún no ha sido finalizada, por lo que no hay registro de auditoría disponible.
+  # Column headers and event chip labels live in EnhancedTable's head cell pool,
+  # which keys its group filters and chip colours off plain English strings.
+  desc:
+    state_initial: 'La elección entró en el estado "{{to}}"'
+    state_change: 'El estado cambió de "{{from}}" a "{{to}}"'
+    preliminary_on: Resultados preliminares habilitados
+    preliminary_off: Resultados preliminares deshabilitados
+    ballots_milestone: 'Las boletas emitidas alcanzaron {{count}}'
+    upload_ballots: 'Un administrador cargó {{count}} boleta(s)'
+    edits_milestone: 'Las ediciones de boletas alcanzaron {{count}}'
+    voter_revealed: Un administrador usó la herramienta de emergencia para revelar el ID de un votante
+
 # Ballot (ex. https://bettervoting.com/j87vqp/vote)
 ballot:
   this_election_uses: Esta {{election}} usará el {{voting_method}} para eligir a {{spelled_count}} $t(winner.winner).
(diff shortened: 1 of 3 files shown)


ArendPeter (changes requested):
Requested removing es and pt-BR

PR #1571, opened 2026-08-18
Stop the results endpoint 500ing on a race with no ballots

jacksonloper:
`https://bettervoting.com/mkvb4f/results` currently returns a 500. Two tabulators throw when a race has no votes in it, and since `getElectionResults` tabulates every race in one loop with no per-race error handling, either throw takes down the whole page for all seven races.

## The IRV crash

From the production log for that request:

```
ctx:c254e80d Tabulating results for Plurality election
ctx:c254e80d Tabulating results for IRV election
ctx:c254e80d Cannot read properties of undefined (reading 'hareScores')
ctx:c254e80d RES: GET /API/ElectionResult/mkvb4f  status:500
```

With no active votes nobody can reach the quota, so every round eliminates another candidate. Once the last one is gone the loop keeps running and reads `remainingCandidates[0]` off an empty array, throwing at `IRV.ts:99`.

STV never reaches this: the branch below seats the survivors as soon as they can fill the remaining seats, and that branch is gated on `proportional`. #1427 fixed the STV shape of this bug inside `distributeVotes` and its test notes that *"plain IRV can't hit this because its winner branch rebuilds the candidate pool"* — true right up until the pool empties by elimination instead.

**This is not limited to multi-winner races.** An ordinary single-winner IRV race that nobody has voted in yet crashes identically, so any open election with public preliminary results 500s its results page until the first vote arrives. There is a regression test for that case specifically.

## The STAR_PR crash behind it

Fixing IRV alone was not enough — the same page then threw somewhere else. STAR_PR builds its ballot table from the cast ballots, so with none it read `weighted_score` off an empty array in `findSplitPoint`, and past that divided the surplus by a zero `weight_on_split`, which throws in fraction.js. Both are guarded.

The `weight_on_split` guard is not only about empty races: it can reach zero with real ballots once the ballots on the split point have been spent, and that threw too.

## Behaviour note

IRV now seats fewer winners than `num_winners` when every ballot is exhausted. That is the honest answer -- there is nobody left with support to seat.

For the empty-ballot case this is invisible to users: `Results.tsx` gates the winners block on `nTallyVotes > 0`, so a race with no votes renders `waiting_for_results` ("Still waiting for results / No votes have been cast") regardless of what `elected` contains. That is the message the page was always meant to show here; it just never got the chance, because the API 500'd before the frontend had anything to render.

It is visible in the rarer case where ballots exist but all exhaust before the seats are filled (see the third new IRV test: three ballots, all ranking only Alice). There the page shows the winners that could actually be seated and no placeholder for the rest, which still seems right, but flagging it since it is a real change.

Note a ballot-count short-circuit in the controller would not have been sufficient on its own -- that exhaustion case has `nTallyVotes > 0` and crashed identically. Happy to add one as belt-and-braces if wanted.

I also considered wrapping each race in a try/catch so one bad race degrades to an error for that race instead of 500ing all seven. That is a bigger behavioural change and I did not want to bundle it here, but it would have contained this incident.

## Verification

- New `NoBallots` suite walks `VotingMethods`, so any method added later has to survive an empty ballot set.
- Reverting both source fixes with the tests in place: 5 tests fail with the exact production error, and pass with the fixes.
- Full backend suite: 24 suites, 174 tests passing. `npm run build` clean.

### DIFF
--- a/packages/backend/src/Tabulators/AllocatedScore.ts
+++ b/packages/backend/src/Tabulators/AllocatedScore.ts
@@ -149,7 +149,11 @@ export function AllocatedScore(candidates: candidate[], votes: rawVote[], nWinne
         let weight_on_split = findWeightOnSplit(cand_df, split_point);
 
         // quota = spent_above + weight_on_split*new_weight
-        let new_weight = (quota.sub(spent_above)).div(weight_on_split);
+        // weight_on_split is the total weight sitting exactly on the split point, so
+        // when it is zero there is no partially represented ballot to reweight and the
+        // surplus has nowhere to go. Dividing by it throws in fraction.js, which used to
+        // 500 the results endpoint for any race with no ballots.
+        let new_weight = weight_on_split.equals(0) ? new Fraction(0) : (quota.sub(spent_above)).div(weight_on_split);
         results.logs.push(
             `The ${rounded(weight_on_split)} voters who gave ${summaryData.candidates[w].name} ${rounded(split_point.mul(MAX_SCORE))} stars are partially represented. `+
             `${percent(new_weight)} of their remaining vote will go toward ${summaryData.candidates[w].name} and ${percent(new Fraction(1).sub(new_weight))} will be preserved for future rounds.`)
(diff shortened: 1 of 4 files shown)


ArendPeter (approved):
Looks good, thanks for making the fix!

PR #1583, opened 2026-08-21
chore(shared, backend): upgrade sanitize-html dependency

JonBlauvelt:
## Description
- Upgrades sanitize-html issue and resolves ecma import issued noted in original pr https://github.com/Equal-Vote/bettervoting/pull/1467
- Removes unused dependencies 
- Upgrades jest and related deps
- Moves dev deps to correct dep level


ArendPeter (approved):
Looks good, thanks!

PR #1594, opened 2026-09-02
Adjust nav bar

ArendPeter:
## Description

Move Vote in a Poll down and add Paper Ballots in the features


SaraWolk (approved):
Nice changes to top nav and feature list

PR #1611, opened 2026-09-08
Upload ballots refactor

### LINKED ISSUE #1604: Enforce allowed_submit_types in the ballot-submission pipeline
## Parent

Part of #1603.

## What to build

The ballot-submission pipeline enforces which submission channels an election allows, independent of the existing `canUploadBallots` role check. An admin-submitted (or browser-submitted) ballot is rejected with a clear 400 error when its submission type isn't in the election's `allowed_submit_types`, and accepted when it is — including the default case where the field is absent entirely (defaults to browser + Discord allowed, paper/admin-upload off).

This is the enforcement layer only — no new UI. It's verified through the ballot-submission API contract and the settings-validation function's accept/reject behavior, not by exercising any dialog.

## Acceptance criteria

- [ ] `BallotSubmitType` (`'submitted_via_browser' | 'submitted_via_admin' | 'submitted_via_discord'`) and a new `BallotActionType` (currently an alias of `BallotSubmitType`) move from `castVoteController.ts` into the shared `Ballot` domain model; the existing untyped `action_type` field on ballot history entries retypes to `BallotActionType`.
- [ ] `NewBallot` gains `election_id`, `status`, and `date_submitted` as optional fields (joining the already-optional `ballot_id`/`create_date`/`update_date`/`head`).
- [ ] New optional `ElectionSettings.allowed_submit_types?: BallotSubmitType[]` field. Every read site resolves it as `election.settings.allowed_submit_types ?? DEFAULT_ALLOWED_SUBMIT_TYPES`, where `DEFAULT_ALLOWED_SUBMIT_TYPES = ['submitted_via_browser', 'submitted_via_discord']`. No database migration needed (schema-less JSON settings column).
- [ ] `electionSettingsValidation` rejects a resolved `allowed_submit_types` that would end up empty.
- [ ] The ballot-submission pipeline (`makeBallotEvent` / the shared function all ballot writes funnel through) rejects a submission whose type isn't in the election's resolved `allowed_submit_types`, with a 400 naming the disallowed type.
- [ ] This check sits inside the same existing bypass block that already skips auth/roll/validation for a draft election or a `prior_election`-sourced election — never blocking draft-building or prior-election imports.
- [ ] The existing `canUploadBallots` role permission (system admin / election owner only) is unchanged and remains a fully independent gate on the upload endpoint — this new setting doesn't replace it.
- [ ] The local constant hardcoding `'submitted_via_admin'` in the admin-upload code path is typed against the new shared `BallotSubmitType`.
- [ ] Backend tests: an admin-submitted ballot is rejected (400) when the election's settings exclude `submitted_via_admin`, and accepted when included (including the default-array case where the field is absent). `electionSettingsValidation` rejects an empty-resolving `allowed_submit_types`. The existing closed-election already-voted/per-row-rejection behavior for bulk upload is exercised in combination with the new setting to confirm the two checks compose. Follow the existing pattern: build a fixture `Election` with the relevant `settings`, drive it through the real controller against the test database.

## Blocked by

None — can start immediately.

### LINKED ISSUE #1605: Extract shared upload utilities and rewire Upload Elections onto them
## Parent

Part of #1603.

## What to build

Extract the race-order computation, ballot-encoding, and batched-upload-with-retry logic that today lives inline inside the Upload Elections tool (`packages/frontend/src/components/Election/Upload/UploadElections.tsx`) into a standalone, unit-tested module, then rewire Upload Elections onto it. This is a prefactor: it doesn't change Upload Elections' user-visible behavior, but it removes the one hand-rolled implementation of this logic so the upcoming Upload Ballots dialog (a later ticket) can reuse it instead of duplicating it.

## Acceptance criteria

- [ ] New shared module exports:
  - `computeRaceOrder(election): RaceCandidateOrder[]` — derived from the election's races/candidates, independent of any ballot row.
  - `encodeBallotRow(row, raceOrder): OrderedNewBallot` — looks up each race in a row's votes by race ID (not positional index); a race missing from a row's votes encodes as all-undecided for that race.
  - `uploadBallotsBatched(electionId, raceOrder, ballots, onProgress, options)` — batch/retry/progress loop matching today's tuning (batch size 700, shrinking 25% per failure down to a floor of 10), returning partial results plus a synthetic failure entry for any untried remainder if the shrink floor is hit (today this silently discards progress on abort — that's a real bugfix bundled into the extraction).
- [ ] Upload Elections is rewired onto this module: its inline race-order computation, ballot encoding, and hand-rolled batch/retry loop are deleted — no duplicate implementation left behind. Its election-creation/inference logic and post-reupload ballot-deletion logic (public-archive-only concerns) are untouched.
- [ ] Unit tests for `computeRaceOrder` and `encodeBallotRow`, including the case where a race is missing from a row (encodes as all-undecided).
- [ ] Manual/E2E regression check: Upload Elections still successfully imports a CVR file end-to-end after the rewire (no behavior change from a user's perspective).

## Blocked by

None — can start immediately.

ArendPeter:
## Description

This change begins refactoring the frontend in preparation for a upload ballots feature. It particular it adds an allowed_submit_types so that we can gate different voting submissions (web vs admin upload vs discord), and starts to move the uploadElection functions into a shared utility. 

- fixes #1604 
- fixes #1605


packages/frontend/jest.config.cjs
```diff
@@ -0,0 +1,15 @@
+module.exports = {
```
ArendPeter: Adding unit tests to the frontend is a good idea, but I'd use vitest instead of jest

packages/shared/src/domain_model/Ballot.ts
```diff
@@ -5,6 +5,10 @@ import { Score } from "./Score";
 import { Uid } from "./Uid";
 import { OrderedVote, Vote } from "./Vote";
 
+export const BALLOT_SUBMIT_TYPES = ['submitted_via_browser', 'submitted_via_admin', 'submitted_via_discord'] as const;
+export type BallotSubmitType = typeof BALLOT_SUBMIT_TYPES[number];
+export type BallotActionType = BallotSubmitType;
```
ArendPeter, line 10: TODO: Insert comment explaining why these are identical. We expect to have more action types in the future, but for now the submit types are our only actions

package-lock.json
ArendPeter, line 1: This looks like it was deleted by accident

packages/frontend/src/components/Election/Upload/uploadUtils.test.ts
```diff
@@ -0,0 +1,146 @@
...
+            voting_method: 'IRV' as const,
+            num_winners: 1,
+            candidates: [
+                { candidate_id: 'c3', candidate_name: 'Charlie' },
+                { candidate_id: 'c4', candidate_name: 'Dana' },
+                { candidate_id: 'c5', candidate_name: 'Eve' },
+            ],
+        },
+    ],
+} as unknown as Election;
```
ArendPeter, line 42: What's going on here 👀

ArendPeter: I see, "as unknown as Election" is used when the original type doesn't quite match Election but it's close enough for the purposes of the test.

packages/frontend/src/components/Election/Upload/uploadUtils.test.ts
```diff
@@ -0,0 +1,146 @@
+import { computeRaceOrder, encodeBallotRow } from './uploadUtils';
+import { Election } from '@equal-vote/star-vote-shared/domain_model/Election';
+import { NewBallot, RaceCandidateOrder } from '@equal-vote/star-vote-shared/domain_model/Ballot';
+import { ORDERED_VOTE_TAIL_LENGTH } from '@equal-vote/star-vote-shared/domain_model/OrderedVoteCodec';
+
+const mockElection = {
```
ArendPeter, line 7: NOTE, I think this is fine for now, but for future reference this is somewhat redundant with our library of mocks on the backend. If we proceed with building out the frontend unit testing library, then we could consider sharing a library via the shared package, or defining a separate library for frontend

packages/frontend/src/components/Election/Upload/uploadUtils.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    it('handles a row with no votes at all (all races missing)', () => {
+        const row = makeRow([]);
+        const result = encodeBallotRow(row, raceOrder);
+        expect(result.orderedVotes).toHaveLength(2);
+        // Both races should be all-undecided
+        const race1Vote = result.orderedVotes[0];
+        expect(race1Vote[0]).toBeNull();
+        expect(race1Vote[1]).toBeNull();
```
ArendPeter, line 118: Shouldn't this also be validating race2Vote?

packages/backend/src/test/allowedSubmitTypes.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+const th = new TestHelper();
+
+afterEach(() => {
+    jest.clearAllMocks();
+    th.afterEach();
+});
+
+// A simple open election with one STAR race and two candidates.
+const makeElection = (allowed_submit_types?: string[]): Election => ({
```
ArendPeter, line 17: Consider using testInputs for this. You could add a new election, but it's even better if you can reuse one

packages/backend/src/test/allowedSubmitTypes.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        ...(allowed_submit_types !== undefined ? { allowed_submit_types } : {}),
+    } as ElectionSettings,
+} as Election);
+
+// race_order and a single ballot in ordered-vote format for the election above.
+const RACE_ORDER = [{ race_id: "race0", candidate_id_order: ["cand0", "cand1"] }];
+// orderedVote = [aliceScore, bobScore, overvote_rank, has_duplicate_rank]
+const ORDERED_BALLOT = { orderedVotes: [[5, 0, null, null]] };
+
+describe("electionSettingsValidation — allowed_submit_types", () => {
```
ArendPeter: Can we move electionSettingsValidation tests to the shared pacakge?

PR #1649, opened 2026-09-17
Stop storing the SendGrid response in election roll history

jacksonloper:
## Human Summary

Right now, it seems to me that we are storing email history that we don't need to store in the rolls.  That's bad because the rolls get copied on write, and get read for all sorts of important things.  Good news: we no longer need to store email history in the rolls, because in May we made a database for storing that info.  This may be somewhat important right now b/c I think a large election will blow up our servers right now because of all the email stuff in the rolls.  But if we take it out (and afaik there's no harm in taking it out) there will be fewer things that will break our servers for a large election.

## Summary

Removes one line: `email_data: emailResponse` from the `ElectionRollAction` appended on every email send in `sendEmailController.ts`.

**What was being stored.** The full SendGrid API response, per voter, per email. Measured on prod, each action is ~794 bytes:

| field | bytes |
| --- | --- |
| `headers` | 661 |
| `statusCode` | 3 |
| `body` | 2 (empty string) |

The `headers` object holds 15 HTTP headers. Exactly one — `x-message-id` — carries information. The rest are `access-control-*` (×4), `cache-control`, `connection`, `content-length`, `content-security-policy`, `date`, `referrer-policy`, `server`, `strict-transport-security`, `x-content-type-options`, `x-no-cors-reason`.

**Why it's safe to drop.**

- `sanitizeHistory` (`getElectionRollController.ts:39`) already destructures `email_data` out of every history entry before it reaches a client. Its own comment says it "strips out email_data entirely." No consumer has ever seen this field.
- The two informative fields are already persisted properly. `EmailEventsDB.insert` sits ~15 lines above the removed line in the same function and extracts `x-message-id` into `message_id` and `statusCode` into `details.status_code`, from the same `emailResponse` object.
- `getElectionHistoryController.ts:198`, which builds the public election audit log, reads only `action_type` and `timestamp`.

**Why it's worth doing.** `electionRollDB` is append-only copy-on-write — measured at ~4.6 versions per roll on `j87vqp` (3,184 head rows → 14,548 versions), so each payload is stored several times over. On that election, campaign actions account for 8.6MB of 8.9MB of history: **96%**.

Because `getRollsByElectionID` is an unpaginated `selectAll()`, that payload is also read out of Postgres and JSON-parsed into JS objects on every admin voter-list load and every email campaign build, inside a 256Mi pod.

Projected to a 25k-voter election at the observed ~3.5 emails/voter, this is the difference between roughly **70MB and 7MB** of history text per fetch.

**What is deliberately not touched.** The action itself is kept (`action_type`, `actor`, `timestamp` — ~80 bytes) since the audit log and admin views use it. `ElectionRollAction.email_data` stays on the type, marked deprecated, so historical rows still parse and `sanitizeHistory` keeps stripping them. The separate `electionRollDB.email_data` **column** is untouched — `sendInvitesController.ts:52-55` reads it to decide who still needs an invite, so it is live logic, not dead weight.

This only changes what is written going forward. Existing rows keep their payloads; clearing those is a separate migration, and worth noting that a 2024 `x-message-id` is no longer resolvable against SendGrid anyway given their event retention.

## Test plan

- [x] `npm run build -w @equal-vote/star-vote-shared` — clean
- [x] `npm test -w @equal-vote/star-vote-backend` — 211 tests, 27 suites, all pass, including `electionHistory.test.ts` (the audit-log consumer that reads roll history)
- [x] Grepped both packages for readers of `email_data` inside a history action — none exist
- [ ] Post-merge: confirm a real campaign still produces `emailEventsDB` rows with `message_id` populated, and that the admin voter list still renders history


packages/backend/src/Controllers/Election/sendEmailController.ts
```diff
@@ -223,11 +223,16 @@ async function handleSendEmailEvent(job: { id: string; data: email_request_event
...
+    // fields are x-message-id and statusCode, both already recorded in emailEventsDB
+    // just above; the remaining ~660 bytes are HTTP boilerplate (CORS, HSTS, Date).
+    // sanitizeHistory strips email_data before any client sees it, so nothing read it.
+    // electionRollDB is copy-on-write (~4.6 versions/roll), so each payload was stored
+    // several times over -- 96% of history bytes on a large emailed election.
     const historyUpdate: ElectionRollAction = {
         action_type: event.message_id,
         actor: event.sender,
         timestamp: Date.now(),
-        email_data: emailResponse,
```
ArendPeter, line 230: https://github.com/Equal-Vote/bettervoting/blob/164ee983a463bd1f877708048bacd129395a9314/packages/backend/src/Controllers/Election/sendInvitesController.ts#L52

Would we still be able to determine which voters have received an invitation without the email data? Should we store a minimal version to satisfy the above code instead?

ArendPeter: https://github.com/Equal-Vote/bettervoting/blob/164ee983a463bd1f877708048bacd129395a9314/packages/backend/src/Controllers/Roll/getElectionRollController.ts#L181

It's also still included there even though it's scrubbed from the history

ArendPeter: https://github.com/Equal-Vote/bettervoting/blob/164ee983a463bd1f877708048bacd129395a9314/packages/frontend/src/components/EnhancedTable.tsx#L114

Can enhanced table still filter on who received the email?

ArendPeter: from jackson: sendInvitesController.ts isn't affected since that's from a different table

ArendPeter (approved):
I took a close look at all the email_data matches and confirmed that this is safe. Looks good!

PR #1650, opened 2026-09-17
Stop fetching the whole election roll for two single-row operations

jacksonloper:
## Human Summary

(1) the break glass voter id reveal was fetching *all* entries on the roll and then using javascript to find the right one.  filtering in sql will be better. (2) when we finalize the election for no reason we fetched the roll and throw away the result (actually we gate on whether it is null, but it could never be null the way it was written 😂) which is potentially a lot of megs for nothing.

The first one matters most here because whoever is doing bounce triage is going to be breaking a lot of glass for a large election, so we don't want that path to be slow.  

## Summary

Two callers of `getRollsByElectionID` were doing an unpaginated `selectAll()` over every head roll in an election to accomplish something that touches one row or none. `electionRollDB` rows average ~1.3KB stored on an emailed election, so at 25k voters that is tens of megabytes read out of Postgres and JSON-parsed inside a 256Mi pod.

### `revealVoterIdController`

Fetched the entire roll, then `.find()`'d one voter by email in JS. Now uses a new `getByElectionIdAndEmail` model method that filters in SQL.

- `election_id` is the leading column of `electionRollDB_pkey`, so the scan is confined to that election's rows.
- Matching stays case-insensitive (`lower(email) = lower($1)`), identical to the previous `.toLowerCase()` comparison.
- Results are ordered by `voter_id`. Duplicate emails on one roll shouldn't exist — `addElectionRollController` rejects them — but if a legacy duplicate does exist, this resolves to the same voter every time instead of depending on Postgres's return order. The old `.find()` was non-deterministic in that case.

This is the break-glass audit path, so it should be the cheap one.

### `finalizeElectionController`

Fetched the entire roll into a local that was never read again. The accompanying `if (!electionRoll)` guard could never fire either, since `getRollsByElectionID` resolves to `[]` rather than `null` for an empty roll — `.execute()` doesn't return null. Removed, along with the three declarations it left unused (`ElectionRoll` import, `ElectionRollModel`, `electionId`).

**One thing left deliberately undone.** That dead guard may have been intended as "don't finalize an email election with no voters." Implementing it would be a real behaviour change — elections that finalize today would start failing — so it is flagged in a comment rather than smuggled into a performance PR. Happy to add it as an explicit `count(*)` check if that's the desired semantics.

### Context

This is the first of several similar cleanups. Of the six callers of `getRollsByElectionID`, only `getElectionRollController` (the admin voter list) is a UI that wants rows; the rest want a projection or a targeted query. These two were the clearest and lowest-risk, so they go first. `sendEmailController` (needs 3 columns) and `addElectionRollController` (needs an existence check plus a count, currently the source of the n²/1000 cost on chunked uploads) are the higher-value follow-ups.

Independent of #1649 — no overlapping files, either can merge first.

## Test plan

- [x] `npm run build -w @equal-vote/star-vote-shared` — clean
- [x] `npm test -w @equal-vote/star-vote-backend` — 211 tests, 27 suites, all pass
- [x] `finalizeElection.test.ts` passes — covers the modified finalize path
- [x] `electionHistory.test.ts` passes — exercises the `VOTER_ID_REVEALED` audit trail through the new lookup via the mock store
- [x] Mock store (`__mocks__/ElectionRolls.ts`) implements `getByElectionIdAndEmail` with the same case-insensitive + ordered semantics as the real one
- [x] Verified no imports became unused beyond the three removed
- [ ] Post-merge: reveal a voter ID on a real email-list election and confirm the returned voter matches, and that the 🚨 action still lands in the roll's history


ArendPeter (approved):
Looks good!

PR #1651, opened 2026-09-17
Point the Feedback link at Fider, drop the Freshdesk widget

jacksonloper:
## Human Summary

Fider isn't hooked up to the frontend.  This hooks it up to the frontend.  We are abandoning freshdesk (it isn't live anymore) because it costed us money.

## Summary

Points the Feedback link at the Fider instance already running at `feedback.prod.equal.vote` and removes the Freshdesk widget entirely.

### What was there

`openFeedback()` loaded the Freshworks widget script, then reached across into its iframe to drive it:

```ts
const launcherFrame = document.getElementById("launcher-frame");
const button = (launcherFrame as HTMLIFrameElement)
  .contentWindow.document.getElementsByClassName("launcher-button")[0];
(button as HTMLButtonElement).click();
```

A companion `body onload` handler waited 500ms before recoloring that button, with a comment conceding *"there's a race condition with the load order, the timeout helps but we still have the undefined just in case."* All of it depended on markup inside a third-party iframe and would break silently if Freshworks changed their DOM.

To be clear about the current state: **the widget is still live.** `widget.freshworks.com/widgets/63000001746.json` still returns 200 (account 1391020), and the "Feedback?" button still renders on bettervoting.com today. This PR is driven by cost and by consolidating feedback into Fider, not by the widget being broken.

### What it does now

`openFeedback()` opens Fider in a new tab:

```ts
window.open(FEEDBACK_URL, '_blank', 'noopener,noreferrer');
```

Fider is a separate app with its own email sign-in (SMTP via SendGrid, already configured) and no embeddable widget, so a link-out is the shape this wants. New tab rather than same tab so an admin part-way through building an election doesn't lose their place. It fires from a click handler, so popup blockers don't interfere.

The `/feedback` route previously rendered the landing page and popped the widget open from an effect. It now redirects in place through the existing `Redirect` component — someone navigating there directly has already chosen to go.

`FEEDBACK_URL` reads `REACT_APP_FEEDBACK_URL` and falls back to prod, so non-prod deployments can point elsewhere. Documented in `sample.env`.

### The floating button

Dropping the widget also drops its launcher from the lower-right of every page, which is the entry point people actually use — `nav.beta_warning` still tells them to "use the feedback button". So it is rebuilt as a local component, reproduced from the live launcher rather than approximated.

The offsets and label come from the widget config (`widgets/63000001746.json`); the remaining values were read out of the widget's own bundle, where the launcher styles are CSS-in-JS inside `widgetBase/widget.js` rather than in any stylesheet:

| property | live launcher | this PR |
| --- | --- | --- |
| label | `button_text: "Feedback?"` | `nav.feedback`, already "Feedback?" |
| offsets | 30px right, 30px bottom | same |
| `border-radius` | `30px 8px 30px 30px` | same |
| `padding` | `8px` | same |
| `box-shadow` | `0 2px 8px rgba(0,0,0,0.2)` | same |
| `font-size` | `0.875rem` | same |
| `font-weight` | `600` | same |
| icon | 16×16 white question-mark-in-circle, left of label | MUI `HelpOutlineOutlined`, 16×16, white |
| text colour | `#ffffff` | same |
| background | `rgb(134,198,106)` | `var(--brand-pop)` — see below |

Two details worth calling out. The shape is **not** a pill: three corners round at 30px while the top-right stays near-square at 8px, which is what gives it the speech-bubble look. And there **is** an icon — the widget's own SVG is titled "Help", so MUI's `HelpOutlineOutlined` is the same glyph, used instead of copying their path.

Built on `ButtonBase` rather than `Fab`, so the asymmetric radius, padding and elevation are declared outright instead of overriding Fab's defaults.

The background deserves a note. The widget's configured `button_color` is `#006063`, a dark teal — but that was never what shipped. The `onload` hack reassigned it to `#86C66A` on every page load, and you can still see the result in the live DOM as an inline `background-color: rgb(134, 198, 106); animation: 0s; transition: all`. `#86C66A` is `--brand-pop`, so the override simply becomes the declared colour and the timing hack disappears.

Hidden below 900px and in print, matching the two `#launcher-frame` rules that governed the old button.

One deliberate difference: the label inherits the site's font rather than Freshworks' `-apple-system` stack. Now that this is our component rather than a third-party iframe, matching the surrounding UI seems more correct than matching the widget. Easy to pin if preferred.
### Removed

- The Freshworks widget script and `fwSettings` block (`index.html`, 34 lines)
- The `onload` recolor hack and `<body onload="onload()">`
- Both `#launcher-frame` CSS rules, which hid the old button on mobile and in print

Server side needed nothing — Fider is up, `/api/v1/posts` returns 200, `BASE_URL` is correct, and email signup is configured.

## Test plan

- [x] `npm run build -w @equal-vote/star-vote-frontend` — clean
- [x] No `freshworks` / `fwSettings` / `launcher-frame` strings remain in `src/`, `index.html`, or the build output
- [x] `feedback.prod.equal.vote` is correctly inlined into the JS bundle by rsbuild's `publicVars`
- [x] `https://feedback.prod.equal.vote` returns 200 and `/api/v1/posts` returns 200
- [ ] Manual: nav menu → Feedback opens Fider in a new tab, original tab keeps its place
- [ ] Manual: navigating directly to `/feedback` redirects to Fider in the same tab

`npm run lint` fails on this repo for an unrelated pre-existing reason — `eslint-plugin-jsx-a11y` is imported by `eslint.config.js` but not installed, and it fails identically on a clean `main`. Left alone rather than widening this PR; worth a separate fix.


ArendPeter (approved):
Looks good, thanks!

PR #1655, opened 2026-09-18
Pace invite and campaign emails with startAfter; refuse overlapping blasts

jacksonloper:
## Human Summary

for email server reputation if nothing else, we may want to avoid a sudden spike in emails.  This PR lets you send out blasts more slowly.  It doesn't let you pick per election or per blast, it's just a global environment variable.  Maybe good enough for this upcoming potential election?

this pr also makes it so you can't send a blast while a blast you sent earlier is still working.

## Summary

Both email paths enqueued every job with no schedule, so a campaign drained as fast as the workers could pull — `teamSize: 5` per replica, ~10 jobs/s measured, bounded only by how expensive each job happened to be. That's slow by accident, not by design: it speeds up whenever a job gets cheaper or a replica is added, and there is no knob to slow it down.

`publishBatch` now takes an optional `spacingMs` and stamps each job with an evenly spaced `startAfter`. pg-boss fetches on `(name, startafter)`, so the pacing lives in the job rows themselves:

- survives a restart
- independent of replica count
- **aborting a send is deleting the rows whose `startafter` hasn't arrived** — no code needed

The rate comes from `EMAIL_SEND_RATE_PER_MINUTE`. **Default is 600/min, matching unpaced throughput, so an unset variable changes nothing** — small elections behave exactly as today. Prod sets it lower deliberately for a large send. `castVote`'s batch is untouched.

### Measured

Locally at 60/min against a 25k-voter election:

| | |
| --- | --- |
| rows scheduled | 25,000 spanning 417 min |
| gap between consecutive `startafter` | exactly 1000 ms |
| after 33 s | 32 completed, 24,968 still scheduled |
| abort | `DELETE … WHERE startafter > now()` removed 24,938 rows in one statement |

Each enqueue logs its plan: `enqueuing 25000 emails at 60/min (~416.7 min)`.

### In-flight guard

With pacing, a blast to a large election is hours of queued jobs — and nothing stopped an admin from starting another. A double-click, or re-running "send" because the first looked slow, stacked a second full blast behind the first. Nothing checked queue depth at all.

Both send paths now call `assertNoSendInFlight()` before enqueueing. It counts un-started jobs (`created` or `retry`) for the election on **both** queues — an in-progress invite blocks a campaign and vice versa — and refuses with a 400 that uses the pacing rate to say how long the wait is:

> A send to this election is still in progress: 24,998 emails remaining, about 6.9 hours at the current rate. Wait for it to finish before starting another.

It runs **before** `email_campaign_count` increments, so a refused attempt neither bumps the counter nor versions the election row. Verified live: second blast refused, counter unchanged; after deleting the un-started rows, the next blast accepted.

`countUnstarted()` is a new `IEventQueue` method. pg-boss only exposes per-queue totals, so the real implementation queries `pgboss.job` directly (`data` is jsonb). Invite jobs gain a top-level `election_id` so both queues count the same way. The jest mock counts its own pending list, and with its `pause()` the guard is fully exercised in tests.

### Retry policy — deliberately unchanged

Batch jobs have `retrylimit 0` (pg-boss `insert()` default) while `publish()` sets 3. That looks like an inconsistency, and an earlier draft "fixed" it. It shouldn't be: every batch handler **sends before it records** — campaign (`sendEmailController:204` then `:213`/`:241`), invite (`sendInvitesController:137` then `:155`/`:177`), and `castVote`, which sends the vote receipt. A retry after a partial failure would re-send the email. `retryLimit`/`expireInSeconds` are exposed on `PublishBatchOptions` as opt-in only, for a future handler that is idempotent; none currently are.

Worth its own ticket: the single-send `publish()` path *does* retry 3×, and it's used by the manual one-voter invite and the single-vote `castVote` receipt — so the double-send hazard already exists at small scale. The real fix is idempotent handlers, not retry counts.

### Not in this PR

- Per-election or per-blast rate — it's one global variable.
- Any budget on *total* emails per election or per month — the in-flight guard stops overlap, not volume. That's the next piece of work.
- The jest `MockEventQueue` accepts the option and ignores spacing so tests run immediately; the schedule is exercised only against real pg-boss.

## Test plan

- [x] `npm test -w @equal-vote/star-vote-backend` — 211 tests, 27 suites, pass
- [x] Live pg-boss at 60/min: 25,000 rows, 1000 ms spacing, drain follows schedule not worker capacity
- [x] Plan line appears in logs on enqueue
- [x] Deleting un-started rows aborts the remainder
- [x] `sendInFlight.test.ts` (5 tests): accept, refuse while pending, cross-queue refuse of invites, accept after drain
- [x] Live: second blast → 400 with remaining count + ETA; `email_campaign_count` unchanged by the refusal; accepted again after abort
- [x] With the variable unset, spacing is 100 ms (600/min) — indistinguishable from today's throughput
- [ ] Prod: set `EMAIL_SEND_RATE_PER_MINUTE` in the star-server env before the large election; confirm the plan line in Loki after the first blast


ArendPeter (approved):
Looks great!

PR #1656, opened 2026-09-18
Fix card height and background color on Other Tools cards

saberprivateer:
### Human Summary
There was a UI bug on the cards where the color wasn't consistent and was cut-off at the bottom. This fix made it so that independent of the lines of text the entire card color will be consistent.

### Agent Summary
- **Component**: `LandingPageOtherTools.tsx`
- **Issue**: The flex row container stretched all three cards to match the height of the tallest card (which had 3 lines of description), while the custom `lightShade.main` background color was only applied to the inner `CardActionArea` without `height: '100%'`. This left an exposed white strip from the underlying MUI `Card` background at the bottom of the shorter cards.
- **Fix**:
  - Moved `backgroundColor: 'lightShade.main'` to `<Card>` to guarantee consistent background surface color regardless of content.
  - Set `height: '100%'`, `display: 'flex'`, `flexDirection: 'column'`, `alignItems: 'stretch'`, and `justifyContent: 'flex-start'` on `<CardActionArea>` so that the interactive ripple and click surface expands to fill the full card height.
  - Set `height: '100%'` and `width: '100%'` on `<CardContent>`.
- **Verification**: Built the frontend bundle with `npm run build -w @equal-vote/star-vote-frontend` (clean build in 8.34s).

Before
before_cards_white_on_bottom

After
image

Still okay on mobile
image

## Related Issues

None


ArendPeter (approved):
Thanks Dan!

PR #1659, opened 2026-09-19
Allow email blasts for admin-managed voter ID elections

ArendPeter:
## Human Summary
This allows admins to specify both a voter id and an email. This opens up email blasts for users who choose to have admin managed voter ids, and it's particularly relevant for the upcoming FOSS United election.

## Summary
- Admin-managed voter IDs and BetterVoting-managed voter IDs previously treated "who manages the voter ID" and "can we email voters" as mutually exclusive, even though the roll-upload flow already allowed an admin to attach an email to a voter_id-managed row.
- Removed the `invitation === 'email'` requirement from the backend `sendEmailsController` gate — email blasts are now available to any closed-access election, not just BV-managed ones. Roll entries without an email on file are filtered out of the recipient list (with a clear error if none qualify) instead of the whole request being blocked.
- `ViewElectionRolls.tsx`: split the old `usesEmail` flag into `redactsVoterIds` (BV-managed only, controls voter_id column visibility) and `canSendEmailBlast` (any closed election), so "Draft Email Blast" now shows for admin-managed elections too.
- `EditElectionRoll.tsx`: the per-voter "Draft Email" button and email-events history now show whenever a roll entry has an email, regardless of mode; the BV-specific auto-invite status labels stay gated on the existing reveal-flow check.

## Test plan
- `tsc --noEmit` passes for backend, frontend, and shared packages.
- Manually reasoned through the admin-managed + email roll upload path (`AddElectionRoll.tsx` / `addElectionRollController.ts`), which already supported storing an email alongside an admin-managed voter_id with no changes needed.
- Could not run the local Jest suite — `setupTests.ts` fails to resolve in this environment even on an unmodified file, so it's a pre-existing local environment issue unrelated to this change.


ArendPeter (commented):
Looks good to me 😅 .

I've reviewed it carefully, and I think I'll merge this one since the FOSS United Team could feasibly hit this error before the voter limit override goes through.

PR #1664, opened 2026-09-21
Support bulk voter upload via batched requests

ArendPeter:
## Human Summary

This PR allows for large batches of voters to be added to the roll. Previously we'd hit a limit around 1,000 voters. The PR solves this in two ways:
1. By increasing the express limit from 100 KB to 500 KB
2. By uploading the roll across multiple batches using an adaptive batch size

The frontend will filter out any roles that have already been added so that ensures that this operation is safe to repeat even if it fails midway. 

## Summary

**Backend**
- `app.ts`: `express.json` limit raised from the 100 KB default to 500 KB. This is global, so it also applies to the ballot upload route.
- `addElectionRollController.ts`: the duplicate check now uses `Set`s instead of an O(n×m) `.some()` scan. Emails match case-insensitively and trimmed (the voter login lookup is already case-insensitive); voter IDs match trimmed. A duplicate still returns the same 400 error.
- `app.ts` also has some pre-existing lint errors fixed (`require` → `import`, unused vars, `any`), since the diff-scoped lint hook/CI lints whole touched files.

**Frontend**
- New `Election/Admin/rollUploadUtils.ts`:
  - `findRollConflicts` finds entries already on the roll or repeated within the upload (first occurrence wins).
  - `uploadRollsBatched` follows the batching convention from #1611 (`INITIAL_BATCH_SIZE = 700`, `BATCH_SHRINK_FACTOR = 0.75`, `MIN_BATCH_SIZE = 10`, `{ responses, aborted }` result). It shrinks and retries only on 413 / 5xx / network errors; any other error (voter limit, permissions, duplicates) aborts immediately with the server's message. No code is shared with the ballot upload, to keep the two independent until #1611 lands.
- `AddElectionRoll.tsx`: the text field and CSV upload now share one papaparse-based pipeline. Before uploading it re-fetches the roll, checks the voter limit, and finds conflicts. A confirm pop-up appears when there are conflicts to skip, and always for CSV files ("Upload N voters"). A progress bar shows during upload. If the upload aborts, the error says how many voters were added and to re-upload the same list.
- `ViewElectionRolls.tsx`: the Add Voters dialog can't be closed while an upload is running, and closing it now always refreshes the voter list so a partial upload shows up.

**Docs**
- `CONTEXT.md`: added **Voter Roll** and **Roll Conflict**. #1611 also appends to this file, so expect a small merge conflict at the end of the file.

**Not in this PR**
- Other bottlenecks for very large rolls (the full-roll fetch per batch, per-voter ID generation queries) are deliberately deferred.
- No frontend unit tests, since `vitest` only arrives with #1611.

## Test plan
- [x] `npm test -w @equal-vote/star-vote-backend` — 219 passing, including 8 new tests in `addElectionRoll.test.ts` (voter ID / email conflicts, whitespace and case variants, whole-batch rejection, empty emails, request body over 100 KB)
- [x] `tsc --noEmit` clean for frontend and backend; eslint clean on all touched files
- [ ] Manual: upload a ~25k-row CSV to a closed election with a raised voter limit and confirm it completes, then re-upload the same file and confirm every row is reported as skipped
- [ ] Manual: interrupt an upload midway, re-upload the same file, and confirm only the remaining voters are added
- [ ] Manual: confirm the Add Voters dialog can't be closed mid-upload

https://claude.ai/code/session_01AhDMrtGR1PKtxVJPnVPtNn


jacksonloper:
I think `v-abc\nv-def\nv-ghi` would fail as an upload.  ("UndetectableDelimiter").  Is that desired?

jacksonloper:
What happens if the uploaded list has the same voter twice (in the same batch)?

packages/backend/src/Controllers/Roll/addElectionRollController.ts
```diff
@@ -72,20 +72,23 @@ const addElectionRoll = async (req: IElectionRequest & { body: { electionRoll: E
 
     if (existingRolls) {
         // Check if rolls already exist
-        const duplicateRolls = req.body.electionRoll.filter((roll: ElectionRoll) => {
-            return existingRolls.some(existingRoll => {
-                if (existingRoll.email && roll.email && existingRoll.email === roll.email) return true
```
jacksonloper, line 77: If the emails are the same except for case, this will return false.  Is that bad?

jacksonloper:
~I might be crazy but I feel like more of this should happen server-side.  Like I feel we should post a *file to the backend* rather than the frontend parsing the csv and sending a json object (which is indeed going to blow past a 100kb limit).~

~we already have a pattern for uploading files (for the images).  can we just use that?~

~i think uploading ballots should also be more like files?  like all this retry machinery is complex.  and uploading files isn't.~

(i'm not exactly wrong but now isn't the time for it :joy:)

packages/frontend/src/components/Election/Admin/AddElectionRoll.tsx
```diff
@@ -1,187 +1,182 @@
...
+            setProgress(null)
+        }
     }
 
-    function removeDuplicates(checkRolls: RollInput[]): RollInput[] {
-        const seen = new Set<string>();
-        const uniqueRolls: RollInput[] = [];
+    const onSubmit = async (e) => {
+        e.preventDefault()
+        const parsed = Papa.parse<string[]>(voterIDList, { skipEmptyLines: 'greedy' })
```
jacksonloper: I suspect you want `{ delimiter: ',', skipEmptyLines: 'greedy' }` here so that newline-separated list works

ArendPeter: ~~"," is already the default delimiter right?~~

Nevermind, I see

packages/backend/src/Controllers/Roll/addElectionRollController.ts
ArendPeter: The old implementation does a case sensitive check, whereas the new one does a case insensitive check. I think insensitive makes more sense. If the db already has an email with a different case but the same letters then that should be flagged as a duplicate

ArendPeter:
> What happens if the uploaded list has the same voter twice (in the same batch)?

Confirmed, here's what I got
image

jacksonloper:
> > What happens if the uploaded list has the same voter twice (in the same batch)?
> 
> Confirmed, here's what I got image

Ah I see.  So frontend protects.  But if the frontend somehow failed to notice the upload will come back as "failed" and the browser would then retry with a smaller batch size (16 times I believe 😂 ).

jacksonloper (approved):
Hooray!

